PR #384, opened 2024-08-14
Add language to get

jwinston-bn:


src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Endpoint.cs
```diff
@@ -0,0 +1,49 @@
...
+
+public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, IEnumerable<Response>>
+{
+    public override void Configure()
+    {
+        Get("/resources/{ContentId}/available-languages");
+        Description(d => d.ProducesProblemFE(404));
+        Summary(s =>
+        {
+            s.Summary = "For a given resource content id, see in what other languages it is available.";
```
MrGrinst, line 15: Are we using the term `resource content id` in a public facing way? I was thinking we were saying `resource id`. It seems like the term `content id` and `resource id` are being used interchangeably but I'm wondering if it would be clearer if we aligned on just one term.

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Endpoint.cs
```diff
@@ -0,0 +1,49 @@
...
+            .Select(x => new Response
+            {
+                ContentId = x.ResourceContentId,
+                ContentDisplayName = x.DisplayName,
+                LanguageId = x.ResourceContent.LanguageId,
+                LanguageDisplayName = x.ResourceContent.Language.DisplayName,
+                LanguageCode = x.ResourceContent.Language.ISO6393Code,
+                LanguageEnglishDisplayName = x.ResourceContent.Language.EnglishDisplay
+            })
+            .ToListAsync(ct);
```
MrGrinst, line 45: Could we simplify this to one query and return the 404 if the response list is empty?

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Response.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Aquifer.Public.API.Endpoints.Resources.Get.AvailableLanguages;
+
+public class Response
+{
+    public int ContentId { get; set; }
+    public string ContentDisplayName { get; set; } = null!;
+    public int LanguageId { get; set; }
```
MrGrinst, line 7: Do we need language id here? I can't remember if we assign any relevance to that on the public API.

src/Aquifer.Public.API/Endpoints/Resources/Get/Endpoint.cs
```diff
@@ -0,0 +1,38 @@
...
+            s.Summary = "Get specific resource information.";
+            s.Description =
+                "For a given resource id, return the data for that resource. This can be text content as well as CDN links for image, audio, and video media types.";
+        });
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
+    {
+        Response = await ResourceHelper.GetResourceContentAsync(dbContext,
+            new ByLanguage.Request
```
MrGrinst, line 24: It feels weird for what is essentially the parent implementation to be calling into the helper with the "child" request class `ByLanguage.Request`. Could we add another signature in the helper?

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Endpoint.cs
jwinston-bn: The top of the document gives clarity that a resource is a piece of content as we're defining it in the public API. I was noticing what you mentioned, which is why I started using resource content id to be really explicit, since there's a concept of a resource type or a resource collection. Since we define a resource differently from how the broader community seems to (they would call a resource what we call a parent resource), I think it's important to keep the "content" part in there.

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Endpoint.cs
jwinston-bn: I was struggling to come up with a good way to do it, and I don't want to send back the content id that's provided. If you have an idea that runs faster I'm open to it, but this is at least pretty clean and clear what's going on and I don't think the performance is bad.

jwinston-bn: Also, if the list is empty, implying no other languages were found, I don't think it's correct to send back a 404. I see the 404 as being the content id wasn't found at all, whereas an empty list is a 200 and indicates there are simply no other available languages.

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Response.cs
jwinston-bn: You can query the search endpoint with it.

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Endpoint.cs
MrGrinst: Ok that's fair, I'd suggest maybe then we should do an audit of what we've got in the public API (not in this PR) and make sure we're being consistent with `resource content id`.

src/Aquifer.Public.API/Endpoints/Resources/Get/AvailableLanguages/Endpoint.cs
MrGrinst: Could you:
- pull the list (including current language)
- if empty 404
- filter out given content id
- respond

If it's not trivial to switch though then it's not worth it, I just thought it could improve it slightly.

jwinston-bn: I think it'll create a weird query that won't perform any better and isn't worth it personally.

MrGrinst (approved):

PR #401, opened 2024-08-21
Send To Manager Review Now Making Use of CompanyReviewers Table

carlosmorales125:


src/Aquifer.Data/Entities/UserEntity.cs
```diff
@@ -29,6 +28,8 @@ public class UserEntity : IHasUpdatedTimestamp
     [InverseProperty(nameof(CompanyEntity.DefaultReviewerUser))]
     public ICollection<CompanyEntity> CompaniesAsDefaultReviewer { get; set; } = [];
 
+    public ICollection<CompanyReviewerEntity> CompanyReviewers { get; set; } = [];
```
MrGrinst, line 31: Do we need this here? Doesn't look like it's being referenced anywhere.

src/Aquifer.Data/Entities/CompanyReviewerEntity.cs
```diff
@@ -12,8 +13,13 @@ public class CompanyReviewerEntity
 
     [SqlDefaultValue("getutcdate()")]
     public DateTime Created { get; set; } = DateTime.UtcNow;
-
+    
+    [ForeignKey(nameof(UserId))]
```
jwinston-bn, line 17: Don't think we need these attributes.

MrGrinst (approved):

MrGrinst (changes requested):
Could you also remove the DefaultReviewerUser stuff from CompanyEntity? We don't want to do the migration to remove the column in this PR, but it would be nice to deploy the change that removes code-level references alongside this.

jwinston-bn:
> Could you also remove the DefaultReviewerUser stuff from CompanyEntity? We don't want to do the migration to remove the column in this PR, but it would be nice to deploy the change that removes code-level references alongside this.

Doesn't that impact the foreign key relationship though, and by removing it, wouldn't that generate a change in the migration file?

MrGrinst:
> > Could you also remove the DefaultReviewerUser stuff from CompanyEntity? We don't want to do the migration to remove the column in this PR, but it would be nice to deploy the change that removes code-level references alongside this.
> 
> Doesn't that impact the foreign key relationship though, and by removing it, wouldn't that generate a change in the migration file?

It would only affect the migrations if we ran the migration add command. But to avoid the 500s we got last time we removed a column I think we need to remove the code-level reference in one PR, then do the migration in a separate PR.

MrGrinst:
> Yeah, I'm struggling to think of what the ordering should be here. It doesn't feel right for an entity to be in prod that doesn't align with what's in the database. Is your suggestion to also remove `DefaultReviewerUserId` from CompanyEntity now as well?

Yeah we would remove both of them, I should've been clearer but that was my suggestion to remove `DefaultReviewerUserId` and `DefaultReviewerUser`. And the entity will only be in prod for a very short duration before we would deploy the migration to remove the column.

jwinston-bn:
All right. I think that's fine. So my suggestion would be to remove it now, and then have a follow-up PR ready to go up to prod with the migration to deploy immediately after the first goes up.

jwinston-bn (approved):

MrGrinst (approved):

PR #405, opened 2024-08-22
add endpoints for creating and deleting resource references

MrGrinst:


src/Aquifer.API/Endpoints/Resources/ResourceReferences/Get/Request.cs
```diff
@@ -0,0 +1,7 @@
+namespace Aquifer.API.Endpoints.Resources.ResourceReferences.Get;
+
+public record Request
+{
+    public int ParentResourceId { get; set; }
+    public string? ResourceIdOrExternalId { get; set; }
```
jwinston-bn, line 6: I generally wouldn't expect a single property to be two different things, and not only that, but treated as two different types. I think the result of it with the `TryParse` ends up being confusing as a result too. Also notice it's nullable but the Validator restricts that.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Get/Endpoint.cs
```diff
@@ -0,0 +1,44 @@
...
+        query = int.TryParse(request.ResourceIdOrExternalId, out var resourceId)
+            ? query.Where(r => r.Id == resourceId || r.ExternalId == request.ResourceIdOrExternalId)
+                .OrderBy(r => r.Id == resourceId ? 1 : 0)
+            : query.Where(r => r.ExternalId == request.ResourceIdOrExternalId);
+
+        var resource = await query
+            .Select(r => new Response
+            {
+                ResourceId = r.Id,
+                EnglishDisplayName =
```
jwinston-bn, line 29: Is this just for display purposes? I'm wondering if we need this, since I think we just show the EnglishLabel everywhere else that displays resources for selection (resources page, dashboards, project creation, etc.)

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Get/Request.cs
jwinston-bn: Thinking about this more, I think the bigger issue is that the data is bad, and that items got inserted with the external id and there was never a second pass to swap them out with the proper Resource Id. The intention wasn't for the external id to be in there. I think the best approach is to script an update that swaps out those external ids with the proper resource id.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Search/Request.cs
```diff
@@ -0,0 +1,7 @@
+namespace Aquifer.API.Endpoints.Resources.ResourceReferences.Search;
+
+public record Request
+{
+    public int ParentResourceId { get; set; }
+    public string? Query { get; set; }
```
jwinston-bn, line 6: Same as above, Validator doesn't allow null value.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Delete/Endpoint.cs
```diff
@@ -0,0 +1,48 @@
...
+            .ThenInclude(r => r.AssociatedResourceChildren)
+            .SingleOrDefaultAsync(rc => rc.Id == request.ResourceContentId, ct);
+
+        if (resourceContent == null)
+        {
+            await SendNotFoundAsync(ct);
+            return;
+        }
+
+        var referenceResource = await dbContext.Resources.FindAsync([request.ReferenceResourceId, ct], ct);
```
jwinston-bn, line 29: Don't think you meant to pass `ct` twice here.

jwinston-bn: Also, what's the reason for pulling this? It doesn't seem like we do anything with it. If `associatedResource` below is null, we could throw 404 there.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Delete/Endpoint.cs
```diff
@@ -0,0 +1,48 @@
...
+        var associatedResource = resourceContent.Resource.AssociatedResourceChildren
+            .FirstOrDefault(r => r.Id == request.ReferenceResourceId);
+
+        if (associatedResource != null)
+        {
+            resourceContent.Resource.AssociatedResourceChildren.Remove(associatedResource);
+        }
+
+        await dbContext.SaveChangesAsync(ct);
+        await SendOkAsync(ct);
```
jwinston-bn, line 46: Nitpick, but I'd return `NoContent` here.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Create/Endpoint.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        var referenceResource = await dbContext.Resources.FindAsync([request.ReferenceResourceId, ct], ct);
+
+        if (referenceResource == null)
+        {
+            await SendNotFoundAsync(ct);
+            return;
+        }
+
+        if (!resourceContent.Resource.AssociatedResourceChildren.Any(r => r.Id == request.ReferenceResourceId))
```
jwinston-bn, line 37: I'd wrap this conditional around everything with `referenceResource`. If this is false, then there's no reason to go to the database to grab `referenceResource` in the first place.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Create/Endpoint.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            resourceContent.Resource.AssociatedResourceChildren.Add(referenceResource);
+        }
+
+        await dbContext.SaveChangesAsync(ct);
+        await SendOkAsync(ct);
```
jwinston-bn, line 43: Same nitpick, `SendNoContentAsync`

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Create/Endpoint.cs
jwinston-bn: Also seems like the `SaveChangesAsync` would be in the conditional also, since no reason to save if nothing's added.

src/Aquifer.API/Endpoints/Resources/ParentResources/List/Response.cs
```diff
@@ -10,6 +10,7 @@ public class Response
     public required bool Enabled { get; set; }
     public required string ShortName { get; set; }
     public required string DisplayName { get; set; }
+    public required string Code { get; set; }
```
jwinston-bn, line 13: Is this leftover or still being used on the front-end somewhere? Didn't see it getting sent back in any of the new endpoints.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Create/Endpoint.cs
```diff
@@ -0,0 +1,45 @@
...
+using FastEndpoints;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Resources.ResourceReferences.Create;
+
+public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request>
+{
+    public override void Configure()
+    {
+        Post("/resources/resource-references");
```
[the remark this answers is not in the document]
MrGrinst: It mirrors `/resources/bible-references`. It's a resource reference within a resource.

src/Aquifer.API/Endpoints/Resources/ResourceReferences/Get/Endpoint.cs
MrGrinst: Yeah I was thinking display name was better for some reason but it's probably not worth handling.

jwinston-bn (approved):

PR #406, opened 2024-08-28
Add SendGridClient

elisabeth-shakespeare:
add SendGridClient with function SendEmail

### DIFF
--- a/src/Aquifer.Common/Clients/SendGridClient.cs
+++ b/src/Aquifer.Common/Clients/SendGridClient.cs
@@ -1,0 +1,49 @@
+using Aquifer.Data;
+using Microsoft.Extensions.Configuration;
+using SendGrid;
+using SendGrid.Helpers.Mail;
+
+namespace Aquifer.Common.Clients;
+
+public interface ISendGridClient
+{
+    Task<Response> SendEmail(AquiferDbContext dbContext, CancellationToken ct, SendGridEmailConfiguration emailConfiguration);
+}
+
+public class SendGridEmailConfiguration
+{
+    public required string FromEmail { get; set; }
+    public required string FromName { get; set; }
+    public required string Subject { get; set; }
+    public required string HtmlContent { get; set; }
+    public string? PlainTextContent { get; set; } = null!;
+    public required List<EmailAddress> ToAddresses { get; set; }
+}
+
+public class SendGridClient : ISendGridClient
+{
+    private const string ApiKeySecretName = "SendGridMarketingApiKey";
+    private readonly string _apiToken;
+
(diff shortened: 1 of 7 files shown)


src/Aquifer.Common/Clients/SendGridClient.cs
```diff
@@ -0,0 +1,49 @@
+using Aquifer.Data;
+using Microsoft.Extensions.Configuration;
+using SendGrid;
+using SendGrid.Helpers.Mail;
+
+namespace Aquifer.Common.Clients;
+
+public interface ISendGridClient
+{
+    Task<Response> SendEmail(AquiferDbContext dbContext, SendGridEmailConfiguration emailConfiguration, CancellationToken ct);
```
jwinston-bn, line 10: I'd inject the AquiferDbContext in the constructor.

jwinston-bn: though I'm not even seeing where it's being used

elisabeth-shakespeare: Thanks for catching that. I don't use it. I'll take that out.

src/Aquifer.Common/Clients/SendGridClient.cs
```diff
@@ -0,0 +1,49 @@
...
+    private readonly string _apiToken;
+
+    public SendGridClient(IAzureKeyVaultClient keyVaultClient, IConfiguration configuration)
+    {
+        _apiToken = keyVaultClient.GetSecretAsync(ApiKeySecretName).GetAwaiter().GetResult();
+    }
+
+    public async Task<Response> SendEmail(AquiferDbContext dbContext, SendGridEmailConfiguration emailConfiguration, CancellationToken ct)
+    {
+        var client = new SendGrid.SendGridClient(_apiToken);
```
jwinston-bn, line 35: The whole point of making this a singleton is to new this up once.

elisabeth-shakespeare: This is from the SendGrid library, not the Client that I've written.

src/Aquifer.Common/Clients/SendGridClient.cs
```diff
@@ -0,0 +1,50 @@
...
+    public required string Subject { get; set; }
+    public required string HtmlContent { get; set; }
+    public string? PlainTextContent { get; set; } = null!;
+    public required List<EmailAddress> ToAddresses { get; set; }
+}
+
+public class SendGridClient : ISendGridClient
+{
+    private const string ApiKeySecretName = "SendGridMarketingApiKey";
+    private readonly string _apiToken;
```
jwinston-bn, line 26: This can be local to the constructor

jwinston-bn (approved):

PR #407, opened 2024-08-28
Add IpAddressData table

jwinston-bn:
Also handles migration for removed columns from last deployment.

### DIFF
--- a/src/Aquifer.Data/Migrations/20240828185812_AddIpAddressDataTable.cs
+++ b/src/Aquifer.Data/Migrations/20240828185812_AddIpAddressDataTable.cs
@@ -1,0 +1,86 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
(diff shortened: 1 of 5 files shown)


src/Aquifer.Data/Migrations/20240828185812_AddIpAddressDataTable.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Aquifer.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddIpAddressDataTable : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropForeignKey(
```
MrGrinst, line 14: We should double check we've deployed Public API and Jobs to prod before this gets run.

MrGrinst (approved):

PR #410, opened 2024-08-29
add functionality to disable the timestamp updater on the DbContext

MrGrinst:


src/Aquifer.Data/AquiferDbContext.cs
```diff
@@ -92,4 +92,10 @@ private async Task OnSavingChanges(object? sender, SavedChangesEventArgs e)
         var entries = ChangeTracker.Entries();
         await ResourceStatusChangeHandler.HandleAsync(_options, entries);
     }
+
+    // This is useful for scripts you don't want changing the Updated timestamp on tables.
+    public void DisableUpdatedTimestampHandler()
```
jwinston-bn, line 97: What situations would we not want this? Any time we change something in a record the Updated should change.

MrGrinst: For scripting, so it doesn't mess with our timestamps when we run bulk scripts.

PR #413, opened 2024-08-30
Add unsubscribe endpoint

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Marketing/Unsubscribe/Validator.cs
```diff
@@ -0,0 +1,12 @@
+﻿using FastEndpoints;
+using FluentValidation;
+
+namespace Aquifer.API.Endpoints.Marketing.Unsubscribe;
+
+public class Validator : Validator<Request>
+{
+    public Validator()
+    {
+        RuleFor(x => x.UnsubscribeId).NotEmpty().Length(64);
```
elisabeth-shakespeare, line 10: Is having Length validated so specifically a good idea? Would something like MaxLength(100) or something be better so that the string length wouldn't be obvious?

MrGrinst: I'm not sure it would matter having this here since it's just doing a lookup in the DB either way. I don't think it's bad if people know it's 64 chars long but I also don't know if it's worth having the validation 🤷

src/Aquifer.API/Endpoints/Marketing/Unsubscribe/Endpoint.cs
```diff
@@ -0,0 +1,35 @@
...
+            if (subscriberToDisable is not null)
+            {
+                subscriberToDisable.Enabled = false;
+                await dbContext.SaveChangesAsync(ct);
+            }
+        }
+        catch(Exception e)
+        {
+            logger.LogError(e, "Failed to handle subscriber request: {requestContent}", JsonUtilities.DefaultSerialize(req));
+        }
```
MrGrinst, line 29: Do we need this catch here? In most places we aren't doing try catch blocks unless we're reaching out to an external API or something.

elisabeth-shakespeare: I had seen it in the `marketing/subscribers` endpoint, and then BIB-708 says to send OK no matter what, so I had thought maybe having something logged somewhere would be needful. Not sure if that's good enough reasoning though.

MrGrinst: I think the SingleOrDefault shouldn't blow up if it doesn't find something. It'll just skip past the if != null block. Just doing the SendOk at the end without try/catch should be fine.

MrGrinst (approved):

src/Aquifer.API/Endpoints/Marketing/Unsubscribe/Endpoint.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
+    {
+        var subscriberToDisable = dbContext.ContentSubscribers.SingleOrDefault(s => s.UnsubscribeId == req.UnsubscribeId);
+        if (subscriberToDisable is not null)
+        {
+            subscriberToDisable.Enabled = false;
+            await dbContext.SaveChangesAsync(ct);
+        }
+        await SendOkAsync("You have been successfully unsubscribed.", ct);
```
jwinston-bn, line 22: I disagree about removing that try/catch.

The thing that could potentially blow up here is the save, in which case someone legitimately trying to unsubscribe wouldn't be, and we could then take some manual action for that. If the SingleOrDefault doesn't find anything then it'll just move on.

src/Aquifer.API/Endpoints/Marketing/Unsubscribe/Validator.cs
jwinston-bn: If someone wants to be obnoxious they'll see right away to just send a 64 character string, and then keep at it. In the meantime, if we ever decided to change the unsubscribe id to 128 characters or something, we'd have to remember to update this. So I'd just remove it.

src/Aquifer.API/Endpoints/Marketing/Unsubscribe/Endpoint.cs
jwinston-bn: And if it does blow up, we can send back some nice message to try again or whatever, whereas now if it blows up they just get a 500. Remember that this is something the user is going to directly see in their browser.

src/Aquifer.API/Endpoints/Marketing/Unsubscribe/Endpoint.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        Get("/marketing/unsubscribe/{UnsubscribeId}");
+        AllowAnonymous();
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
+    {
+        try
+        {
+            var subscriberToDisable = dbContext.ContentSubscribers.SingleOrDefault(s => s.UnsubscribeId == req.UnsubscribeId);
```
jwinston-bn, line 18: Why aren't we awaiting and using async here?

elisabeth-shakespeare: tried await, but ide said that it is not awaitable,  and async is in the method signature.

jwinston-bn (approved):

PR #421, opened 2024-09-06
Search Now Accepts Optional Offset And Limit

carlosmorales125:


src/Aquifer.API/Endpoints/Resources/Search/Endpoint.cs
```diff
@@ -27,7 +27,7 @@ private async Task<List<Response>> GetResources(Request req, CancellationToken c
             ? ((int?)null, (int?)null)
             : BibleUtilities.GetVerseIds(req.BookCode, req.StartChapter, req.EndChapter, req.StartVerse, req.EndVerse);
 
-        return await dbContext.ResourceContentVersions.Where(x =>
+        var query = dbContext.ResourceContentVersions.Where(x =>
```
MrGrinst, line 30: I think we need to add support for passing a `ParentResourceId` instead of a list of `ResourceType`s. As it stands we will get all resources back for a given type instead of just the one parent resource the user wants. For instance `Dictionary` will end up getting Tyndale, Biblica, and SRV key terms.

MrGrinst (approved):

PR #425, opened 2024-09-09
pin GitHub Action versions

MrGrinst:
To make builds/deployments fully reproducible and avoid untimely failures this pins all GitHub Action versions.


.github/actions/deploy-to-env/action.yml
```diff
@@ -57,12 +57,12 @@ runs:
...
               GITHUB_TOKEN: ${{ inputs.github-token }}
               LINEAR_API_KEY: ${{ inputs.linear-api-key }}
-        - uses: actions/download-artifact@v4
+        - uses: actions/download-artifact@v4.1.8
           with:
               name: build
               path: build
         - name: Login to Azure
-          uses: azure/login@v1
+          uses: azure/login@v2.1.1
```
MrGrinst, line 65: This major version is actually changing here so we should keep an eye out on the next deploy.

jwinston-bn (approved):

jwinston-bn (approved):

PR #428, opened 2024-09-09
add table and endpoint for saving feedback for a resource content version

MrGrinst:


src/Aquifer.Data/Entities/ResourceContentVersionFeedbackEntity.cs
```diff
@@ -0,0 +1,17 @@
+namespace Aquifer.Data.Entities;
+
+public class ResourceContentVersionFeedbackEntity
+{
+    public int Id { get; set; }
+    public int ResourceContentVersionId { get; set; }
+    public string? ContactValue { get; set; }
+    public string? ContactType { get; set; }
```
jwinston-bn, line 8: I'm thinking this should be normalized to a ContactTypes table or at least an enum. We presumably have a fairly fixed group of items here, and I could see us consuming this in more places as time goes on. Rather than have "WhatsApp" as a string repeated, I think it should just have an integer value. That will also force consistency.

jwinston-bn (approved):

PR #429, opened 2024-09-10
Send content updated from previous month email

elisabeth-shakespeare:


src/Aquifer.Common/Enums.cs
```diff
@@ -0,0 +1,6 @@
+namespace Aquifer.Common;
+
+public enum EmailTemplate
+{
+    AquiferMarketingNotification = 1
```
jwinston-bn, line 5: This seems too vague to me. I would call it something like "MarketingNewContentNotification".

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+            var resourcesLanguages =
+                newItems.Aggregate("", (current, item) => current + $"{item.DisplayName} - {item.EnglishDisplay}<br />");
+            var htmlContent = htmlTemplate.Template
+                .Replace("[NAME]", subscriber.Name)
+                .Replace("[RESOURCES]", resourcesLanguages)
+                .Replace("[RESOURCE_LINK]", options.Value.MarketingEmail.ResourceLink)
+                .Replace("[UNSUBSCRIBE]",
+                    $"{options.Value.MarketingEmail.UnsubscribeBaseUrl}/marketing/unsubscribe/{subscriber.UnsubscribeId}?api-key=none");
+            await client.SendEmail(new SendGridEmailConfiguration
```
jwinston-bn, line 70: When operations are spanning multiple lines like this, I'd like to see line breaks for easier readability. So linebreak at line 64 and line 70.

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Extensions.Options;
+using SendGrid.Helpers.Mail;
+
+namespace Aquifer.Jobs;
+
+public class SendNewContentEmails(AquiferDbContext dbContext, SendGridClient client, IOptions<ConfigurationOptions> options)
+{
+    [Function(nameof(SendNewContentEmails))]
+    public async Task Run([TimerTrigger("%NewContentEmail:CronSchedule%")] TimerInfo timerInfo, CancellationToken ct)
+    {
```
jwinston-bn, line 17: This method feels too big to me. I'd generally like to see this split into some private methods like `GetSubscribers` and `GetAllNewItems` etc. so that we're following SRP better. It helps with readability and maintainability.

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+                UnsubscribeId = cs.UnsubscribeId,
+                Languages = cs.ContentSubscriberLanguages.Select(csl => csl.Language),
+                ParentResources = cs.ContentSubscriberParentResources.Select(cspr => cspr.ParentResource)
+            })
+            .ToListAsync(ct);
+
+        var alLanguageIds = subscribers.SelectMany(x => x.Languages.Select(l => l.Id));
+        var allParentResourceIds = subscribers.SelectMany(x => x.ParentResources.Select(pr => pr.Id));
+
+        var allNewItems = await dbContext.ResourceContentVersions.Where(x =>
```
jwinston-bn, line 36: Seems like we may as well check if this is length 0 and quit if it is.

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+                x.IsPublished &&
+                x.Updated >= firstOfLastMonth &&
+                x.Updated <= firstOfThisMonth &&
```
jwinston-bn, line 39: As a massive nitpick, this one should technically be just `<` so it doesn't actually capture part of the current month.

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+                alLanguageIds.Contains(x.ResourceContent.LanguageId) &&
+                allParentResourceIds.Contains(x.ResourceContent.Resource.ParentResourceId))
+            .Select(x => new UpdatedParentResources{ParentResourceId = x.ResourceContent.Resource.ParentResourceId,
+                DisplayName = x.ResourceContent.Resource.ParentResource.DisplayName,
+                LanguageId = x.ResourceContent.LanguageId,
+                EnglishDisplay = x.ResourceContent.Language.EnglishDisplay})
+            .ToListAsync(ct);
+
+        var htmlTemplate = dbContext.EmailTemplates.Single(t => t.Id == (int)EmailTemplate.AquiferMarketingNotification);
```
jwinston-bn, line 48: This should be async

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+                continue;
+            }
+            var resourcesLanguages =
+                newItems.Aggregate("", (current, item) => current + $"{item.DisplayName} - {item.EnglishDisplay}<br />");
+            var htmlContent = htmlTemplate.Template
+                .Replace("[NAME]", subscriber.Name)
+                .Replace("[RESOURCES]", resourcesLanguages)
+                .Replace("[RESOURCE_LINK]", options.Value.MarketingEmail.ResourceLink)
+                .Replace("[UNSUBSCRIBE]",
+                    $"{options.Value.MarketingEmail.UnsubscribeBaseUrl}/marketing/unsubscribe/{subscriber.UnsubscribeId}?api-key=none");
```
jwinston-bn, line 69: Seems like this should just be some sort of self root url, because it's not specific to marketing, it's the server's own self.

src/Aquifer.Jobs/SendNewContentEmails.cs
jwinston-bn, line 70: By convention, I'd expect this to be called `SendEmailAsync`.

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,100 @@
...
+                allParentResourceIds.Contains(x.ResourceContent.Resource.ParentResourceId))
+            .Select(x => new UpdatedParentResources{ParentResourceId = x.ResourceContent.Resource.ParentResourceId,
+                DisplayName = x.ResourceContent.Resource.ParentResource.DisplayName,
+                LanguageId = x.ResourceContent.LanguageId,
+                EnglishDisplay = x.ResourceContent.Language.EnglishDisplay})
+            .ToListAsync(ct);
+
+        var htmlTemplate = dbContext.EmailTemplates.Single(t => t.Id == (int)EmailTemplate.AquiferMarketingNotification);
+
+        foreach (var subscriber in subscribers)
```
jwinston-bn, line 50: I think we're missing some sort of logging on this. I think we should log to app insights an event for each email sent and what was sent.

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -0,0 +1,131 @@
...
+
+public class SendNewContentEmails(AquiferDbContext dbContext, SendGridClient client, IOptions<ConfigurationOptions> options, IHttpContextAccessor httpContextAccessor)
+{
+    [Function(nameof(SendNewContentEmails))]
+    public async Task Run([TimerTrigger("%NewContentEmail:CronSchedule%")] TimerInfo timerInfo, CancellationToken ct)
+    {
+        if (httpContextAccessor.HttpContext == null)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
jwinston-bn: What is this check even for?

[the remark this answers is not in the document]
elisabeth-shakespeare: It's for when getting the server's hostname  for the unsubscribe url, making sure the HttpContext is not null before it does that.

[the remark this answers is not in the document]
jwinston-bn: But this is for Jobs, I don't think that would even work. I think I misspoke before, we need a config that's just root of internal API.

src/Aquifer.Jobs/SendNewContentEmails.cs
elisabeth-shakespeare: Pushed some changes for adding insights for WorkerServices. Going to add more detail for the event, but thought you could take a look at what I have right now while I put the details together

src/Aquifer.Jobs/Aquifer.Jobs.csproj
```diff
@@ -20,7 +20,9 @@
         </PackageReference>
 
     </ItemGroup>
-
+    <ItemGroup>
+        <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
+    </ItemGroup>
```
jwinston-bn, line 25: It's weird that this created a new ItemGroup instead of putting it in the one above that has the package references already.

elisabeth-shakespeare: I added the package wrong.

elisabeth-shakespeare: pushed a fix

jwinston-bn (approved):

PR #431, opened 2024-09-12
add review level column to versions and populate in the right places

MrGrinst:


src/Aquifer.Data/Entities/BibleEntity.cs
```diff
@@ -20,6 +20,7 @@ public class BibleEntity : IHasUpdatedTimestamp
     [SqlDefaultValue("getutcdate()")]
     public DateTime Created { get; set; } = DateTime.UtcNow;
 
+    public LanguageEntity Language { get; set; } = null!;
```
MrGrinst, line 23: This was helpful for some content-loader scripting and as a nice byproduct adds a foreign key relationship there.

jwinston-bn (approved):

PR #432, opened 2024-09-12
return review level in content metadata

MrGrinst:


src/Aquifer.API/Endpoints/Resources/Content/Metadata/Endpoint.cs
```diff
@@ -51,8 +51,9 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
 
         var response = new Response
         {
-            Id = contentVersion.Id,
+            Id = contentVersion.ResourceContentId,
```
MrGrinst, line 54: Noticed that this was pointing at the wrong id and causing some issues.

jwinston-bn (approved):

PR #433, opened 2024-09-13
Fix Projects load time

jwinston-bn:
This fixes the Projects page load times because it won't pull many thousands of resources for the count. However, it slows down the other two pages generally because they have far fewer resources to pull, so counting in memory was then faster.

I think the broader solution is that the progress bar should lazy load on the page, because that part of the query is what's slowing everything down.

### DIFF
--- a/src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCounts.cs
+++ b/src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCounts.cs
@@ -4,19 +4,19 @@ namespace Aquifer.API.Endpoints.Projects;
 
 public class ProjectResourceStatusCounts
 {
-    private readonly List<ResourceContentStatus> _inManagerReviewStatuses =
+    internal static readonly List<ResourceContentStatus> InManagerReviewStatuses =
     [
         ResourceContentStatus.AquiferizeManagerReview,
         ResourceContentStatus.TranslationManagerReview
     ];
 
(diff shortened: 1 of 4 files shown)


MrGrinst (approved):

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCounts.cs
```diff
@@ -4,44 +4,35 @@ namespace Aquifer.API.Endpoints.Projects;
 
 public class ProjectResourceStatusCounts
 {
-    private readonly List<ResourceContentStatus> _inManagerReviewStatuses =
+    internal static readonly List<ResourceContentStatus> InManagerReviewStatuses =
```
NateMerritt, line 7: Nit: `HashSet<T>` is more performant (`O(1)`) than `List<T>` (`O(N)`) when doing a `Contains()` call.  In these cases there are only two items in each list so it won't have a significant impact but this _is_ a performance related PR. 😄

jwinston-bn: Probably not since it's getting translated to a SQL query.

NateMerritt: Ah, good point.  I'm not used to thinking in terms of `IQueryable`.

PR #435, opened 2024-09-13
add display name to machine translations table and endpoint

MrGrinst:


kaseywright (commented):
Nit: Is 'Name' consistent with other places that we reference title(s) for resource content. I.E. instead of using something like `DisplayTitle`?

MrGrinst:
> Nit: Is 'Name' consistent with other places that we reference title(s) for resource content. I.E. instead of using something like `DisplayTitle`?

Yeah good question, `DisplayName` is what we're using in most places I believe

kaseywright (approved):

jwinston-bn (approved):
Thanks

PR #436, opened 2024-09-13
Fix project's resource contents' count and assigned user when there are no draft resource versions.

NateMerritt:
Always use the latest Resource Content Version for a Resource Content when looking up word count and assigned user.


src/Aquifer.API/Endpoints/Projects/Get/Endpoint.cs
```diff
@@ -97,7 +97,18 @@ ELSE U.FirstName + ' ' + U.LastName
...
+                             (
+                                 SELECT *
+                                 FROM
+                                 (
+                                     SELECT
+                                         *,
+                                         ROW_NUMBER() OVER (PARTITION BY ResourceContentId ORDER BY [Version] DESC) AS LatestVersionRank
+                                     FROM ResourceContentVersions
+                                 ) x
+                                 WHERE x.LatestVersionRank = 1
```
NateMerritt, line 110: This chooses the top ranked row when ranked by highest version first.

jwinston-bn (approved):

PR #443, opened 2024-09-18
Add Bibles and Bible Texts endpoints to Public API.

NateMerritt:


src/Aquifer.Data/Entities/BibleBookContentEntity.cs
```diff
@@ -22,4 +27,38 @@ public class BibleBookContentEntity : IHasUpdatedTimestamp
...
+        public required IReadOnlyList<Chapter> Chapters { get; set; }
+
+        public sealed class Chapter
+        {
+            /// <summary>
+            /// The chapter number is a JSON string type but is currently only populated with integer values (this might change in the future).
+            /// </summary>
+            public required string Number { get; set; }
+            public required AudioUrl? Webm { get; set; }
+            public required AudioUrl? Mp3 { get; set; }
```
NateMerritt, line 45: It was clear to me from the JSON data if these are nullable or not.  I assumed that they can be `null` just in case.

src/Aquifer.Public.API/Endpoints/Bibles/List/Request.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Optional language to search on by internal Aquifer language ID. If no language is specified then Bibles for all languages will be returned. Only one of language ID or code may be provided.
+    /// </summary>
+    public int? LanguageId { get; init; }
+
+    /// <summary>
+    /// Optional language to search on by ISO 639-3 code that is 3 characters in length (e.g. "eng"). If no language value is specified then Bibles for all languages will be returned. Only one of language ID or code may be provided.
+    /// </summary>
+    public string? LanguageCode { get; init; }
```
NateMerritt, line 13: I decided to support both `LanguageId` and `LanguageCode` to match the [Search Resources](https://github.com/BiblioNexusStudio/aquifer-server/blob/0f9cb13ccf06e6ba1a82e5131a3acd0fa0cfef04/src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Request.cs#L44-L53) endpoint.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Aquifer.Public.API.Endpoints.Bibles.List;
+
+public class Validator : Validator<Request>
+{
+    public Validator()
+    {
+        RuleFor(x => x.LanguageId)
+            .Empty()
+            .When(x => x.LanguageCode is not null)
+            .WithMessage(x => "'{PropertyName}' must be empty when 'Language Code' is specified.");
```
NateMerritt, line 13: Opinionated rant: I'm not a fan of fluent validation's property name `ToString()` algorithm:
1. It inserts spaces between what it thinks are word boundaries based upon casing.
2. It doesn't have an algorithm to make the new words presentable for public consumption instead of just splitting dev names.  For example, `LanguageId` becomes `'Langage Id'` instead of `'Language ID'`.
3. The default message doesn't do what I want because it doesn't capture the `When` condition, resulting in a potentially confusing error message, so I have to write my own message, anyway.
4. Using `WithMessage` and reaching into the [string format templating args](https://docs.fluentvalidation.net/en/latest/built-in-validators.html) requires looking at the fluent validation docs and make it harder for devs to know what property is meant without cross-referencing the rule lambda in their heads.
5. I couldn't find a way to reach into this naming convention if the property I want to include in the error message is not mentioned in the baked in templating like `{PropertyName}` (hence my having to explicitly reference `'Language Code'` instead of something like `$"{ValidatorOptions.Global.DisplayNameResolver.GetPropertyName(x => x.LanguageCode)}"`).

I'd prefer it just keep the name as the API portrays it (or even better use `languageId` and convert to camel casing just like the JSON or query string presentation).  It's also possible to [globally overwrite the property name algorithm](https://docs.fluentvalidation.net/en/latest/configuring.html#overriding-the-property-name) if we want and then we could inject our own algorithm that we could also reach into for injecting property names that fluent validation doesn't provide via its built-in templating.

Nate, why do you care so much?  Because having clear, consistent error messages is very helpful for consumers of a public API.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
```diff
@@ -0,0 +1,24 @@
...
+        RuleFor(x => x.LanguageCode)
+            .Empty()
+            .When(x => x.LanguageId is not null)
+            .WithMessage(x => "'{PropertyName}' must be empty when 'Language Id' is specified.");
+
+        RuleFor(x => x.LanguageId)
+            .InclusiveBetween(1, 32);
```
NateMerritt, line 20: I copied this from [here](https://github.com/BiblioNexusStudio/aquifer-server/blob/0f9cb13ccf06e6ba1a82e5131a3acd0fa0cfef04/src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Validator.cs#L11).  That said, I'm curious if fluent validation supports any form of DB checks.  Could I, say, fetch all languages from the database, cache the maximum ID, and use that ID in the validator?  Are there examples of doing things like that with DI into the validator class?

Context: I'm used to validation logic sitting at the top of my endpoint's definition where you can intermix both basic request correctness checks with data validation of request parameters.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -0,0 +1,133 @@
...
+using Aquifer.Data;
+using Aquifer.Data.Entities;
+using FastEndpoints;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.Public.API.Endpoints.Bibles.Texts.Get;
+
+public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, Response>
+{
+    private readonly AquiferDbContext _dbContext = dbContext;
```
NateMerritt, line 11: This is a deviation from what looked like standard practice.

I'm fairly new to primary constructors and I don't like the following:
1. The variable is in-scope for my whole class but looks just like a local variable in its naming (`dbContext in this case).
2. I can't mark the variable as read-only.

Therefore, I kept the primary constructor syntax but assigned it to a `private readonly` field here.  The compiler will nicely yell at you if you try to now use `dbContext` elsewhere in the class (solving my first concern) and the field begins with `_` which helps me to recognize it as a field instead of a local variable (solving my second concern).

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -0,0 +1,133 @@
...
+    public override void Configure()
+    {
+        Get("/bibles/{BibleId}/texts");
+        Description(d => d.ProducesProblemFE());
```
NateMerritt, line 15: I didn't add caching here even though I saw it elsewhere because the query string parameters substantially impact the response.  Is that handled by the default caching?

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Response.cs
```diff
@@ -0,0 +1,43 @@
...
+    public required string BibleAbbreviation { get; set; }
+    public required string BookName { get; set; }
+    public required string BookCode { get; set; }
+    public required IReadOnlyList<ResponseChapter> Chapters { get; set; }
+}
+
+public class ResponseChapter
+{
+    public required int Number { get; set; }
+    public required ResponseChapterAudio? Audio { get; set; }
```
NateMerritt, line 16: Note that specifying `false` for `ShouldReturnAudioData` will still result in this `Audio` property being returned in the response JSON, though it will have a `null` (same for the verse's `audioTimestamp`).  If we are earnest about saving bandwidth then there are two options to improve this:
1. Use `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` on these properties.
2. Adjust global JSON serialization settings to omit `null` values for all responses.

I'm in favor of (1) here because the consumer didn't explicitly ask for audio data.  That said, removing the audio properties altogether does make the fact that the route returns audio less discoverable.

Thoughts?

src/Aquifer.Public.API/Endpoints/Bibles/List/Response.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public record Response
+{
+    public required int Id { get; set; }
+    public required string Name { get; set; }
+    public required string Abbreviation { get; set; }
+    public required int LanguageId { get; set; }
+    public required bool IsLanguageDefault { get; set; }
+    public required bool HasGreekAlignment { get; set; }
+    public required bool HasAudio { get; set; }
```
NateMerritt, line 14: I'm a big fan of `bool` property names beginning with `should`, `is`, `has`, etc. because it uses naming convention to convey that the variable is a binary option.  For example, I'd expect a property named `Audio` to contain audio information, not to be a flag telling me if the Bible has audio data.  However, this naming philosophy is not consistent with other responses.

I can change this to be consistent if desired but I wanted to raise the discussion.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Response.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace Aquifer.Public.API.Endpoints.Bibles.Texts.Get;
+
+public class Response
+{
+    public required int BibleId { get; set; }
+    public required string BibleName { get; set; }
+    public required string BibleAbbreviation { get; set; }
+    public required string BookName { get; set; }
+    public required string BookCode { get; set; }
+    public required IReadOnlyList<ResponseChapter> Chapters { get; set; }
```
NateMerritt, line 10: I'm a big fan of `IReadOnlyList<T>` over `ICollection<T>`.  It's a lightweight wrapper that communicates to other devs that the list should not be mutated elsewhere and you can arrays and `List<T>`s to it.

`ReadOnlySet<T>` is coming soon in .net 9.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
```diff
@@ -0,0 +1,53 @@
+﻿using Aquifer.Common.Utilities;
+using Aquifer.Data.Enums;
+using FastEndpoints;
+using FluentValidation;
+
+namespace Aquifer.Public.API.Endpoints.Bibles.Texts.Get;
+
+public class Validator : Validator<Request>
+{
+    public Validator()
```
NateMerritt, line 10: Opinion: I found fluent validation harder to write for some of these more advanced use cases than just standard conditionals at the beginning of my endpoint's handler such as:
```
var validationErrors = new List<string>();
if (myCondition1)
{
    validationErrors.Add("error 1");
}

if (myCondition2)
{
    validationErrors.Add("error 2");
}

if (validationErrors.Any())
    SendErrorResponse(validationErrors);
```

I admittedly had a learning curve.  It seemed like when the rule was easy it was great but when things got complicated it was harder to use than not having the framework.

src/Aquifer.Public.API/Endpoints/Bibles/List/Request.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Aquifer.Public.API.Endpoints.Bibles.List;
+
+public class Request
```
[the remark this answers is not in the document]
MrGrinst: That approach makes sense to me also.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
MrGrinst: I could see the `{PropertyName}` stuff being more beneficial if we factored out some of our validation handling to a util somewhere. In that case yeah maybe it would make sense to have a custom algorithm?

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
MrGrinst: At that point do we gain benefit from having the primary constructor? Is .NET smart enough to dependency inject if you just had: `private readonly AquiferDbContext _dbContext` declared?

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
MrGrinst: Yeah the caching is setup to differ based on any query param. Probably need to double check that is working in the Public API though. I know it works in the internal one.

src/Aquifer.Public.API/Endpoints/Bibles/List/Response.cs
MrGrinst: +1

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
```diff
@@ -22,4 +27,38 @@ public class BibleBookContentEntity : IHasUpdatedTimestamp
 
     [SqlDefaultValue("getutcdate()")]
     public DateTime Updated { get; set; } = DateTime.UtcNow;
+
+    /// <summary>
+    /// POCO for deserializing the AudioUrls column.
+    /// </summary>
+    public sealed class AudioUrlsData
```
NateMerritt, line 34: Opinion: I like the `sealed` keyword on most every class to indicate that I didn't design the class with inheritance in mind.  Any future dev could remove it and take on that responsibility.

I only used it here for these JSON POCOs because I didn't see `sealed` elsewhere in the codebase.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
MrGrinst: I see the value in at as a first layer of validation. To your point and related to a comment above, if it requires a DB check then it's going to have to be in the endpoint, but the `Validator` is nice since:
1. It's declarative vs imperative.
2. The `Request` that the endpoint gets has been checked at a basic level.

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
jwinston-bn: I'm fine with that. I think it's more useful for codebases that are being used by other teams, since I doubt anyone of us are gonna try to inherit anyway, and if we did we'd just remove it. So either way.

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
```diff
@@ -22,4 +27,38 @@ public class BibleBookContentEntity : IHasUpdatedTimestamp
...
+    {
+        public required IReadOnlyList<Chapter> Chapters { get; set; }
```
jwinston-bn, line 36: I wouldn't put `required` on these if they're being instantiated from a deserializer, because if it's not there it's gonna blow up. For example, you've got `Webm` and `Mp3` below as nullable. If they're actually null then it's fine, but if they're simply not there, the `required` is gonna cause the deserialize to blow up.

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
```diff
@@ -22,4 +27,38 @@ public class BibleBookContentEntity : IHasUpdatedTimestamp
...
+                public required string Url { get; set; }
+                public required int Size { get; set; }
+            }
+            public sealed class AudioTimestamp
+            {
+                /// <summary>
+                /// The verse number is a JSON string type but is currently only populated with integer values (this might change in the future).
+                /// </summary>
+                public required string VerseNumber { get; set; }
+                public required decimal Start { get; set; }
```
jwinston-bn, line 59: Nit: Do we need decimal here as opposed to float? I think decimal has more overhead, and don't think we need to worry about the precision here since it's just gonna get passed as is anyway.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
```diff
@@ -0,0 +1,49 @@
...
+                (!request.HasAudio.HasValue || b.BibleBookContents.Any(bbc => bbc.AudioUrls != null) == request.HasAudio))
+            .Select(bible => new Response
+            {
+                Name = bible.Name,
+                Abbreviation = bible.Abbreviation,
+                Id = bible.Id,
+                SerializedLicenseInfo = bible.LicenseInfo,
+                LanguageId = bible.LanguageId,
+                IsLanguageDefault = bible.LanguageDefault,
+                HasGreekAlignment = bible.GreekAlignment,
```
jwinston-bn, line 42: Eh, let's comment this out until we actually add Greek Alignment to the API. If this is there, someone's gonna complain that they can't get it.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Aquifer.Data;
+using FastEndpoints;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.Public.API.Endpoints.Bibles.List;
+
+public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, IReadOnlyList<Response>>
+{
+    private readonly AquiferDbContext _dbContext = dbContext;
```
jwinston-bn, line 9: Don't think we need this. Can just use `dbContext` throughout.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
```diff
@@ -0,0 +1,49 @@
...
+                Id = bible.Id,
+                SerializedLicenseInfo = bible.LicenseInfo,
+                LanguageId = bible.LanguageId,
+                IsLanguageDefault = bible.LanguageDefault,
+                HasGreekAlignment = bible.GreekAlignment,
+                HasAudio = bible.BibleBookContents.Any(bbc => bbc.AudioUrls != null),
+            })
+            .ToListAsync(ct);
+
+        await SendOkAsync(bibles, ct);
```
jwinston-bn, line 47: This is fine and we're not consistent, but as an FYI, you could just say `Response = ...` above (instead of `var bibles = ...`) and remove this line

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public override void Configure()
+    {
+        Get("/bibles");
+        Description(d => d.ProducesProblemFE());
+        Summary(s =>
+        {
+            s.Summary = "Get a list of Bibles.";
+            s.Description =
+                "For a given optional language ID, returns the Bibles in the system including language information, abbreviation codes, whether or not the Bible supports a Greek alignment, and license information.";
```
jwinston-bn, line 19: I know this is a massive nitpick, but we're just saying "id" in most places I think, rather than "ID". Would prefer to be consistent with our grammar regardless of what we go with.

src/Aquifer.Public.API/Endpoints/Bibles/List/Response.cs
jwinston-bn: 100% agree.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
jwinston-bn: You can inject singletons into the validator, but I'm hesitant to add extra db calls without good reason. In a case like this, the chances of someone spamming us with bad language ids is extremely low, so sometimes having some pointless hits to the database because they send a language id higher than what we support seems like far less churn on the database than having to go and check for the list of language ids on _every_ request.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
jwinston-bn: Can you simplify this with pattern matching? See `MachineTranslation/Update/Validator.cs`

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
jwinston-bn: This is just a reality of the primary constructor (I saw people complaining about this while the feature was in preview). Having it declared a second time here is redundant and makes no sense to me. To me the injected services are consistent and it should generally be obvious that's what they are when you call them in the code. There's always a tradeoff with syntactic sugar, and in this case I wouldn't want to go cluttering up classes with constructors and props over a naming issue or because we're worried that someone's gonna try and set an injected value (which shouldn't pass PR review any more than if they did it after injecting it and _not_ setting it as `readonly`).

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
NateMerritt: TIL: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/required-properties.  I can remove `required` but at what point should I assume a data schema?  For example, I'd expect chapter number and verse number to still be required.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
jwinston-bn: I don't think there's any value in caching these. The use case isn't likely going to be someone hitting the same ones over and over again, or different actors hitting it close together.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            s.Summary = "Gets the Bible text contained within a book of the Bible.";
+            s.Description =
+                "For a given Bible and book of the Bible, returns the Bible text (and optional audio information if available) for all verses within the chapter and verse parameters.";
+        });
+    }
+
+    public override async Task HandleAsync(Request request, CancellationToken ct)
+    {
+        var startChapter = request.StartChapter ?? 1;
```
jwinston-bn, line 26: I disagree with this approach. Just set the default value on the prop. Who cares if it was set in the request or not? This just clutters up the code with extra variables and extra lines.

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
NateMerritt: Yes, decimal has more overhead, though that overhead should be insignificant for only a few fields.  I considered using `float` and opted for the precision of decimal which should guarantee the data from the JSON string is represented in .net precisely without strange rounding errors.  See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types:

> The decimal type is appropriate when the required degree of precision is determined by the number of digits to the right of the decimal point. Such numbers are commonly used in financial applications, for currency amounts (for example, $1.00), interest rates (for example, 2.625%), and so forth. Even numbers that are precise to only one decimal digit are handled more accurately by the decimal type: 0.1, for example, can be exactly represented by a decimal instance, while there's no double or float instance that exactly represents 0.1. Because of this difference in numeric types, unexpected rounding errors can occur in arithmetic calculations when you use double or float for decimal data. You can use double instead of decimal when optimizing performance is more important than ensuring accuracy. However, any difference in performance would go unnoticed by all but the most calculation-intensive applications. Another possible reason to avoid decimal is to minimize storage requirements. For example, [ML.NET](https://learn.microsoft.com/en-us/dotnet/machine-learning/how-does-mldotnet-work) uses float because the difference between 4 bytes and 16 bytes adds up for very large data sets. For more information, see [System.Decimal](https://learn.microsoft.com/en-us/dotnet/api/system.decimal).

Do rounding errors on partial seconds within an occasional timestamp matter here?  Probably not that much.  But the overhead to remove them by using `decimal` seems acceptable.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Request.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <summary>
+    /// The ID of the Bible for which to return text.
+    /// </summary>
+    public int BibleId { get; init; }
+
+    /// <summary>
+    /// Book code based off USFM book identifier (e.g. GEN, EXO, etc.). Can get a list of available books and identifiers
+    /// from the /bible-books endpoint. This parameter is required.
+    /// </summary>
+    public string? BookCode { get; init; }
```
jwinston-bn, line 14: If it's required from the validator then no reason for it to be null. Then that removes the need for the bang operator in the endpoint.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Response.cs
jwinston-bn: Yeah, this sounds good to me.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: > At that point do we gain benefit from having the primary constructor?

Not really.  This is why I haven't used primary constructors much and have opted to keep using standard constructors with boilerplate code to assign fields.

> This is just a reality of the primary constructor.

Yep.  I was following the [Microsoft advice](https://devblogs.microsoft.com/dotnet/csharp-primary-constructors-refactoring/#additional-considerations-🤔) for keeping `readonly` which I really like being enforced by the compiler instead of convention or code review.

Thanks for hearing me out.  I'll switch it back to just use the standard primary constructor syntax but wanted to get your thoughts on it first.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: See https://github.com/BiblioNexusStudio/aquifer-server/pull/443/files#r1765532162 for discussion on this.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: It's not a massive nitpick.  At my last job I made everyone use `FooId`/`fooId` for code naming but use `Foo ID` for human readable strings and was notorious for leaving that in code review. 😄 I can argue at length for why [ID](https://en.wikipedia.org/wiki/ID) is correct over "Id", "Id.", and "id" but I'll spare you. 🤣 

I'll switch to `id` instead for consistency.  I just might accidentally slip in an `ID` once in a while because old habits die hard.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
jwinston-bn: These situations are uniquely obnoxious and don't come up too often. Usually the validator is really simple `NotEmpty()` etc. That said, for anything more complex, FastEndpoints has ThrowValidationError or something like that that we can utilize in the endpoint.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: Good call.  I should have confirmed it was available on the public API before copying the internal API filtering.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: I copied [the internal API](https://github.com/BiblioNexusStudio/aquifer-server/blob/81626d70a745af94b4eada04109100b5169ca973/src/Aquifer.API/Endpoints/Bibles/List/Endpoint.cs#L34).

I can update both new endpoints to set the response property if that's best practice.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
NateMerritt: I don't see pattern matching helping with the property naming concern in the error message but it definitely will come in useful for `Must()` validation.

src/Aquifer.Public.API/Endpoints/Bibles/List/Validator.cs
NateMerritt: > on every request

That's why I mentioned caching it. 😄 

In general I agree that this is fine as is.  I was asking more for the theory behind validation.

My main concern with putting `32` into the validation is that now if anyone adds another language in the DB they have to go find all such language ID validation (I copy/pasted this from somewhere else which copy/pasted it from another place) and update it to `33`.  Or more likely they will forget and the API won't be able to return that new language information until a user submits a bug report.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: Some thoughts in reply:
* Yes, it takes an extra line in the endpoint here but I removed a line from the Request (the attribute specifying the default added to each Request property) so it's breaking even on line/code count.
* Putting the default value here helps to put the business logic in one place here in the endpoint instead of having the request contain knowledge about default values.
* The `Validator` is where having the nullability was most useful because I could tell if the user passed a value.
* Nullability exists for cases like this where data may or may not be provided.  This accurately represents the data given to us over the wire.

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
NateMerritt: [Fix commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/c5958ec535fecf230ab2929f1aba1eb78a0214d2) that assumes nested JSON objects/collections might be missing but if those objects/collections are present then they will have expected `string`/`number` properties. I could also go on the defensive for those if desired but I'd argue we should enforce better data at that point.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: [Commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/144e1c77cd0e32015d7103ac7385c5811a2153bb) to remove `HasGreekAlignment` filtering.

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: I did some more research here to understand what FastEndpoints is doing under the hood.

My question: Does setting the `Response` property directly [per these docs](https://fast-endpoints.com/docs/get-started#sending-responses) in `HandleAsync()` have the semantic equivalence to calling `await SendOkAsync(myResponseObject, cancellationToken);` inside `HandleAsync()`?

After some source code digging it looks like the auto-response is [sent here](https://github.com/FastEndpoints/FastEndpoints/blob/1ba7e459633674ac26fd883bb32b69a2ec7d8057/Src/Library/Endpoint/Endpoint.cs#L98) using the same `CancellationToken` that is passed into the `HandleAsync()` [method call](https://github.com/FastEndpoints/FastEndpoints/blob/1ba7e459633674ac26fd883bb32b69a2ec7d8057/Src/Library/Endpoint/Endpoint.cs#L95) that we we override.

It sends [using the default status code](https://github.com/FastEndpoints/FastEndpoints/blob/1ba7e459633674ac26fd883bb32b69a2ec7d8057/Src/Library/Endpoint/Endpoint.Static.cs#L95C58-L95C77) for the framework's `HttpContext.Response` [which is OK according to docs](https://learn.microsoft.com/en-us/dotnet/api/system.web.httpresponse.statuscode?view=netframework-4.8.1&viewFallbackFrom=net-8.0).

So yes, it has the samantic equivalence to `await SendOkAsync(bibles, ct);` which does an equivalent operation [here](https://github.com/FastEndpoints/FastEndpoints/blob/1ba7e459633674ac26fd883bb32b69a2ec7d8057/Src/Library/Endpoint/Endpoint.Send.cs#L138-L143) under the hood.

🤔 In general I don't like magic properties that do something after a method ends in a framework, especially if I only use them some of the time.  Understanding them requires digging into the framework to understand how it works.  Every developer new to the framework (AKA me) has to break through the nice abstraction and go digging in source code to understand how it works or just assume that it's good without understanding it.

This was hard to discover and I prefer the explicit `await SendOkAsync(bibles, ct);` because it's easier to understand how the response is being sent.

*Is there a strong objection against using the explicit send?*

Philosophy: If you haven't noticed, I tend to weigh readability and understandability higher than fewer keystrokes or lines of code when coding.  Yes, fewer keystrokes and fewer lines of code lends itself to readability but sometimes it obscures what is going on.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
NateMerritt: Thanks for elaborating.  It's helpful to hear the philosophy behind validation to understand why the code is the way it is.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: Removed [here](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/3c7bb04481abd2addde6afb8c36e8d6638429d7c).

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
NateMerritt: Fix [commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/15bbd09583a8162999f91dca61b65816c7142a23).

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Response.cs
NateMerritt: I went ahead and did (1) and confirmed that it works as expected: [commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/2573f29a855ad8a628d9f201ed7711002c56dd76).

Note that if `shouldReturnAudioData` is `true` but there is no audio data for the given Bible/book/chapter/verse then we _also_ won't include the JSON properties.  This may be unexpected for consumers.

I'd argue that this is fine because the Bibles List route added in this PR states if Bibles have audio and consumers of JSON can generally handle missing JSON properties by supplying `null`.

If this isn't acceptable and we still want the bandwidth improvements then I see three potential ways to solve this:
1. Create a [custom serialization contract](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/custom-contracts) that only omits `null` properties for the audio types based upon another (non-serialized) property on the object.  AKA add `shouldReturnAudioData` to the response object, don't serialize it, and use it to conditionally exclude `null` audio properties.
2. Use two different response objects (probably through inheritance).  Return one that has audio properties if `shouldReturnAudioData == true` and the other that doesn't have audio properties if `shouldReturnAudioData == false`.
3. Use a separate route for audio data.

Without objection I'll leave this as-is (with my commit at the top of this comment in place).

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: Anther thought: In this case there is a difference in meaning between `null` and `0` or `999` for these start/end parameters.  We only filter text if the user asks for filtering by supplying these parameters and have validation in place to ensure that they are specified together or not at all.  Using non-nullable defaults make it harder to check if the user supplied a value or not.  For example, if the user supplies a start chapter and end chapter of `1` and also `5` and `999` for start verse and end verse, respectively, which is a legitimate request, then we will say the request is invalid because the validation will have to assume that because `999` is the default value that the user didn't supply it in the request (even though they did).

It looks like other validation in [Resource search](https://github.com/BiblioNexusStudio/aquifer-server/blob/4918bd0293f47df24b56b0ed120067af89ab4d0a/src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Validator.cs#L35-L38) gets around this problem by assuming that `0` is essentially equivalent to `null` but it still fails to validate consumers passing `0`.  I'd expect https://qa.api.aquifer.bible/resources/search?languageCode=eng&limit=100&bookCode=gen&startVerse=0&endVerse=0&api-key=741ad467d648470487e0a853e2eb7f64 to return an error but it ignores the consumer input because it has to assume that the consumer didn't pass those `startVerse` and `endVerse` values when doing validation.

src/Aquifer.Data/Entities/BibleBookContentEntity.cs
jwinston-bn: I guess I think about it more around if you're doing operations with the numbers, and these are just read-only, so it wouldn't matter. But either way.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Request.cs
NateMerritt: Good call: [fix commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/a6799102add9ff609464cdead5ab9395c522df78).

src/Aquifer.Public.API/Endpoints/Bibles/List/Endpoint.cs
jwinston-bn: I don't care either way. I was just pointing it out that it's available in the framework.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
jwinston-bn: I don't know what you mean about the validation. It checks that it's greater than or equal to 1 in all cases, whether they pass it or not that validation wouldn't change. If the user passes 0 then it fails and tells them they need to pass 1 or greater. Also don't understand your reasoning around the situation where the user passes 999 for end verse that that would fail. The validator works off the request object and has the default values. If they pass it then the values are what they pass, if they don't, it's what the default is. Whether they pass 999 or not the validation succeeds.

I wouldn't even put default values in the docs on this because it's optional and doesn't matter. The explanation for the fields should say anyway that if they're not provided it will grab the whole book or whole chapter whatever.

By doing this we're allocating four new variables that don't do anything so it's inefficient, not to mention the extra overhead of wrapping int in Nullable (we're gettin' pedantic now!).

It's confusing as a reader, because now I have to try and understand why the request values aren't being used and have a default value being set to a different variable instead of just having a default themselves.

Again I come back to the question of who cares if the user supplied it or not as long as they get the right data? Why does it matter if they sent 1 for the start verse or didn't send anything for the start verse? The result is the same. No one is looking at these request payloads. And saying nullability exists for reasons like this is also why default values exist. We generally don't want things to be null, and if there's a sensible default value, we should use it to avoid having to deal with null. Null is generally a bad thing and creates bugs. There's a reason you have to explicitly call out your strings as nullable now. Nullability introduces the possibility for bugs, and should be avoided as much as possible. There are people who even go to such extremes as wrapping all their objects in a NotNull object. My experience has always been that null is a bad thing and you should avoid it as much as reasonable.

The thing with the business logic doesn't really resonate with me, since it having a null value default is just as much business logic as it having a default of 1. Should every bool be nullable so that it doesn't have a default value of false? Should every int be nullable so it doesn't have a default value of 0? If a model's default value is business logic, then is the model itself business logic that should be deleted? The model dictates what the default values are for itself, whether that's null or something else. Same with collections, I would always make a collection default empty, rather than null, because it makes the code safer. That's not business logic, that's just defensive coding.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        RuleFor(x => x.BookCode).NotEmpty();
+        RuleFor(x => x.BookCode)
```
jwinston-bn, line 13: I think this can be simplified by putting `.NotEmpty()` on this one and remove the `.When`, because we're already saying it can't be null. The `.When` isn't a valid case.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
```diff
@@ -0,0 +1,53 @@
...
+            .NotEmpty()
+            .When(x => x.StartVerse is not null)
+            .WithMessage(x => "'{PropertyName}' must also be provided when 'Start Verse' is provided.");
+        RuleFor(x => x.EndChapter)
+            .NotEmpty()
+            .When(x => x.EndVerse is not null)
+            .WithMessage(x => "'{PropertyName}' must also be provided when 'End Verse' is provided.");
+
+        RuleFor(x => x.StartChapter)
+            .GreaterThanOrEqualTo(1);
```
jwinston-bn, line 28: I'd replace all these with `.InclusiveBetween(1, 999)`

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: > Should every bool be nullable so that it doesn't have a default value of false?

Yes, that's the logical conclusion of my argument.  I've done just that in the past for query string parameters and request body parameters.  `null` means "not passed" and non-null means "provided by the consumer."  URL path variables, however, are not optional because they are required in order to invoke the correct route handler and thus they can't be null.

> Null is generally a bad thing and creates bugs.

I agree with avoiding `null` in the cases where `null` doesn't have meaning.  I'm a big fan of the new(er) non-nullable types for that reason (e.g. `string` vs `string?`) because the compiler helps you to avoid worrying about `null` states.

> Again I come back to the question of who cares if the user supplied it or not as long as they get the right data?

I don't think they will get the correct response.  I'll go back to the pre-existing resources search route on the Public API (https://qa.api.aquifer.bible/resources/search?languageCode=eng&limit=100&bookCode=gen&startVerse=0&endVerse=0&api-key={key}) for my example.  I'd argue that this should return an error response to the consumer instead of returning the entire chapter.  But with current defaults of `0` it's not possible to do that.  You could, say, change the defaults to `-1` instead of `0` but then if a consumer passes `-1` for start verse and end verse then they'll get the entire chapter instead of an error message. There's no way to solve this bug without differentiating between "the user passed data" and "the user didn't pass data" which I don't see being possible with default values on non-nullable `int`s for those parameters.

There's an argument that we're fine with this bug and it's worth that weird edge case in order to have simpler code.  But I don't think I'm making the code that more complex or incurring significant performance hits and the compiler kindly reminds me with warnings if I forget to check for `null` which prevents bugs.

src/Aquifer.Public.API/Endpoints/Bibles/List/Request.cs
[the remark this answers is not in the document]
NateMerritt: See https://github.com/BiblioNexusStudio/aquifer-server/pull/443/files#r1765718263 for conversation on this.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
NateMerritt: That doesn't work unless I set the default value for `BookCode` on the request to `""` instead of `null!`.  Is that what you're thinking?

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
jwinston-bn: I think it's fair that the search example is a little weird because it treats 0 as a starting point. But it's also a bit of a different situation because start and end verse being null is actually a valid value. So I could see the argument for those actually being nullable, because verse references aren't required for finding a resource. However, in the case of the Bible text endpoint, they are required to pull the correct text.

There's no bug though. The validator specifically allows 0 as an input in the search. In the case of this request, we wouldn't be allowing 0 as a valid value, so someone passing startVerse=0 would return an error that they need to pass a valid value between 1 and 999. So I'm unclear still on how you say that we must know if the user passed a value or not, because whether they pass 1 or they pass nothing and the default is 1, they get the same result.

Overall though, I'm just not hearing a reason for why we would sacrifice simplicity and cleaner code for something that has no impact on the correctness of the system. I think it's still coming down to the question for me of why we care if the user passed the default or didn't, when the result is the same.

NateMerritt: > So I'm unclear still on how you say that we must know if the user passed a value or not, because whether they pass 1 or they pass nothing and the default is 1, they get the same result.

Ok.  I looked at the public API for validation examples for the start/end chapter and verse parameters and found [the public API's Resource search validation](https://github.com/BiblioNexusStudio/aquifer-server/blob/4918bd0293f47df24b56b0ed120067af89ab4d0a/src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Validator.cs#L30-L41).  This validation is more complicated and does things like requiring start verse to be present if end verse is also present and requiring start chapter to be present if start verse is present.

It sounds like I should not copy that example and should instead do something more simple like the internal API's [bible text validation](https://github.com/BiblioNexusStudio/aquifer-server/blob/4918bd0293f47df24b56b0ed120067af89ab4d0a/src/Aquifer.API/Endpoints/Bibles/Texts/Get/Validator.cs#L12-L13) which only confirms that `start <= end` (and probably add in the `InclusiveBetween(1, 999)`?

jwinston-bn: I would, yeah.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
jwinston-bn: I'm not sure what you mean. `NotEmpty()` checks against null. It shouldn't go into the `Must` if it's null or empty string. So I think this is all you need:
```csharp
RuleFor(x => x.BookCode).NotEmpty()
            .Must(code => BibleBookCodeUtilities.IdFromCode(code) != BookId.None)
            .WithMessage(x => "Invalid '{PropertyName}': \"{PropertyValue}\".");
```

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
```diff
@@ -0,0 +1,50 @@
...
+
+public class Validator : Validator<Request>
+{
+    public Validator()
+    {
+        RuleFor(x => x.BookCode).NotEmpty();
+        RuleFor(x => x.BookCode)
+            .Must(code => BibleBookCodeUtilities.IdFromCode(code) != BookId.None)
+            .When(x => x.BookCode is not null)
+            .WithMessage(x => "Invalid '{PropertyName}': \"{PropertyValue}\".");
```
jwinston-bn, line 16: I don't think you need the `x =>` on any of these. `WithMessage` takes just a string.

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
jwinston-bn: Never mind, I'm wrong about this.

NateMerritt: I figured out two ways to do the rule chaining.  Given
```
    public string BookCode { get; init; } = null!;
```
both of these work on my local:

1. [Set the cascade mode to stop](https://docs.fluentvalidation.net/en/latest/cascade.html):
```
        RuleFor(x => x.BookCode)
            .Cascade(CascadeMode.Stop)
            .NotEmpty()
            .Must(code => BibleBookCodeUtilities.IdFromCode(code) != BookId.None)
            .WithMessage(x => "Invalid '{PropertyName}': \"{PropertyValue}\".");
```

2. Use [Dependent Rules](https://docs.fluentvalidation.net/en/latest/dependentrules.html) (but the author of the library prefers using the cascade mode):
```
        RuleFor(x => x.BookCode)
            .NotEmpty()
            .DependentRules(() => RuleFor(x => x.BookCode)
                .Must(code => BibleBookCodeUtilities.IdFromCode(code) != BookId.None)
                .WithMessage(x => "Invalid '{PropertyName}': \"{PropertyValue}\"."));
```

NateMerritt: I went with `CascadeMode.Stop`: [commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/915feb0d16be4d3215c483cc56b3e3975e5ae498).

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: I switched to using defaults instead of nullable here: [commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/915feb0d16be4d3215c483cc56b3e3975e5ae498).

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Validator.cs
NateMerritt: Thanks. That was an artifact from when I was generating my own property names using `nameof()` and I forgot to remove it after I discovered FluentValidation's templating system.

Fixed [here](https://github.com/BiblioNexusStudio/aquifer-server/pull/443/commits/915feb0d16be4d3215c483cc56b3e3975e5ae498).

jwinston-bn (approved):

jwinston-bn (dismissed):

jwinston-bn (approved):

PR #445, opened 2024-09-18
Move List Bible Books endpoint to /bibles/books.

NateMerritt:


src/Aquifer.Public.API/Endpoints/Bibles/Books/List/Endpoint.cs
```diff
@@ -2,13 +2,14 @@
...
 
-namespace Aquifer.Public.API.Endpoints.BibleBooks.List;
+namespace Aquifer.Public.API.Endpoints.Bibles.Books.List;
 
 public class Endpoint : EndpointWithoutRequest<IEnumerable<Response>>
 {
     public override void Configure()
     {
-        Get("/bible-books");
+        // "/bible-books" is still supported for legacy compatibility but "/bibles/books" is the new hotness
```
jwinston-bn, line 11: I think we can remove these comments since the LegacyEndpoint implies it.

NateMerritt:
I updated this PR to use inheritance so that we could have separate docs on the legacy endpoint and the new endpoint in order to have a deprecation message.

src/Aquifer.Public.API/Endpoints/Bibles/Books/List/LegacyEndpoint.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public class LegacyEndpoint : Endpoint
+{
+    public override void Configure()
+    {
+        // "/bible-books" is still supported for legacy compatibility but "/bibles/books" is the new hotness
+        Get("/bible-books");
+        Options(EndpointHelpers.UnauthenticatedServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
+
+        // we can eventually hide this endpoint from the docs altogether: https://fast-endpoints.com/docs/swagger-support#filtering-endpoints
```
jwinston-bn, line 13: Can we just remove it actually?

NateMerritt: Sure.  I'll look into apply endpoint labels tomorrow.  I'm hoping we can just make a "no-docs" label or something like that and apply it with a single line of configuration after I wire things up.

src/Aquifer.Public.API/Endpoints/Bibles/Books/List/Endpoint.cs
NateMerritt: Done.

src/Aquifer.Public.API/Endpoints/Bibles/Books/List/LegacyEndpoint.cs
NateMerritt: Done [here](https://github.com/BiblioNexusStudio/aquifer-server/pull/445/commits/f6621945b6df3c8565e0404f3648361e715887c9).  I'm pretty happy with this solution.

NateMerritt:
> I updated this PR to use inheritance so that we could have separate docs on the legacy endpoint and the new endpoint in order to have a deprecation message.

Instead of adding the word "deprecated" to the docs this PR now uses Swagger endpoint filtering to omit the deprecated endpoint from the docs entirely:
image

jwinston-bn (approved):

jwinston-bn (approved):

PR #448, opened 2024-09-20
add resource contents table index for performance

MrGrinst:
Another high impact index suggested by Azure SQL.


NateMerritt:
Note that Azure SQL also supports auto-tuning for SQL DBs: https://learn.microsoft.com/en-us/azure/azure-sql/database/automatic-tuning-overview?view=azuresql.  It can create and drop indexes as it deems fit based upon traffic and performance.

I don't like two things about this:
1. The index is no longer represented in source code.
2. The dev and prod DBs can have different indexes.

That said, it takes a lot of the pain away and in general "just works."  In the past I've turned it on but still manually created basic indexes (PKs, FKs, unique constraints, and indexes on other columns I know will be needed for lookups such as enums).

MrGrinst:
@NateMerritt Yeah good callout, I've seen that but agree with what you said about the downsides. I think having the control is worth the extra hassle of a PR now and then to address things. I have been wanting to setup alerts that would ping our alerts channel when there's a new index recommendation but haven't been able to figure out how.

jwinston-bn (approved):

PR #449, opened 2024-09-20
force the resources content endpoint main query to be split

MrGrinst:
Tested locally and in dev. Almost doubles the speed of the endpoint.
Before: 700-1200ms
After: 300-800ms


src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -66,6 +66,7 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
                 ProjectEntity = rc.Projects.FirstOrDefault(),
                 HasPublishedVersion = rc.Versions.Any(rcv => rcv.IsPublished)
             })
+            .AsSplitQuery()
```
NateMerritt, line 69: TIL: https://learn.microsoft.com/en-us/ef/core/querying/single-split-queries

jwinston-bn (approved):

PR #451, opened 2024-09-20
add /resources/content/needs-translation endpoint for community reviewers to search with

MrGrinst:


src/Aquifer.API/Common/Constants.cs
```diff
@@ -54,31 +54,33 @@ public static class Constants
...
         ReadProject = "read:projects",
         ReadProjectsInCompany = "read:projects-in-company",
-        CreateUser = "create:user",
-        CreateUsersInCompany = "create:users-in-company",
         ReadReports = "read:reports",
-        AiSimplify = "ai:simplify",
-        AiTranslate = "ai:translate",
-        DisableUser = "disable:users",
-        DisableUsersInCompany = "disable:users-in-company";
+        ReadResourceLists = "read:resource-lists",
```
MrGrinst, line 79: Sorted these all and added this one

src/Aquifer.API/Endpoints/Resources/Content/AssignedToSelf/History/Endpoint.cs
```diff
@@ -10,15 +11,16 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService) : En
     public override void Configure()
     {
         Get("/resources/content/assigned-to-self/history");
+        Permissions(PermissionName.ReadResources);
```
MrGrinst, line 14: Locks this down so community reviewers and report viewers can't use it.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToSelf/List/Endpoint.cs
```diff
@@ -32,6 +33,7 @@ ORDER BY RCVAUH.Id DESC
     public override void Configure()
     {
         Get("/resources/content/assigned-to-self");
+        Permissions(PermissionName.ReadResources);
```
MrGrinst, line 36: Locks this down so report viewers can't use it.

src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -15,6 +15,7 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService) : En
     public override void Configure()
     {
         Get("/resources/content/{Id}");
+        Permissions(PermissionName.ReadResources);
```
MrGrinst, line 18: Locks this down so community reviewers and report viewers can't use it.

src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -15,6 +15,7 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService) : En
     public override void Configure()
     {
         Get("/resources/content/{Id}");
+        Permissions(PermissionName.ReadResourceLists);
```
MrGrinst, line 18: Locks this down so community reviewers and report viewers can't use it.

src/Aquifer.API/Endpoints/Resources/Content/List/Endpoint.cs
```diff
@@ -11,6 +13,7 @@ public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, Response>
     public override void Configure()
     {
         Get("/resources/content");
+        Permissions(PermissionName.ReadResourceLists);
```
MrGrinst, line 16: Locks this down so community reviewers and report viewers can't use it.

src/Aquifer.API/Endpoints/Resources/Content/List/NeedsTranslation/Endpoint.cs
```diff
@@ -0,0 +1,135 @@
...
+
+public class Endpoint(AquiferDbContext dbContext, IUserService userService) : Endpoint<Request, Response>
+{
+    public override void Configure()
+    {
+        Get("/resources/content/needs-translation");
+        Permissions(PermissionName.ReadResources);
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
```
MrGrinst, line 20: This is largely a copy of `Resources.Content.List` but was different enough to be it's own thing and not try to factor out any shared logic.

src/Aquifer.API/Endpoints/Resources/Content/List/NeedsTranslation/Endpoint.cs
jwinston-bn, line 1: Kind of a nitpick, don't think it matters too much either way, but it feels like it would better follow our current folder structure if it was /NeedsTranslation/List instead of the other way around.

src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -134,13 +132,7 @@ await dbContext.ResourceContentVersionAssignedUserHistory
...
             .Where(rcv => rcv.Id != relevantContentVersion.Id && rcv.ResourceContentId == relevantContentVersion.ResourceContentId)
             .OrderByDescending(rcv => rcv.Created)
-            .Select(rcv => new VersionResponse
-            {
-                Id = rcv.Id,
-                Created = rcv.Created,
-                Version = rcv.Version,
-                IsPublished = rcv.IsPublished
-            })
+            .Select(rcv => new VersionResponse { Id = rcv.Id, Created = rcv.Created, Version = rcv.Version, IsPublished = rcv.IsPublished })
```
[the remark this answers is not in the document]
MrGrinst: Yeah I'm not sure what is happening on these, I'm using the Rider formatter with our .editorconfig so I'm wondering if it's a .editorconfig prop that's missing. I'll dig in more.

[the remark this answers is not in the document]
MrGrinst: @jwinston-bn updated a couple .editorconfig settings and got it consistent.

jwinston-bn (approved):

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Validator.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        RuleFor(x => x.ParentResourceId)
+            .GreaterThan(0).When(x => x.ParentResourceId.HasValue);
+
+        RuleFor(x => x.BookCode)
+            .NotEmpty().When(x => x.StartChapter.HasValue && x.EndChapter.HasValue)
+            .WithMessage("BookCode must be specified if StartChapter and EndChapter are specified.");
+
+        RuleFor(x => x.StartChapter)
+            .GreaterThan(0).When(x => x.StartChapter.HasValue);
```
NateMerritt, line 18: I was messing around with validation on nullable properties and IIRC the `When(x => x.Foo.HasValue)` is not necessary.  The validator's smart enough to only do the greater than check if the value is non-null.

MrGrinst: Ah good find, this was copypasta from a different endpoint but would be good to update that in both spots at some point

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Validator.cs
```diff
@@ -0,0 +1,40 @@
...
+        RuleFor(x => x)
+            .Must(x => (x.StartChapter.HasValue && x.EndChapter.HasValue) || (!x.StartChapter.HasValue && !x.EndChapter.HasValue))
+            .WithMessage("Both StartChapter and EndChapter must be specified together.");
+
+        RuleFor(x => x.SearchQuery)
+            .NotEmpty().When(x => x.SearchQuery != null)
+            .WithMessage("SearchQuery must be non-empty if specified.");
+
+        RuleFor(x => x)
+            .Must(x => x.ParentResourceId.HasValue || x.BookCode != null || x.SearchQuery != null)
```
NateMerritt, line 37:
```suggestion
            .Must(x => !x.ParentResourceId.HasValue || x.BookCode != null || x.SearchQuery != null)
```

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Validator.cs
jwinston-bn: Yeah, if want to verify something isn't null have to add the `NotEmpty()` on there too. Feels weird, but seems to work best most of the time.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private static string ApplySearchQueryFilter(bool hasSearchQuery, bool isExactSearch)
+    {
+        if (!hasSearchQuery)
+        {
+            return "";
+        }
+
+        return isExactSearch ? "AND R.EnglishLabel = @SearchQuery" : "AND R.EnglishLabel LIKE @SearchQuery";
```
NateMerritt, line 131: Heads up that using wildcards in a search query such as `LIKE N'%query%'` can have a significant performance impact by causing table scans to be necessary.  It may be fine on some tables but larger tables could see issues because the SQL server can't use any index and has to compare row by row.

In this case I suggest doing some timings of exact vs wildcard searching to see if there are any issues because my understanding is that the resources table is fairly large.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+                {ApplySearchQueryFilter(hasSearchQuery, isExactSearch)}
+                {ApplyBookAndChapterFilter(req.BookCode, req.StartChapter, req.EndChapter)}
+                {(getTotalCount ? "" : groupingWithOffset)}
+                """;
+    }
+
+    private static string ApplyLanguageIdFilter(int languageId)
+    {
+        return $"""
+                AND RC.LanguageId = 1 AND RCV.IsPublished = 1 AND NOT EXISTS (
```
NateMerritt, line 77: > `RC.LanguageId = 1`

Is this supposed to be hardcoded to 1?

jwinston-bn: 1 is always English, and in this case we base needing a translation off the existence of a published English resource.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+    private static string ApplyBookAndChapterFilter(string? bookCode, int? startChapter, int? endChapter)
+    {
+        var verseRange = BibleUtilities.VerseRangeForBookAndChapters(bookCode, startChapter, endChapter);
+        if (verseRange is null)
+        {
+            return "";
+        }
+
+        var startVerseId = verseRange.Value.Item1;
+        var endVerseId = verseRange.Value.Item2;
```
NateMerritt, line 98: I suggest updating `BibleUtilities.VerseRangeForBookAndChapters()` [here](BibleUtilities.VerseRangeForBookAndChapters) to return a `ValueTuple` with named items (like the other methods in that class [such as this one](https://github.com/BiblioNexusStudio/aquifer-server/blob/bf045f44c56bd5ce8b272da67a218b3dbe85f792/src/Aquifer.Common/Utilities/BibleUtilities.cs#L77)) so that you can refer to them by name here instead of `Item1`/`Item2`.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        return $"""
+                AND (
+                EXISTS (SELECT 1
+                    FROM [VerseResources] AS [v]
+                    WHERE [R].[Id] = [v].[ResourceId] AND [v].[VerseId] >= {startVerseId} AND [v].[VerseId] <= {endVerseId})
```
NateMerritt, line 104: This could use the `BETWEEN` operator.  I don't know if there's a performance difference but I find it more readable:
```suggestion
                    WHERE [R].[Id] = [v].[ResourceId] AND [v].[VerseId] BETWEEN {startVerseId} AND {endVerseId})
```
Same below.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+                EXISTS (SELECT 1
+                    FROM [PassageResources] AS [p1]
+                    INNER JOIN [Passages] AS [p2] ON [p1].[PassageId] = [p2].[Id]
+                    WHERE [R].[Id] = [p1].[ResourceId] AND [p2].[EndVerseId] >= {startVerseId} AND [p2].[EndVerseId] <= {endVerseId})
+                OR
+                EXISTS (SELECT 1
+                    FROM [PassageResources] AS [p3]
+                    INNER JOIN [Passages] AS [p4] ON [p3].[PassageId] = [p4].[Id]
+                    WHERE [R].[Id] = [p3].[ResourceId] AND [p4].[StartVerseId] <= {startVerseId} AND [p4].[EndVerseId] >= {endVerseId})
+                )
```
NateMerritt, line 120: I'm guessing that you'll see similar or possibly better performance if you combine these three SELECTS into one SELECT with OR clauses in the WHERE condition.  I also think it's easier to understand.
```suggestion
                EXISTS
                (
                    SELECT 1
                    FROM [PassageResources] AS [p]
                    INNER JOIN [Passages] AS [p0] ON [p].[PassageId] = [p0].[Id]
                    WHERE [R].[Id] = [p].[ResourceId] AND
                    (
                        [p0].[StartVerseId] BETWEEN {startVerseId} AND {endVerseId} OR
                        [p0].[EndVerseId] BETWEEN {startVerseId} AND {endVerseId} OR
                        ([p0].[StartVerseId] <= {startVerseId} AND [p0].[EndVerseId] >= {endVerseId})
                    )
                )
```

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            return "";
+        }
+
+        return isExactSearch ? "AND R.EnglishLabel = @SearchQuery" : "AND R.EnglishLabel LIKE @SearchQuery";
+    }
+
+    private static (bool hasSearchQuery, bool isExactQuery) GetSearchParams(string? searchQuery)
+    {
+        var hasSearchQuery = searchQuery is not null;
```
NateMerritt, line 136: The validator is protecting against `searchQuery == ""` but that code is in a different class and could accidentally change without reference to this method.  For defense in depth I suggest also return `hasSearchQuery = false` when `searchQuery` is empty or whitespace.
```suggestion
        var hasSearchQuery = searchQuery is not null && !string.IsNullOrWhiteSpace(searchQuery);
```

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
```diff
@@ -0,0 +1,139 @@
...
+        var resourceContent = total == 0
+            ? []
+            : await dbContext.Database
+                .SqlQueryRaw<ResourceContentResponse>(BuildQuery(req, user, false, hasSearchQuery, isExactSearch), searchParameter)
+                .ToListAsync(ct);
+
+        var response = new Response
+        {
+            ResourceContents = resourceContent,
+            Total = total
```
NateMerritt, line 38: It looks like the SQL query to fetch the total count is only used to populate this property.  Will total be shown in the UI to the user according to the current spec?  And even if so, is it necessary to show a total count in the UI or can we get rid of it to avoid making the additional SQL call altogether?

I'm not opposed to fetching the total if there is a good case to show it to the user.  I just want to make sure it's necessary.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
MrGrinst: Yeah in my mind this was a known/accepted part of allowing partial searches. We haven't seen awful performance with it so far but worth keeping an eye on and potentially index and switch to CONTAINS if it gets bad.

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
MrGrinst: Good call

src/Aquifer.API/Endpoints/Resources/Content/NeedsTranslation/List/Endpoint.cs
MrGrinst: 👍  Yeah it's being used to determine number of pages available on the frontend

PR #453, opened 2024-09-26
address leftover feedback from previous PR

MrGrinst:


src/Aquifer.API/Endpoints/Resources/Content/List/Endpoint.cs
```diff
@@ -89,29 +89,21 @@ private static string ApplyBookAndChapterFilter(string? bookCode, int? startChap
...
-                    WHERE [R].[Id] = [p1].[ResourceId] AND [p2].[EndVerseId] >= {startVerseId} AND [p2].[EndVerseId] <= {endVerseId})
-                OR
-                EXISTS (SELECT 1
-                    FROM [PassageResources] AS [p3]
-                    INNER JOIN [Passages] AS [p4] ON [p3].[PassageId] = [p4].[Id]
-                    WHERE [R].[Id] = [p3].[ResourceId] AND [p4].[StartVerseId] <= {startVerseId} AND [p4].[EndVerseId] >= {endVerseId})
+                    WHERE [R].[Id] = [p].[ResourceId]
+                          AND [p0].[StartVerseId] BETWEEN {startVerseId} AND {endVerseId}
+                          AND [p0].[EndVerseId] BETWEEN {startVerseId} AND {endVerseId}
+                          AND [p0].[StartVerseId] <= {startVerseId} AND [p0].[EndVerseId] >= {endVerseId})
```
NateMerritt, line 106: In my original suggestion I used `OR` here for these verse conditions instead of `AND` in order to keep functionality similar to the initial set of three `SELECT`s.  If I'm understanding correctly it looks like using `AND` won't work because the third condition can never be true if the first condition is true except for the equality case.

MrGrinst: 🤦 yep missed that

NateMerritt (approved):

PR #455, opened 2024-09-26
Bib 710 modify view edit page for community reviewer users

kaseywright:
These biggest change here was that I decided to add a separate endpoint for the community users to send their work to be reviewed for publishers. Due to the constraints of the user, it made more sense to create the new endpoint rather than modifying the existing `send-for-publisher` endpoint. Also, it was a good exercise in understanding the codebase, EF and FastEndpoints.

### DIFF
--- a/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
@@ -99,7 +99,11 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
         );
 
         if (isCommunityUser) {
-            await historyService.AddSnapshotHistoryAsync(newResourceContentVersion, user.Id, ResourceContentStatus.TranslationInProgress, ct);
+            await historyService.AddSnapshotHistoryAsync(
+                newResourceContentVersion, 
+                user.Id, 
+                ResourceContentStatus.TranslationInProgress, 
+                ct);
             await historyService.AddAssignedUserHistoryAsync(newResourceContentVersion, user.Id, user.Id, ct);
         }
 
--- a/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
@@ -19,7 +19,7 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
     public override async Task HandleAsync(Request request, CancellationToken ct)
     {
         var user = await userService.GetUserFromJwtAsync(ct);
-        var isCommunityUser = userService.HasPermission(PermissionName.CreateCommunityContent);
+        var isCommunityUser = !userService.HasPermission(PermissionName.CreateContent);
 
         if (isCommunityUser) {
             var isUserAssigned = await dbContext.ResourceContentVersions
--- a/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
@@ -100,6 +100,11 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
             ct
         );
 
+        if (isCommunityUser) {
+            await historyService.AddSnapshotHistoryAsync(newResourceContentVersion, user.Id, ResourceContentStatus.TranslationInProgress, ct);
+            await historyService.AddAssignedUserHistoryAsync(newResourceContentVersion, user.Id, user.Id, ct);
+        }
+
         await dbContext.SaveChangesAsync(ct);
         await SendOkAsync(new Response(newResourceContent.Id), ct);
     }
--- a/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
@@ -18,6 +18,18 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
 
     public override async Task HandleAsync(Request request, CancellationToken ct)
     {
+        var user = await userService.GetUserFromJwtAsync(ct);
+        var isCommunityUser = userService.HasPermission(PermissionName.CreateCommunityContent);
+
+        if (isCommunityUser) {
+            var isUserAssigned = await dbContext.ResourceContentVersions
+                .AnyAsync(x => x.AssignedUserId == user.Id, ct);
+
+            if (isUserAssigned) {
+                ThrowError("User can only create one translation at a time");
+            }
+        }
+
         var baseContent = await dbContext.ResourceContents.Where(x => x.Id == request.BaseContentId)
             .Include(x => x.Versions)
             .SingleOrDefaultAsync(ct);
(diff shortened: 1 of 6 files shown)


src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
```diff
@@ -18,6 +18,18 @@ public override void Configure()
 
     public override async Task HandleAsync(Request request, CancellationToken ct)
     {
+        var user = await userService.GetUserFromJwtAsync(ct);
+        var isCommunityUser = userService.HasPermission(PermissionName.CreateCommunityContent);
```
jwinston-bn, line 22: I think the absence of a permission would be better to determine this. Reason being that there could be situations where a user has extra permissions (say an admin type of user) and then they go into the wrong path. So I would check that they _don't_ have the `PermissionName.CreateContent` permission.

kaseywright: is there a situation when a user would not have the `CreateContent` permission but also should not have the ability to `CreateCommunityContent` ?

jwinston-bn: Because of the `Permissions(PermissionName.CreateContent, PermissionName.CreateCommunityContent);` in the configure you have to have at least one of them.

src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
```diff
@@ -62,20 +74,36 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
...
             Trusted = true,
             Versions = [newResourceContentVersion]
         };
 
         await dbContext.ResourceContents.AddAsync(newResourceContent, ct);
+        await historyService.AddStatusHistoryAsync(
+            newResourceContentVersion, 
+            isCommunityUser 
+                ? ResourceContentStatus.TranslationInProgress 
+                : ResourceContentStatus.TranslationNotStarted,
```
MrGrinst, line 98: Nit: we could do `newResourceContent.Status` here since the same ternary has already done the work above

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Resources.Content.SendForPublisherReview.Community;
+
+public class Endpoint(AquiferDbContext dbContext, IUserService userService, IResourceHistoryService historyService) 
+    : Endpoint<Request, Response>
+{
+    public override void Configure()
+    {
+        Post("/resources/content/{ContentId}/send-for-publisher-review-community", "/resources/content/send-for-publisher-review-community");
```
MrGrinst, line 15: I don't think we'll need the second arg here. That one is specifically for managers to use to bulk send items to publisher review and won't be needed for this one

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Request.cs
```diff
@@ -0,0 +1,6 @@
+namespace Aquifer.API.Endpoints.Resources.Content.SendForPublisherReview.Community;
+
+public class Request
+{
+    public int? ContentId { get; set; }
```
MrGrinst, line 5: This probably can be `int` not optional since once the second path in the endpoint is removed like my above comment, it'll always be part of the path.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
```diff
@@ -0,0 +1,60 @@
...
+        Permissions(PermissionName.SendReviewCommunityContent);
+    }
+
+    public override async Task HandleAsync(Request request, CancellationToken ct)
+    {
+        var user = await userService.GetUserFromJwtAsync(ct);
+        var contentVersion = await dbContext.ResourceContentVersions
+            .Where(x => x.ResourceContentId == request.ContentId && x.AssignedUserId == user.Id && x.IsDraft)
+            .Include(x => x.ResourceContent)
+            .SingleAsync(ct);
```
jwinston-bn, line 25: This will blow up if nothing is found. `Single` means there must be 1 and only 1 or it will blow up. `First` means there must be at least 1 and it will return the first 1, or it will blow up. `SingleOrDefault` and `FirstOrDefault` will allow a case where nothing is found. So we want `SingleOrDefaultAsync` here to say that we allow either no record or 1 and only 1 record.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        if (contentVersion is null)
+        {
+            ThrowError("Content not found");
```
jwinston-bn, line 29: This will return a 500 which isn't really right. I would `await SendNotFoundAsync(); return;` to return a 404 which implies nothing was found.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        if (contentVersion.ResourceContent.Status != ResourceContentStatus.TranslationInProgress)
+        {
+            ThrowError("Content not in correct status");
```
jwinston-bn, line 34: As above, I'd return a bad request here with this message, rather than blow up and force the 500.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        await historyService.AddAssignedUserHistoryAsync(
+            contentVersion,
+            null,
+            user.Id,
+            ct
+        );
+
+        await dbContext.SaveChangesAsync(ct);
+        await SendOkAsync(ct);
```
jwinston-bn, line 58: Would use `SendNoContentAsync` here.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
kaseywright: Does this not send a 400? The IDE context says `ThrowError` is:
> add a "GeneralError" to the validation failure list and send back a 400 bad request with error details immediately interrupting handler execution flow. if there are any vallidation failures, no execution will continue past this call.

src/Aquifer.API/Endpoints/Users/GetSelf/Response.cs
```diff
@@ -5,8 +5,9 @@ public class Response
     public int Id { get; set; }
     public required string Name { get; set; }
     public required IEnumerable<string> Permissions { get; set; }
-
     public required CompanyResponse Company { get; set; }
+    public int? LanguageId { get; set; }
+    public bool? HasAssignedContent { get; set; }
```
jwinston-bn, line 10: I think this will be confusing without context, and it's not really correct when talking about a non-community user. Most users will have content assigned to them, but this will be null. I'd also argue that null isn't really a valid state for this, since a user either does or doesn't have content assigned. So I would rename this to something like `bool CanBeAssignedContent` that should be true unless it's a community user with something already assigned.

src/Aquifer.API/Endpoints/Users/GetSelf/Endpoint.cs
```diff
@@ -20,9 +23,15 @@ public override async Task HandleAsync(CancellationToken ct)
             Id = user.Id,
             Name = $"{user.FirstName} {user.LastName}",
             Permissions = permissions,
-            Company = new CompanyResponse { Id = user.CompanyId }
+            Company = new CompanyResponse { Id = user.CompanyId },
+            LanguageId = user.LanguageId,
         };
 
+        if (userService.HasPermission(PermissionName.CreateCommunityContent)) {
```
jwinston-bn, line 30: Similar to create translation, I think the absence of the `CreateContent` permission should play a part here. I would keep the existing check for `CreateCommunityContent` but also add `&& !userService.HasPermission(PermissionName.CreateContent`.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
jwinston-bn: Okay, I assumed it was sending a 500 since it was an error, so sounds like it's equivalent to the 400 then.

jwinston-bn: I think the only thing to consider is if this would trigger a log to app insights that could be alerted. Because it's not really an error, so much as a state we're just not allowing, so it's not something that should be throwing an exception under the hood. It does throw a `ValidationFailureException`, so we should verify if this is the correct flow (the Validators might already do the exact same thing).

kaseywright: Yes, however, to me, if this is how the docs for fast endpoints says to return validation errors I would wonder why we would be alerting this in App Insights at all. It seems overly verbose for App Insights to alert on 400 errors. logging them, yes, but alerting, maybe not.

src/Aquifer.API/Endpoints/Users/GetSelf/Response.cs
kaseywright: This caused changed on the front end. So, there's a new PR for it there as well.

jwinston-bn (approved):

PR #458, opened 2024-09-27
Add Greek alignment data to public API.

NateMerritt:
The new endpoint is at `GET /bibles/1/alignments/greek` on the public API.

Some changes I made when compared to the internal API's `GET /bibles/1/alignments/greek/book/mat/chapter/1`:
* Allow fetching multiple chapters at once to be consistent with the public API's `GET /bibles/texts` endpoint.
* Allow fetching data for only a single Bible word if desired to avoid consumers having to fetch too much data depending upon the use case.

I encountered significant performance issues using EF and the existing raw SQL queries from the internal API route.  Given that I was opting to return more than one chapter this became a problem when fetching entire books at once.  I was able to refactor out redundant JOINs and prevent duplicate data from being returned by the server but it required adding in Dapper in order to fetch the values for many IDs via parameterization; I couldn't find a way to do either of the following with EF (a) pass many parameters from an array or (b) execute a read multiple query and use temp DB on the server.  Dapper supports both.  I did the former for this PR which has the best possibilities for future caching by Greek word ID and Greek sense ID.  The latter would probably be faster overall by storing IDs in temp DB instead of round-tripping them to our servers and back.

Still TODO if we decide we're ok with using Dapper: Figure out how to log SQL and timings like we get with EF core.

Here's timings I took with SQL Extended Events profiling (on the SQL server itself so my latency isn't included):
```
Example timing for Matthew 1 from *existing* internal API route:
* Base query: 42 ms
* Greek words query: 43 ms
* Greek senses query: 68 ms
Total: 153 ms

Example timing for Matthew 1 from *new* public API route:
* Fetch Greek Alignment source (name of Greek bible): 1 ms
* Fetch Bible Book summary data: 1 ms
* Fetch words with Greek alignment foreign keys: 83 ms
* Fetch Greek words from foreign keys: 3 ms
* Fetch Greek senses from foreign keys: 6 ms
Total: 94 ms

Example timing for Matthew (entire book) from *new* public API route:
* Fetch Greek Alignment source (name of Greek bible): 1 ms
* Fetch Bible Book summary data: 1 ms
* Fetch words with Greek alignment foreign keys: 353 ms
* Fetch Greek words from foreign keys in batches: 59 ms, 68 ms, 3 ms
* Fetch Greek senses from foreign keys in batches: 65 ms, 8 ms
Total: 558 ms

Example timing for Luke (entire book; longest book in NT so max response time) from *new* public API route :
* Fetch Greek Alignment source (name of Greek bible): 1 ms
* Fetch Bible Book summary data: 1 ms
* Fetch words with Greek alignment foreign keys: 613 ms
* Fetch Greek words from foreign keys in batches: 66 ms, 59 ms, 16 ms
* Fetch Greek senses from foreign keys in batches: 67 ms, 18 ms
Total: 841 ms
```
There are lots of server side in-memory caching possibilities for this data but I don't think we should look into that until we see these routes getting significant use.


Aquifer.Common.UnitTests/Aquifer.Common.UnitTests.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <PackageReference Include="xunit" Version="2.5.3" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\src\Aquifer.Common\Aquifer.Common.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Using Include="Xunit" />
```
NateMerritt, line 23: After talking with @jwinston-bn we decided to use Xunit over Nunit.  Other possibilities either don't have a lot of industry usage (MSTest) or are too new (Tunit).

Aquifer.Common.UnitTests/BibleWordIdentifierTests.cs
```diff
@@ -0,0 +1,346 @@
+using Aquifer.Common.Utilities;
+using Aquifer.Data.Enums;
+
+namespace Aquifer.Common.UnitTests;
+
+public sealed class BibleWordIdentiferTests
```
NateMerritt, line 6: The new `BibleWordIdentifier` class is needed so that I can easily create word identifiers and also parse word identifiers.  There are a decent number of methods on the new class with some complicated math so I decided to write an exhaustive suite of unit tests.   This didn't take too long because many tests are similar but just hit a different constructor overload.

src/Aquifer.API/Program.cs
```diff
@@ -23,8 +23,9 @@
     .AddOutputCache()
     .AddApplicationInsightsTelemetry()
     .AddSingleton<ITelemetryInitializer, RequestTelemetryInitializer>()
-    .AddDbContext<AquiferDbContext>(options =>
-        options.UseSqlServer(configuration?.ConnectionStrings.BiblioNexusDb, providerOptions => providerOptions.EnableRetryOnFailure(3)))
+    .AddDbContext<AquiferDbContext>(options => options
+        .UseSqlServer(configuration?.ConnectionStrings.BiblioNexusDb, providerOptions => providerOptions.EnableRetryOnFailure(3))
+        .EnableSensitiveDataLogging(sensitiveDataLoggingEnabled: builder.Environment.IsDevelopment()))
```
NateMerritt, line 28: Adding this line allows seeing parameters passed to SQL from EF in app logging but only in the Development environment.

src/Aquifer.Common/Utilities/AlignmentUtilities.cs
```diff
@@ -1,16 +0,0 @@
-﻿using Aquifer.Data.Enums;
```
NateMerritt, line 1: Logic was moved into to `BibleWordIdentifer` class.

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
```diff
@@ -0,0 +1,182 @@
...
+/// Format for Bible word identifiers is BBCCCVVVWWWS.
+/// Format for Greek word identifiers is BBCCCVVVWWW and thus this class should not be used with Greek data.
+/// Key:
+///   B: Book
+///   C: Chapter
+///   V: Verse
+///   W: Word
+///   S: Word segment
+/// </summary>
+public sealed class BibleWordIdentifier
```
NateMerritt, line 18: I normally tend towards POCOs and separate utility classes but it made sense to me to contain this complexity in one class that is well tested and easy to use.

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
```diff
@@ -0,0 +1,182 @@
...
+    public static BibleWordIdentifier GetUpperBoundOfBook(BookId bookId)
+    {
+        return new BibleWordIdentifier(bookId, _maxChapter, _maxVerse, _maxWord, _maxWordSegment);
+    }
+
+    public static BibleWordIdentifier GetUpperBoundOfChapter(BookId bookId, int chapter)
+    {
+        return new BibleWordIdentifier(bookId, chapter, _maxVerse, _maxWord, _maxWordSegment);
+    }
+    public static BibleWordIdentifier GetUpperBoundOfVerse(BookId bookId, int chapter, int verse)
```
NateMerritt, line 136:
```suggestion

    public static BibleWordIdentifier GetUpperBoundOfVerse(BookId bookId, int chapter, int verse)
```

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,228 @@
...
+    private const string GreekWordsQuery = """
+        SELECT bvw.Id AS EnglishWordId, gntw.Id AS GreekNewTestamentWordId, gw.[Text] AS GreekWord
+        FROM BibleVersionWords bvw
+        LEFT JOIN BibleVersionWordGroupWords bvwgw ON bvwgw.BibleVersionWordId = bvw.Id
+        LEFT JOIN NewTestamentAlignments nta ON nta.BibleVersionWordGroupId = bvwgw.BibleVersionWordGroupId
+        LEFT JOIN GreekNewTestamentWordGroupWords gntwgw ON gntwgw.GreekNewTestamentWordGroupId = nta.GreekNewTestamentWordGroupId
+        LEFT JOIN GreekNewTestamentWords gntw ON gntw.Id = gntwgw.GreekNewTestamentWordId
+        LEFT JOIN GreekWords gw ON gntw.GreekWordId = gw.Id
+        WHERE BibleId = @BibleId AND bvw.WordIdentifier BETWEEN @LowerBounds AND @UpperBounds AND gntw.Id IS NOT NULL
+        """;
```
NateMerritt, line 40: I'm having trouble with the performance of this query (see PR description).  This PR is blocked until I can address it.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,228 @@
...
+            LEFT JOIN GreekNewTestamentWordSenses gntws ON gntws.GreekNewTestamentWordId = gntw.Id
+            LEFT JOIN GreekSenses gs ON gs.Id = gntws.GreekSenseId
+            LEFT JOIN GreekSenseGlosses gsg ON gsg.GreekSenseId = gs.Id
+            WHERE BibleId = @BibleId AND bvw.WordIdentifier BETWEEN @LowerBounds AND @UpperBounds AND gntw.Id IS NOT NULL AND gs.DefinitionShort IS NOT NULL
+        )
+        SELECT GreekNewTestamentWordId, Definition, STRING_AGG(Gloss, '||') AS Glosses
+        FROM CTE
+        WHERE RowNum = 1
+        GROUP BY GreekNewTestamentWordId, Definition
+        """;
```
NateMerritt, line 62: I'm having trouble with the performance of this query (see PR description).  This PR is blocked until I can address it.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,228 @@
...
+        var response = new Response
+        {
+            BibleId = bookData.BibleId,
+            BibleName = bookData.BibleName,
+            BibleAbbreviation = bookData.BibleAbbreviation,
+            GreekBibleAbbreviation = greekAlignmentSource.Name,
+            BookCode = bookData.BookCode,
+            BookName = bookData.BookName,
+            Chapters = baseQueryResults
+                .GroupBy(r => r.BibleWordIdentifier.Chapter)
```
NateMerritt, line 151: The groupings here for chapter and below for verse are what motivated me to split out a bible word identifier into its components in the new `BibleWordIdentifier` class.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,228 @@
+﻿using Aquifer.Common.Utilities;
```
NateMerritt, line 1: Much of this is copy/pasted from https://github.com/BiblioNexusStudio/aquifer-server/blob/master/src/Aquifer.API/Endpoints/Bibles/Alignments/Greek/Book/Chapter/Endpoint.cs as well as the `Request`, `Response`, and `Validator`.  I did modify it to expand out to full chapters or down to only words, though.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
NateMerritt: Really weird.  Removing one level of indentation (four spaces) under the SELECT fixed it.  I think that the whitespace must somehow impact the query plan chosen by SQL Server.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,272 @@
...
+            })
+            .FirstOrDefaultAsync(ct);
+
+        if (bookData == null)
+        {
+            await SendNotFoundAsync(ct);
+            return;
+        }
+
+        var greekAlignmentSource = await dbContext.Database.GetDbConnection()
```
NateMerritt, line 121: This query and the ones below are using Dapper (only the last two needed it but I did it for all of them for consistency).  I propose that instead of using raw SQL queries with EF we use Dapper instead for the scenarios where we want to pass straight SQL.  Dapper's usage is similar to the raw SQL EF method call but it provides much more power when we want to improve performance.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,272 @@
...
+                        {
+                            lowerBounds,
+                            upperBounds,
+                            bibleId = request.BibleId,
+                        },
+                        cancellationToken: ct)))
+            .ToList();
+
+        List<GreekWordResult> greekWordResults = [];
+        foreach (var batch in bibleText.Select(r => r.GreekWordId).Distinct().Order().Chunk(size: 2000))
```
NateMerritt, line 144: The base query now only retrieves `GreekWordId`s and this second query here batches over those to fetch the actual Greek word data.  SQL has a 2100 parameter limit so I arbitrarily chose 2000 as the batch size.

This is powerful because we could cache all of this data by ID and only fetch data from the DB for future requests if those Greek Words are not yet in the cache.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Request.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// Optional end word in the end verse.
+    /// </summary>
+    [DefaultValue(999)]
+    public int EndWord { get; init; } = 999;
+
+    /// <summary>
+    /// Optional. If <c>true</c> then any available Greek word sense information for each Greek word (definition and glosses) will be included in the response.
+    /// </summary>
+    [DefaultValue(false)]
+    public bool ShouldReturnSenseData { get; init; }
```
NateMerritt, line 58: This is also a new addition.  By default the API won't return Greek word definitions and glosses; this data must be explicitly requested.

Aquifer.Common.UnitTests/BibleWordIdentifierTests.cs
jwinston-bn: Typo in Identifier

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,275 @@
...
+                                            Word = word.EnglishWord,
+                                            NextWordIsInGroup = bibleVersionWordIdBookendsInBibleVersionWordGroupByBibleVersionWordGroupId[word.BibleVersionWordGroupId].Last != word.Id,
+                                            GreekWords =
+                                                bibleVersionWordIdBookendsInBibleVersionWordGroupByBibleVersionWordGroupId[word.BibleVersionWordGroupId].First == word.Id
+                                                ? wordGrouping
+                                                    .GroupBy(r => r.GreekWordId)
+                                                    .Select(greekWordGrouping =>
+                                                    {
+                                                        var greekWordResult = greekWordResultById[greekWordGrouping.Key];
+                                                        var greekSenseResults = greekWordGrouping
```
jwinston-bn, line 206: This local variable hides the outer greekSenseResults above, recommend changing the name of one of them.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,275 @@
...
+            foreach (var batch in bibleText.Select(r => r.GreekSenseId).Distinct().Order().Chunk(size: 2000))
+            {
+                greekSenseResults.AddRange(await dbContext.Database.GetDbConnection()
+                    .QueryAsync<GreekSenseResult>(new CommandDefinition(GreekSensesQuery, new { greekSenseIds = batch }, cancellationToken: ct)));
+            }
+        }
+
+        var greekSenseResultById = greekSenseResults.ToLookup(r => r.Id);
+
+        var response = new Response
```
jwinston-bn, line 165: I think this whole thing is really hard to read. I would break this into multiple private methods to clearly state what each part is building, and then that will also help with the excessive nesting.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,275 @@
...
+        Description(d => d.ProducesProblemFE());
+        Summary(s =>
+        {
+            s.Summary = "Gets a Bible's Greek alignment information.";
+            s.Description =
+                "For a given Bible and book of the Bible, returns the Bible's text along with associated Greek alignment information and Greek sense data for each verse, similar to a reverse interlinear Bible. Data is returned for all verses within the chapter and verse parameters.";
+        });
+    }
+
+    public override async Task HandleAsync(Request request, CancellationToken ct)
```
jwinston-bn, line 93: Overall I think this method is very long, and I would advocate for breaking it up into some smaller private methods (specifically the response building as separately commented below)

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -0,0 +1,275 @@
...
+        public long WordIdentifier { get; set; }
+        public string EnglishWord { get; set; } = null!;
+        public bool IsPunctuation { get; set; }
+        public int BibleVersionWordGroupId { get; set; }
+        public int GreekSenseId { get; set; }
+        public int GreekWordId { get; set; }
+
+        public BibleWordIdentifier BibleWordIdentifier => _bibleWordIdentifier ??= new BibleWordIdentifier(WordIdentifier);
+
+        private BibleWordIdentifier? _bibleWordIdentifier = null;
```
jwinston-bn, line 256: Rider says this is a redundant instantiation.

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
```diff
@@ -0,0 +1,195 @@
...
+    {
+        return WordIdentifier.ToString("D12");
+    }
+
+    private static long CalculateWordIdentifier(int book, int chapter, int verse, int word, int wordSegment)
+    {
+        return (book * _bookMultiplier) + (chapter * _chapterMultiplier) + (verse * _verseMultiplier) + (word * _wordMultiplier) + wordSegment;
+    }
+
+    private static readonly BookId _maxBookId = Enum.GetValues(typeof(BookId)).Cast<BookId>().Max();
```
jwinston-bn, line 174: Same message about the consts too.
image

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
```diff
@@ -0,0 +1,195 @@
+﻿using Aquifer.Data.Enums;
+
+namespace Aquifer.Common.Utilities;
+
+/// <summary>
+/// Only use with Bible word identifiers, NOT Greek word identifers!!!
```
jwinston-bn, line 6: Identifiers typo again

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// <summary>
+    /// Creates a fully specified <see cref="BibleWordIdentifier"/>.
+    /// </summary>
+    public BibleWordIdentifier(BookId bookId, int chapter, int verse, int word, int wordSegment)
+    {
+        if (bookId <= BookId.None || bookId > _maxBookId)
+        {
+            throw new ArgumentOutOfRangeException($"Invalid {nameof(BookId)} value for calculating word identifiers: \"{bookId}\".", nameof(bookId));
```
jwinston-bn, line 95: Parameters are reversed

Aquifer.Common.UnitTests/BibleWordIdentifierTests.cs
NateMerritt: Thanks.  I used to get IDE warnings for the class name not matching the file name but maybe that was Resharper, not VS.

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
NateMerritt: Where is this rule defined?  I'm not seeing it in the `.editorconfig` so I'm not sure where Rider is getting that from.

That's a surprising naming convention because public properties are also Pascal case and they would look the same.  I've usually done public `const` values in Pascal and private const values in a different convention to help distinguish between them visually.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Request.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Book code based off USFM book identifier (e.g. GEN, EXO, etc.). Can get a list of available books and identifiers
+    /// from the /bibles/books endpoint. This parameter is required.
+    /// </summary>
+    public string BookCode { get; init; } = null!;
+
+    /// <summary>
+    /// Optional start chapter in the book.
+    /// </summary>
+    [DefaultValue(1)]
```
MrGrinst, line 21: This kinda goes back to earlier discussions, but I'd prefer not having the attribute here since that's what is used to generate the OpenAPI spec and it feels like it's exposing our internal implementation choices by having them (especially the `999` values below). If we remove the attribute it should keep working as it does now but make the API docs look cleaner. Thoughts?

src/Aquifer.Common/Utilities/BibleWordIdentifier.cs
NateMerritt: I updated my PR to follow Rider's suggested styling and will look into updating the `.editorconfig` to enforce it when I get some downtime.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
NateMerritt: I split out a bunch of methods, both for data access (which will now be easier to move to a central location in order for the internal API to get at the performance optimizations I made) and for the response creation.  The response creation might be a little harder to understand as a whole now with splitting out the methods but each method is hopefully easier to understand individually.  I also renamed some things for clarity.

src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Request.cs
NateMerritt: @MrGrinst, this is a byproduct of not using nullable for these start/end request parameters.  @jwinston-bn asked for non-nullable on the GET Bible Texts route with defaults of `1` and `999` for start and end values, respectively.  I noticed later that doing this prevented the documentation page from showing these values as "optional."  We were able to address that by specifying `DefaultValueAttribute` because [our documentation creation code](https://github.com/BiblioNexusStudio/aquifer-server/blob/214ca06cd5a409bb243c8db856433037c93eda7d/src/Aquifer.Public.API/OpenApi/DefaultParameterOperationProcessor.cs#L34-L38) checks for that attribute and  lists the parameter as optional (it also seems that the NSwag itself also checks for nullable types and makes them optional) .  Here's the [original fix](https://github.com/BiblioNexusStudio/aquifer-server/pull/450) for that route.  I then copied the same request parameters to this Greek alignment text route for consistency across the Public API surface.

We _could_ specify `DefaultValue(null)` instead (which @jwinston-bn pointed out in a Slack DM to me).  My response from that Slack thread:

> I agree that `DefaultValue(null)` has the best result for the consumer in this case.  However, I don't like the potential confusion for future developers because that's not actually the default value.

> There's a `RequiredAttribute` (https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.requiredattribute?view=net-8.0) from the same namespace as `DefaultValueAttribute` that we could also reference from the `DefaultParameterOperationProcessor` code but that would likely require slapping that `Required` attribute on all required properties and using it exclusively.  It would be a bit less magical than the current implementation because it makes it more clear why that text appears.

> There might be danger of that `RequiredAttribute` causing deserialization headaches so we could also just make our own `OptionalAttribute` and use that in addition to `DefaultValue` to determine if the "required" text should show.

We eventually decided to table it and just ship with the actual default value in the attribute.

All that said, my preferred solution would be nullable request properties when the route structure allows them to be optional, possibly with the addition of a `RequiredAttribute` for specifying query string parameters that we require for the route to function.  Then we'd get the correct documentation for optional properties without needing an attribute and we could keep the actual default value in our internal implementation.  We'd also know when the user isn't supplying values for validation purposes which helps solve edge cases.

jwinston-bn (approved):

PR #460, opened 2024-10-01
Turn off EF Tracking for all read-only queries.

NateMerritt:
EF tracking is on by default.  After fetching an entity from the DB, tracking incurs a performance cost in order to track updates to that entity so that the changes can be saved back to the DB when `await dbContext.SaveChangesAsync(ct)` is called.  If no save back to the DB is necessary (for read-only routes) then it is advisable to turn off tracking for that query.  See https://learn.microsoft.com/en-us/ef/core/querying/tracking and https://medium.com/@anyanwuraphaelc/benefits-of-asnotracking-in-entity-framework-core-a-guide-to-improved-performance-186ed44a5eb7.

Changes:
Public API:
* Turn off tracking globally for all EF commands; the Public API is read-only.
* 
Internal API:
* Turn off tracking globally for all EF commands.
* Add `.AsTracking()` to all write routes (Post/Put/Patch/Delete/etc.).  In some cases this required whitespace/alignment changes.

Consequences:
* If anyone tries to save changes to the `DbContext` on the Public API in future routes it will not work by default.  They can, however, use `AsTracking()` to enable it for that one query.
* Anyone working on any update route in the Internal API will need to explicitly use `AsTracking()` for any new queries.

### DIFF
--- a/src/Aquifer.API/Endpoints/Comments/Create/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Comments/Create/Endpoint.cs
@@ -16,7 +16,10 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService) : En
 
     public override async Task HandleAsync(Request req, CancellationToken ct)
     {
-        var thread = await dbContext.CommentThreads.SingleOrDefaultAsync(x => x.Id == req.ThreadId && !x.Resolved, ct);
+        var thread = await dbContext.CommentThreads
+            .AsTracking()
+            .SingleOrDefaultAsync(x => x.Id == req.ThreadId && !x.Resolved, ct);
+
         EndpointHelpers.ThrowErrorIfNull<Request>(thread, x => x.ThreadId, "No open thread found for id", 404);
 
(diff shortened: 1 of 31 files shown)


MrGrinst (dismissed):

NateMerritt:
Update: @jwinston-bn asked me to reverse the internal API changes in this PR to instead globally apply `NoTracking` behavior and then to explicitly use `AsTracking()` for all DB write queries.

NateMerritt:
Note that some queries that now have `.NoTracking()` don't actually need them.  I opted to apply no tracking on every query in a class that dealt with write operations in order to have a clear and easy rule for future development.

There are only a few exceptions to this broad rule such as https://github.com/BiblioNexusStudio/aquifer-server/blob/d65bcf38fe15e716900e110cead067116d68cd2a/src/Aquifer.Data/EventHandlers/ResourceStatusChangeHandler.cs#L49-L55 which are very clear from the context that no tracking is needed.

jwinston-bn (approved):

PR #462, opened 2024-10-03
split AssociatedResources into its own entity and add suggested indexes

MrGrinst:
This adds an entity for associated resources as an actual entity class for more control over the `AssociatedResources` table. It does so mainly to allow Azure-suggested indexes to be added. It could potentially help improve performance as well, as we saw with EF-generated `ProjectResourceContents` queries.

### DIFF
--- a/src/Aquifer.API/Modules/Passages/PassagesModule.cs
+++ b/src/Aquifer.API/Modules/Passages/PassagesModule.cs
@@ -1,5 +1,4 @@
 using Aquifer.API.Common;
-using Aquifer.API.Modules.Resources;
 using Aquifer.Common.Utilities;
 using Aquifer.Data;
 using Microsoft.AspNetCore.Http.HttpResults;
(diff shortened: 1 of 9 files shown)


src/Aquifer.API/Modules/Passages/PassagesModule.cs
```diff
@@ -60,108 +58,4 @@ private async Task<Results<Ok<List<PassagesByBookResponse>>, NotFound>> GetPassa
 
         return TypedResults.Ok(passagesByBook);
     }
-
-    private async Task<Results<Ok<PassageWithResourceItemsResponse>, NotFound>> GetPassageDetailsForLanguage(
```
MrGrinst, line 64: This endpoint was used a long time ago by Well but is no longer used. Had forgotten to remove it before.

jwinston-bn (dismissed):

jwinston-bn (approved):

PR #465, opened 2024-10-03
Update Response to accept message string

kaseywright:
Update `ThrowError` to `SendAsync`

### DIFF
--- a/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
@@ -20,10 +20,6 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
     {
         var user = await userService.GetUserFromJwtAsync(ct);
         var isCommunityUser = !userService.HasPermission(PermissionName.CreateContent);
-        var response = new Response
-        {
-            ResourceContentId = 0,
-        };
 
         if (isCommunityUser) {
             var isUserAssigned = await dbContext.ResourceContentVersions
@@ -31,8 +27,8 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
                 .AnyAsync(x => x.AssignedUserId == user.Id, ct);
 
             if (isUserAssigned) {
-                response.Message = "User can only create one translation at a time";
-                await SendAsync(response, 400, ct);
+                AddError("User can only create one translation at a time");
+                await SendErrorsAsync(cancellation: ct);
                 return;
             }
         }
@@ -119,8 +115,6 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
 
         await dbContext.SaveChangesAsync(ct);
 
-        response.ResourceContentId = newResourceContent.Id;
-
-        await SendOkAsync(response, ct);
+        await SendOkAsync(new Response(newResourceContent.Id), ct);
     }
 }
--- a/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
@@ -20,13 +20,19 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService, IRes
     {
         var user = await userService.GetUserFromJwtAsync(ct);
         var isCommunityUser = !userService.HasPermission(PermissionName.CreateContent);
+        var response = new Response
+        {
+            ResourceContentId = 0,
+        };
 
         if (isCommunityUser) {
             var isUserAssigned = await dbContext.ResourceContentVersions
                 .AnyAsync(x => x.AssignedUserId == user.Id, ct);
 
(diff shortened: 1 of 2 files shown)


src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
```diff
@@ -20,14 +20,20 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
...
+        };
 
         if (isCommunityUser) {
             var isUserAssigned = await dbContext.ResourceContentVersions
                 .AsTracking()
                 .AnyAsync(x => x.AssignedUserId == user.Id, ct);
 
             if (isUserAssigned) {
-                ThrowError("User can only create one translation at a time");
+                response.Message = "User can only create one translation at a time";
```
MrGrinst, line 34: Was there an issue with using the `ThrowError` before? We haven't used this pattern anywhere else and I'd be hesitant to break out of it for this case, especially since FastEndpoints already takes the error from `ThrowError` and formats it for us like:
```
{
	"statusCode": 400,
	"message": "One or more errors occurred!",
	"errors": {
		"generalErrors": ["User can only create one translation at a time"]
	}
}
```

kaseywright: As per Jesse, 'we don't need to be throwing errors on that endpoint for trying to start a second translation'.
So, really, the issue on that particular situation is that we don't need that level of verbosity with error reporting - which makes sense. 
I tend to agree as far as following the set paradigm for consistency but I can also see not wanting the error report every time someone clicks translate in a new tab as well.

MrGrinst: Huh, I guess my mindset is:
1) The error shouldn't ever get alerted on since we're swallowing it on the frontend
2) It _feels_ like it's an actual error. The data is in a different state than the user expects and it's an edge case. Granted, we assume people will regularly run into it, but that's no different from things like filesystem APIs that have regularly expected error types like `FileNotFoundException`.

If we're for sure not ok with making it an error then I wonder if having a boolean of `UserAlreadyAssignedContent` would make simplify things a bit.

kaseywright: Totally reasonable. I'll defer to whatever you guys decide on this as I will be ooo until Tuesday.

jwinston-bn: The backend gets the error logged because `ThrowError` does in fact throw an exception to stop things. This doesn't feel like an exceptional situation to me, because it's expected to happen, and while a minor thing, there is more overhead to throwing an exception than just sending back the 400 directly. The caller cares that they were already assigned, and they get that message back. On the back-end I don't think we care.

src/Aquifer.API/Endpoints/Resources/Content/CreateTranslation/Endpoint.cs
jwinston-bn, line 1: I think we can undo all the changes here in favor of just replacing the `ThrowError` on line 30 with

```csharp
await SendStringAsync("User can only create one translation at a time", 400, cancellation: ct);
return;
```

kaseywright: Returning `SendStringAsync` will cause the front end to break out of using the `isApiErrorWithMessage` error handling pattern we already have in place. Then, we'll have to write custom error detection for this one use case on the frontend. That feels bad to create a one-off api response handle when we have so many generic handlers in place already.

jwinston-bn: This is still returning a 400. Wouldn't that trigger line 65 of `http-service.ts`?

kaseywright: Yes, but we are further handling the ApiErrors in `http-errors.ts`. Specifically, the `isApiErrorWithMessage` fn when we expect the error to have a message - which is what that string is. Only, this way, we are not passing the string as  an expected `message` body element. So, we end up having to change the way we catch the error for this response on this endpoint.

jwinston-bn: What about this then?

```csharp
AddError("User can only create one translation at a time");
await SendErrorsAsync(cancellation: ct);
return;
```

kaseywright: isn't that going to just trigger the same error reporting that `ThrowError` does by
> adds a "GeneralError" to the current list of validation failures

jwinston-bn: I wouldn't think so. Going through the code it seems to just add it to a list. It works with `ThrowIfAnyErrors`.

jwinston-bn (approved):

PR #466, opened 2024-10-04
Add missing indices for greek alignment query

NateMerritt:
Add missing index needed for [this JOIN](https://github.com/BiblioNexusStudio/aquifer-server/blob/ef30f99f626ae0db2ccb389b1c8ee3fe96760b9f/src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs#L249).  It will also benefit the [older endpoint](https://github.com/BiblioNexusStudio/aquifer-server/blob/ef30f99f626ae0db2ccb389b1c8ee3fe96760b9f/src/Aquifer.API/Endpoints/Bibles/Alignments/Greek/Book/Chapter/Endpoint.cs#L23).

### DIFF
--- a/src/Aquifer.Data/Entities/BibleVersionWordGroupWordEntity.cs
+++ b/src/Aquifer.Data/Entities/BibleVersionWordGroupWordEntity.cs
@@ -2,7 +2,10 @@
 using Microsoft.EntityFrameworkCore;
 
 namespace Aquifer.Data.Entities;
+
 [PrimaryKey(nameof(BibleVersionWordGroupId), nameof(BibleVersionWordId))]
+[Index(nameof(BibleVersionWordId))]
+
 public class BibleVersionWordGroupWordEntity
 {
     public int BibleVersionWordGroupId { get; set; }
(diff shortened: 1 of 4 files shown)


src/Aquifer.Data/Entities/BibleVersionWordGroupWordEntity.cs
```diff
@@ -2,7 +2,10 @@
 using Microsoft.EntityFrameworkCore;
 
 namespace Aquifer.Data.Entities;
+
 [PrimaryKey(nameof(BibleVersionWordGroupId), nameof(BibleVersionWordId))]
+[Index(nameof(BibleVersionWordId))]
+
```
jwinston-bn, line 8: Nitpick since we were talking about consistency yesterday, we generally don't have an extra line between object and attribute.

NateMerritt: Agreed, this was an accident.  I'm going to turn off the VS "references" and code lens features so that I can see what is actual spacing and what is the IDE showing me "helpful" things that take up what looks like a line of code.  I had them off at my last job but figured I'd try them out again.

PR #470, opened 2024-10-04
fix language id bug during project creation

MrGrinst:


src/Aquifer.API/Endpoints/Projects/Create/Endpoint.cs
```diff
@@ -204,7 +203,7 @@ private async Task<List<ResourceContentEntity>> CreateOrFindTranslationResourceC
 
         foreach (var resourceContent in englishOrLanguageResourceContents)
         {
-            if (languageId == Constants.EnglishLanguageId)
+            if (resourceContent.LanguageId == Constants.EnglishLanguageId)
```
MrGrinst, line 206: This is the part that changed

NateMerritt: This fixes [this change](https://github.com/BiblioNexusStudio/aquifer-server/commit/214ca06cd5a409bb243c8db856433037c93eda7d#diff-d76c8ba163ce8d55c165c0268758e98df3121a336ef211c242a45dd6f6bad2e1L196).  Looks good.

NateMerritt (approved):

jwinston-bn (approved):

PR #471, opened 2024-10-04
Improve public API resource search route performance.

NateMerritt:
This improves performance of basic query searching by a factor of 2-3.  Most of the improvement is from changing the ordering but removing the language JOIN also shaved off some time.  See comments for details on both of those.  See also https://biblionexusworkspace.slack.com/archives/C05D8BE3KUY/p1728062517872939 for some additional ideas for future improvements.

Some data from QA DB profiling.  Searching for "Christ" (https://qa.api.aquifer.bible/resources/search?languageCode=eng&query=christ&offset=0&limit=100&api-key={apiKey}) takes the following duration:
* Getting the total: 1.012 seconds
* Getting the first 100 results (there are only 31): 3.782 seconds

With these changes applied:
* Getting the total: 0.426 seconds
* Getting the first 100 results (there are only 31): 1.700 seconds

### DIFF
--- a/src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Endpoint.cs
+++ b/src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Endpoint.cs
@@ -1,3 +1,4 @@
+using Aquifer.Common.Services.Caching;
 using Aquifer.Common.Utilities;
 using Aquifer.Data;
 using Aquifer.Data.Entities;
@@ -7,7 +8,7 @@ using Microsoft.EntityFrameworkCore;
 
 namespace Aquifer.Public.API.Endpoints.Resources.Search.GetResources;
 
-public class Endpoint(AquiferDbContext _dbContext) : Endpoint<Request, Response>
+public class Endpoint(AquiferDbContext _dbContext, ICachingLanguageService _cachingLanguageService) : Endpoint<Request, Response>
 {
     public override void Configure()
     {
@@ -36,7 +37,7 @@ public class Endpoint(AquiferDbContext _dbContext) : Endpoint<Request, Response>
 
     public override async Task HandleAsync(Request req, CancellationToken ct)
     {
-        var query = GetQuery(req);
+        var query = await GetQueryAsync(req, ct);
         var totalCount = await GetTotalResourceCountAsync(req, query, ct);
 
         if (totalCount == 0)
@@ -51,28 +52,32 @@ public class Endpoint(AquiferDbContext _dbContext) : Endpoint<Request, Response>
         await SendOkAsync(response, ct);
     }
 
-    private IQueryable<ResourceContentVersionEntity> GetQuery(Request req)
+    private async Task<IQueryable<ResourceContentVersionEntity>> GetQueryAsync(Request req, CancellationToken ct)
     {
+        var languageId = req.LanguageCode is not null
+            ? await _cachingLanguageService.GetLanguageIdAsync(req.LanguageCode, ct) ?? 0
+            : req.LanguageId;
+
         var (startVerseId, endVerseId) = req.BookCode is null
             ? ((int?)null, (int?)null)
             : BibleUtilities.GetVerseIds(req.BookCode, req.StartChapter, req.EndChapter, req.StartVerse, req.EndVerse);
 
-        return _dbContext.ResourceContentVersions.Where(x =>
-            x.IsPublished &&
-            x.ResourceContent.Resource.ParentResource.Enabled &&
-            (req.Query == null || x.DisplayName.Contains(req.Query) || x.ResourceContent.Resource.EnglishLabel.Contains(req.Query)) &&
-            (startVerseId == null ||
-             x.ResourceContent.Resource.VerseResources.Any(vr =>
-                 vr.VerseId >= startVerseId && vr.VerseId <= endVerseId) ||
-             x.ResourceContent.Resource.PassageResources.Any(pr =>
-                 (pr.Passage.StartVerseId >= startVerseId && pr.Passage.StartVerseId <= endVerseId) ||
-                 (pr.Passage.EndVerseId >= startVerseId && pr.Passage.EndVerseId <= endVerseId) ||
-                 (pr.Passage.StartVerseId <= startVerseId && pr.Passage.EndVerseId >= endVerseId))) &&
-            (req.ResourceType == default || x.ResourceContent.Resource.ParentResource.ResourceType == req.ResourceType) &&
-            (req.ResourceCollectionCode == null ||
-             x.ResourceContent.Resource.ParentResource.Code.ToLower() == req.ResourceCollectionCode.ToLower()
-            ) &&
-            (x.ResourceContent.LanguageId == req.LanguageId || x.ResourceContent.Language.ISO6393Code == req.LanguageCode));
+        return _dbContext.ResourceContentVersions
+            .Where(x =>
+                x.IsPublished &&
+                x.ResourceContent.Resource.ParentResource.Enabled &&
+                (req.Query == null || x.DisplayName.Contains(req.Query) || x.ResourceContent.Resource.EnglishLabel.Contains(req.Query)) &&
+                (startVerseId == null ||
+                 x.ResourceContent.Resource.VerseResources.Any(vr =>
+                     vr.VerseId >= startVerseId && vr.VerseId <= endVerseId) ||
+                 x.ResourceContent.Resource.PassageResources.Any(pr =>
+                     (pr.Passage.StartVerseId >= startVerseId && pr.Passage.StartVerseId <= endVerseId) ||
+                     (pr.Passage.EndVerseId >= startVerseId && pr.Passage.EndVerseId <= endVerseId) ||
+                     (pr.Passage.StartVerseId <= startVerseId && pr.Passage.EndVerseId >= endVerseId))) &&
+                (req.ResourceType == default || x.ResourceContent.Resource.ParentResource.ResourceType == req.ResourceType) &&
+                (req.ResourceCollectionCode == null ||
+                    x.ResourceContent.Resource.ParentResource.Code.ToLower() == req.ResourceCollectionCode.ToLower()) &&
+                x.ResourceContent.LanguageId == languageId);
     }
 
     private async Task<int> GetTotalResourceCountAsync(Request req, IQueryable<ResourceContentVersionEntity> query, CancellationToken ct)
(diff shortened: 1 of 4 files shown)


src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Endpoint.cs
```diff
@@ -86,19 +91,23 @@ private async Task<int> GetTotalResourceCountAsync(Request req, IQueryable<Resou
...
+    private async Task<List<ResponseContent>> GetResourcesAsync(
+        Request req,
         IQueryable<ResourceContentVersionEntity> query,
         CancellationToken ct)
     {
-        var resources = await query.OrderBy(x => x.ResourceContent.Resource.EnglishLabel)
+        var languageCodeByIdMap = await _cachingLanguageService.GetLanguageCodeByIdMapAsync(ct);
+
+        var resources = await query
+            .OrderBy(x => x.ResourceContent.Resource.Id)
```
NateMerritt, line 102: This one line change is the biggest win in this PR.  However, it is a breaking change because the data returned from the API will now be sorted by the Resource ID (visible in the response) instead of by the Resource's English Label.  I think this makes sense from a consumer perspective even in the response.  From a DB perspective this is _much_ faster because it can just return the clustered index data instead of sorting in memory.

MrGrinst: I'm not sure the breaking change would be acceptable from our users' perspective. I'm pretty sure at least two different projects are using our endpoint and I'm not sure if they're sorting data on their end. Jesse may have a better read on that. We may in fact want to update to respect the `?Resourcses.SortOrder` column. If we were to add an index on `SortOrder` and `EnglishLabel` would that prevent the in memory sort?

NateMerritt: It might.  We could try that if this change is unacceptable for consumers.

src/Aquifer.Common/Services/CachingLanguageService.cs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public async Task<int?> GetLanguageIdAsync(string languageCode, CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(languageCode);
+
+        return (await GetLanguageIdByCodeMapAsync(ct))
+            .TryGetValue(languageCode, out var languageId)
+                ? languageId
+                : null;
```
jwinston-bn, line 38:
```suggestion
return (await GetLanguageCodeByIdMapAsync(ct)).GetValueOrDefault(languageId);
```

src/Aquifer.Common/Services/CachingLanguageService.cs
```diff
@@ -0,0 +1,75 @@
...
+                .Select(l => new LanguageIdentifier(l.Id, l.ISO6393Code))
+                .ToListAsync(ct);
+        }
+    }
+
+    private record LanguageIdentifier(int Id, string Code);
+
+    private const string LanguageDictionariesCacheKey = $"{nameof(CachingLanguageService)}:LanguageDictionaries";
+
+    private static readonly TimeSpan s_cacheLifetime = TimeSpan.FromMinutes(30);
```
jwinston-bn, line 74: I generally see private variables like this at the top of a file (I think it's helpful for a user to quickly see what defaults are being set for the rest of the file). Thoughts?

src/Aquifer.Common/Services/CachingLanguageService.cs
```diff
@@ -0,0 +1,75 @@
...
+                : null;
+    }
+
+    public async Task<IReadOnlyDictionary<string, int>> GetLanguageIdByCodeMapAsync(CancellationToken ct)
+    {
+        return (await GetDictionariesFromCacheAsync(ct))
+            .LanguageIdByCodeMap;
+    }
+
+    private async Task<(IReadOnlyDictionary<int, string> LanguageCodeByIdMap, IReadOnlyDictionary<string, int> LanguageIdByCodeMap)> GetDictionariesFromCacheAsync(CancellationToken ct)
```
jwinston-bn, line 47: Couple lines here going against the editorconfig's set length limit. Again, I'm okay with having a discussion with the team around increasing that, but at this time all of our projects have the 140 limit and I think we should stick to that for consistency.

src/Aquifer.Common/Services/CachingLanguageService.cs
```diff
@@ -0,0 +1,75 @@
...
+                    (LanguageCodeById: languageIdentifiers.ToDictionary(li => li.Id, li => li.Code),
+                    LanguageIdByCode: languageIdentifiers.ToDictionary(li => li.Code, li => li.Id));
+            });
+
+        static async Task<IReadOnlyList<LanguageIdentifier>> GetLanguageIdentifiersFromDatabaseAsync(AquiferDbContext dbContext, CancellationToken ct)
+        {
+            return await dbContext.Languages
+                .Select(l => new LanguageIdentifier(l.Id, l.ISO6393Code))
+                .ToListAsync(ct);
+        }
```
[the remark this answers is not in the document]
jwinston-bn: Also, not sure if this does anything extra for the performance, but a possible simplification:

image

That gets rid of the record, and makes for one fewer dictionary mapping.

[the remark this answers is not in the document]
NateMerritt: I found it's a helpful practice to separate out the DB logic from the caching logic while still keeping them close together.  Local static functions work well for that because I can give the function a clear name which helps with readability, there's no closure over the enclosing method's variables (because its `static`), and the local function can only be called by the enclosing method (unlike a `private static` method).

I've also seen a repository layer implemented where the repository layer doesn't do caching and thus the caching sits on top of it.  If we ever go that direction then this code would be easier to split out.  Performance wise the DB method call shouldn't happen much (because we're caching the result) so I tend toward separating concerns and readability.

That said, in this case the DB call is very straightforward and this does add more lines of code so I see an argument to inline it.

src/Aquifer.Common/Services/CachingLanguageService.cs
NateMerritt: I'm fine implementing whatever practice we prefer, though I might accidentally lapse into old habits if we don't have a compiler or linter enforcing it.

Some Google searching tells me that there aren't any documented C# framework guidelines for best practices:
* https://www.reddit.com/r/csharp/comments/16zihxz/order_of_methods_in_class/
* https://stackoverflow.com/questions/150479/order-of-items-in-classes-fields-properties-constructors-methods/27997842

We could probably get some enforcement from `.editorconfig` and I know that StyleCop can allow enforcing a certain order because I've used that previously.

My reasoning for `private` things being last is summed up in [this answer](https://stackoverflow.com/a/27997842).  I was initially compelled into `private` fields being at the end of the class when I started working at my last job and didn't like it at first because I initially learned with C++ style declarations that _required_ fields to be declared before they were used (and thus they had to be at the top of a class).  I think the `private fields first` is a holdover from that old compiler requirement.  C# allows any ordering and now I see value in putting the constructors and other methods at the top (with `public` ones first) because that's what I tend to want to read first and the private fields are implementation details of those methods (so I want to read them last).  Code is read more than it is written so defaulting to an ordering that helps the reading process helps with understandability.

src/Aquifer.Common/Services/CachingLanguageService.cs
NateMerritt: Yep, I wrote this before we had that discussion and missed going back to fix it.  I'll look into enforcing it somehow in Visual Studio or ideally the build itself.

src/Aquifer.Common/Services/CachingLanguageService.cs
jwinston-bn: Yeah, I definitely like seeing the constructor first, and the primary constructors in general handle that now. My format on save throws them up top, and that's how I've always done it, so I don't think much of it. Throughout the project I think they're mostly up top now. I don't have a strong preference, just wanted to throw it out there since I know some of our IDEs are gonna move them automatically.

src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Endpoint.cs
jwinston-bn: I'm on the fence here. I wouldn't expect a third-party API to handle sorting for me. That's the type of thing that can change without warning, especially if it isn't explicitly documented that you can trust said sorting. If we're seeing a big performance improvement, I'm inclined to just make the change.

src/Aquifer.Common/Services/CachingLanguageService.cs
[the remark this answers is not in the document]
NateMerritt: > a possible simplification

> That gets rid of the record, and makes for one fewer dictionary mapping.

Yeah, the additional record was due to the notion that one day we might want to split out the DB logic into a repository layer.

Do you feel strongly that it should be simplified?

src/Aquifer.Common/Services/CachingLanguageService.cs
NateMerritt: The default value of the cached dictionary is `default<int>` which is `0`.  Thus I used the `TryGetValue` strategy so that I could return `null` instead for the default to represent the fact that the there is no available ID for the given language code (and vice versa for the other method above).

You would _think_ that we could still use `GetValueOrDefault()` with a `defaultValue` of `null` in that scenario like so:
```
return (await GetLanguageCodeByIdMapAsync(ct)).GetValueOrDefault(languageId, defaultValue: null);
```
but it unfortunately doesn't work due to the return type of that extension method not being nullable.

jwinston-bn: Ah, that explains why Rider only suggested the other one be updated, and not this one (I accidentally put the suggestion on this one, when I meant to put it on the one above). I was wondering what the difference was. So the other one can use `GetValueOrDefault` since string is nullable, but this one can't.

NateMerritt: The Dictionary would have to have `string?` but it uses `string` so the other usage also can't use `GetValueOrDefault()`.

src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Endpoint.cs
NateMerritt: @jwinston-bn, @MrGrinst, should I go forward with this sorting change?

src/Aquifer.Common/Services/CachingLanguageService.cs
NateMerritt: I moved the private fields to the top.

src/Aquifer.Public.API/Endpoints/Resources/Search/GetResources/Endpoint.cs
jwinston-bn: You mean add a sorting parameter that the user can send?

jwinston-bn (approved):

PR #472, opened 2024-10-04
add AcceptsCompany option to reports table

MrGrinst:


jwinston-bn (approved):

src/Aquifer.Data/Migrations/20241004205123_AddAcceptsCompanyToReportsTable.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<bool>(
+                name: "AcceptsCompany",
+                table: "Reports",
+                type: "bit",
+                nullable: false,
+                defaultValue: false);
```
NateMerritt, line 18: Out of curiosity, if we add new columns and need to backfill them with a manual process how do we implement that?  A SQL script that lives in source control in an archive folder?  Or some other mechanism?

MrGrinst: The few times we've needed to do that we'll put a script together, apply it in QA once the migration has run, and post it in the #studio-prod-readiness channel. Then when we do prod deploys we check that channel to see if there is anything that will need to be run after a given deploy.

PR #473, opened 2024-10-07
Added Created Date To Response And Create EndPoint Should Allow Extra Translation

carlosmorales125:


src/Aquifer.API/Endpoints/Resources/Content/Get/Response.cs
```diff
@@ -178,4 +178,6 @@ public class MachineTranslationResponse
     public bool ImproveClarity { get; set; }
     public bool ImproveTone { get; set; }
     public bool ImproveConsistency { get; set; }
+    public string? RetranslationReason { get; set; }
```
jwinston-bn, line 181: Do we need to send the reason, or do we just need to send that it was already done? I'm wondering if this should be a bool that's `HadRetranslation` and the check is that `RetranslationReason` isn't null.

src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Create/Endpoint.cs
```diff
@@ -21,9 +21,10 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
             .AsTracking()
             .FirstOrDefaultAsync(x => x.ResourceContentVersionId == req.ResourceContentVersionId && x.ContentIndex == req.ContentIndex, ct);
 
-        if (existingMt is not null)
+        if (existingMt is not null && !(existingMt.Created > DateTime.UtcNow.AddMinutes(-60)))
```
jwinston-bn, line 24: Nit: `.AddHours(-1)`

src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Create/Endpoint.cs
```diff
@@ -34,7 +35,8 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
             Content = req.Content,
             ContentIndex = req.ContentIndex,
             UserId = user.Id,
-            SourceId = req.SourceId
+            SourceId = req.SourceId,
+            RetranslationReason = req.RetranslationReason ?? null
```
jwinston-bn, line 39: The default value for `RetranslationReason` is null, so I don't think we need the `?? null` here.

src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Create/Endpoint.cs
jwinston-bn: I think we also need a check that `existingMt.RetranslationReason is not null`, because they can only do it once.

jwinston-bn: And really, if that isn't null, then `request.RetranslationReason` can't be empty or null.

jwinston-bn (approved):

PR #475, opened 2024-10-07
Adding Migration For ReTranslation button

carlosmorales125:


src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Create/Request.cs
MrGrinst, line 1: You'll need to add code into the `src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Create/Endpoint.cs` file as well to include the new property when saving.

carlosmorales125: I plan to finish that work here: https://github.com/BiblioNexusStudio/aquifer-server/pull/473.

MrGrinst (approved):

PR #478, opened 2024-10-08
Prevent duplicate key exception when writing IP address logging to DB.

NateMerritt:
Concern: I don't know how to test this.  How can I invoke the `TrackResourceContentRequestJob` locally?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -129,7 +129,7 @@ csharp_style_prefer_readonly_struct_member = true
 
 # Code-block preferences
 csharp_prefer_braces = true
-csharp_prefer_simple_using_statement = true
+csharp_prefer_simple_using_statement = false
 csharp_style_namespace_declarations = file_scoped:error
 csharp_style_prefer_method_group_conversion = true:suggestion
 csharp_style_prefer_top_level_statements = true:error
--- a/Aquifer.sln
+++ b/Aquifer.sln
@@ -19,7 +19,20 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aquifer.Common", "src\Aquif
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aquifer.Jobs", "src\Aquifer.Jobs\Aquifer.Jobs.csproj", "{CDDA72AF-C879-4CFC-9015-B0CB4979837F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Aquifer.Common.UnitTests", "Aquifer.Common.UnitTests\Aquifer.Common.UnitTests.csproj", "{93508A60-45FD-4A2E-9EFC-8D7C18259B20}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aquifer.Common.UnitTests", "Aquifer.Common.UnitTests\Aquifer.Common.UnitTests.csproj", "{93508A60-45FD-4A2E-9EFC-8D7C18259B20}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{2CA3FDD3-907D-4B4B-A657-A802AC9D0316}"
+	ProjectSection(SolutionItems) = preProject
+		.editorconfig = .editorconfig
+		.gitignore = .gitignore
+		LICENSE = LICENSE
+		README.md = README.md
+	EndProjectSection
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{6893088B-7C0B-414B-B98F-181474574125}"
+	ProjectSection(SolutionItems) = preProject
+		docs\infrastructure.md = docs\infrastructure.md
+	EndProjectSection
 EndProject
 Global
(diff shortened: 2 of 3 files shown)


Aquifer.sln
```diff
@@ -19,7 +19,20 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aquifer.Common", "src\Aquif
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aquifer.Jobs", "src\Aquifer.Jobs\Aquifer.Jobs.csproj", "{CDDA72AF-C879-4CFC-9015-B0CB4979837F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Aquifer.Common.UnitTests", "Aquifer.Common.UnitTests\Aquifer.Common.UnitTests.csproj", "{93508A60-45FD-4A2E-9EFC-8D7C18259B20}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Aquifer.Common.UnitTests", "Aquifer.Common.UnitTests\Aquifer.Common.UnitTests.csproj", "{93508A60-45FD-4A2E-9EFC-8D7C18259B20}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{2CA3FDD3-907D-4B4B-A657-A802AC9D0316}"
+	ProjectSection(SolutionItems) = preProject
```
NateMerritt, line 25: I added some files from the solution root directory to the Solution view for convenience.

MrGrinst (approved):

.editorconfig
```diff
@@ -129,7 +129,7 @@ csharp_style_prefer_readonly_struct_member = true
 
 # Code-block preferences
 csharp_prefer_braces = true
-csharp_prefer_simple_using_statement = true
+csharp_prefer_simple_using_statement = false
```
NateMerritt, line 132: This doesn't prevent simple using statements but allows using old style ones with braces without getting squiggles.  Clearly demonstrating when the using goes out of scope (and thus the dispose operation runs) is sometimes important and I think it should be left to developer discretion to choose which syntax creates the most readable code.  There may also be performance considerations (see examples below). Thoughts?

Examples:
Simple:
```
private Task<Something> MethodAsync()
{
    using var transaction = await dbContext.Database.BeginTransactionAsync(IsolationLevel.Serializable, ct);
    // fetch data from the DB
    // save data in the DB

    // do some things in memory

    // make a web request

    return something;

    // transaction goes out of scope here due to implicit using scope and all above operations took place in that transaction include extraneous web request which needlessly lock rows in the DB for longer than needed
}
```
Old style:
```
private Task<Something> MethodAsync()
{
    using (var transaction = await dbContext.Database.BeginTransactionAsync(IsolationLevel.Serializable, ct))
    {
        // fetch data from the DB
        // save data in the DB

        // transaction goes out of scope here due to explicit closing brace
    }

    // do some things in memory

    // make a web request

    return something;
}
```

src/Aquifer.Jobs/TrackResourceContentRequestJob.cs
```diff
@@ -62,32 +63,27 @@ private async Task LookupIpAddressAsync(TrackResourceContentRequestMessage track
                 return;
             }
 
-            var ipDataRecord = await dbContext.IpAddressData
-                .AsTracking()
-                .SingleOrDefaultAsync(x => x.IpAddress == trackingMetadata.IpAddress, ct);
-            if (ipDataRecord is null)
+            using (var transaction = await dbContext.Database.BeginTransactionAsync(IsolationLevel.Serializable, ct))
```
NateMerritt, line 66: By default selects are not included in the EF generated transaction created via `SaveChangesAsync()`.  By making the transaction here explicit the read is now included in the same transaction as the write.  The `Serializable` isolation level will cause SQL Server to lock relevant rows and to prevent other transaction from concurrently accessing those rows.  Thus, the duplicate key exception should be prevented.

.editorconfig
jwinston-bn: Yeah, this makes sense.

src/Aquifer.Jobs/TrackResourceContentRequestJob.cs
jwinston-bn: The problem, as I understand it, is two situations where `ipDataRecord` is null, so it tries to do two separate inserts with the same PK (the IP Address). How does the select play into that?

NateMerritt: The DB will lock any rows (and prevent any potential new rows created by other transactions) for our read query by using a range lock, even if there are no results.  By using a `Serializable` transaction the DB guarantees that there will still be no results for that same read query when we commit the writes later.  If we use a different transaction isolation level like `ReadUncommitted` then you wouldn't have the same guarantee about reads (and it wouldn't do range locking).

Note that range locking is more expensive and could result in a full table lock for the duration of the transaction.  If this change causes transaction failures due to that locking then we could switch it back to (a) catch an exception and (b) swallow the error with no logging, but only if (c) the exception is specifically a SQL duplicate key error (by looking at the error number).

jwinston-bn (approved):

NateMerritt:
Closing this PR because Jesse's changes to logging have prevented the `DuplicateKeyException` error from appearing in `studio-app-insights-alerts` channel.

PR #480, opened 2024-10-09
Add parent-resources route to public API.

NateMerritt:


src/Aquifer.Common/Services/Caching/CachingLanguageService.cs
```diff
@@ -61,7 +61,7 @@ public async Task<IReadOnlyDictionary<string, int>> GetLanguageIdByCodeMapAsync(
 
                 return
                     (LanguageCodeById,
-                    LanguageIdByCode: LanguageCodeById.ToDictionary(li => li.Value, li => li.Key));
+                    LanguageIdByCode: LanguageCodeById.ToDictionary(li => li.Value, li => li.Key, StringComparer.OrdinalIgnoreCase));
```
NateMerritt, line 64: This will allow passing language codes like "hiN" and still getting back results.

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
```diff
@@ -0,0 +1,14 @@
+namespace Aquifer.Public.API.Endpoints.ParentResources.List;
+
+public sealed class Request
```
NateMerritt, line 3: Do we also want to allow filtering by `ResourceType`?

src/Aquifer.Public.API/Endpoints/ParentResources/List/Endpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+            s.Description =
+                "Returns all parent resources for a given language including the count of resources associated with each parent.";
+        });
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
+    {
+        var languageId = await GetLanguageIdAsync(req.LanguageId, req.LanguageCode, ct);
+
+        var query = $"""
```
NateMerritt, line 29: This is mostly a copy/paste of the Internal API's code but I removed the default fallback to English for images and videos as well as a few columns we didn't need.

src/Aquifer.Public.API/Endpoints/ParentResources/List/Endpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+            SELECT
+                COALESCE(prl.DisplayName, pr.DisplayName) AS DisplayName,
+                (
+                    SELECT
+                        COUNT(rcv.Id)
```
NateMerritt, line 34: Fetching the count for English parent resources in dev on my local takes just shy of 300 ms.  This seems acceptable but isn't fast.

src/Aquifer.Public.API/Endpoints/ParentResources/List/Endpoint.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.Public.API.Endpoints.ParentResources.List;
+
+public sealed class Endpoint(AquiferDbContext _dbContext, ICachingLanguageService _cachingLanguageService)
+    : Endpoint<Request, List<Response>>
+{
+    public override void Configure()
+    {
+        Get("/resources/parent-resources");
```
jwinston-bn, line 15: The word "resource" is very overloaded, and it was a whole ordeal that we called an individual article a resource. Per the documentation, what we would call a parent resource, is a resource collection. So this needs to be like the `Types` endpoint and be `/resources/collections` (I should have brought that up when you got the ticket, but didn't really think about it till seeing this). Secondly, to be consistent, the path should be reflected by the folder structure. Since this starts with `/resources`, it should be in the `Endpoints\Resources` folder.

jwinston-bn: For reference: 
image

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
jwinston-bn: If you mean by "Dictionary" or "Guide", I'd say might as well.

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    /// <summary>
+    /// Optional language to search on by internal Aquifer language id. If no language is specified then English will be used by default. Only one of language id or code may be provided.
+    /// </summary>
+    public int? LanguageId { get; init; }
```
jwinston-bn, line 8: I could see users wanting to know all languages a collection is available for. I would say that if no language id/code is provided, rather than default to English, it returns all. And to that end, I'd let them send a list of language ids. My thinking is that the response would have the collection name and information, and then a list of `AvailableLanguages` with an object that's like `[{"languageId": 2, "languageCode": "TPI", "resourceItemCount": 153}, etc.]`

src/Aquifer.Public.API/Endpoints/ParentResources/List/Response.cs
```diff
@@ -0,0 +1,23 @@
...
+    public required string DisplayName { get; init; }
+    public required string Code { get; init; }
+    public required int ResourceCountForLanguage { get; init; }
+
+    [JsonIgnore]
+    public string? LicenseInfoValue { get; init; }
+
+    public object? LicenseInfo => LicenseInfoValue != null
+        ? JsonUtilities.DefaultDeserialize(LicenseInfoValue)
+        : null;
```
jwinston-bn, line 20: I bet we're doing this exact thing all over the place. Makes me wonder if this should be its own object that gets inherited when needed. Though there are probably instances where we do something similar with two different properties, and I guess you can't inherit multiple classes in C#, but might still be worth looking into (outside the scope of this ticket).

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
NateMerritt: > I could see users wanting to know all languages a collection is available for.

That's a different use case than the [route that I copied](https://github.com/BiblioNexusStudio/aquifer-server/blob/master/src/Aquifer.API/Endpoints/Resources/ParentResources/List/Endpoint.cs) from the internal API.  That route is language specific and only returns the localized name, not the original collection's name, unless that name happens to be the `code` or in the license information.  That existing route (and thus the route in this PR as I originally implemented it) was designed to show information for only a single language.

>  I would say that if no language id/code is provided, rather than default to English, it returns all.

It's not currently defaulting to English but will throw an error if no language information is provided in the request.

> I'd let them send a list of language ids.

As it stands now they could just hit the route multiple times, once for each language they care about.  I have concerns about returning more than one language in a single response (see below).

> My thinking is that the response would have the collection name and information, and then a list of `AvailableLanguages` with an object that's like `[{"languageId": 2, "languageCode": "TPI", "resourceItemCount": 153}, etc.]`

Fetching the counts for a single language is already pushing what I'd consider the maximum DB performance threshold (~300 ms) and it's only going to get more expensive to count these over time as we fill the Aquifer.  Thus, we'd probably want to leave off the counts if we accept more than one `languageId` or find another solution for the counts like keeping a table up-to-date with total counts or using a view that keeps the counts up-to-date automagically.

So all that said, I'm wondering if the best thing here might be two different routes, one that is "by language" (i.e. this PR's route) and another new route that returns more of a summary view which shows all collections in our system and the languages for which they have any translated content (but without counts).  Thoughts?

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
NateMerritt: Ok, I'll add that in.

NateMerritt, line 3: I just realized that this route should probably also support paging.  We don't have a lot of collections now (only 17 in production) but there could be a larger number in the future and it's impossible to add paging later without breaking consumers.  @jwinston-bn, what do you think?  Are we likely to have more than 100 some day?

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
NateMerritt: I'm actually not likely my idea in the last comment because it looks like the counts _are_ necessary in order to make sense of the data.  Check out the prod data for Malaysian where there are only two collections with any content but we've localized all of the collection names in advance.  Looking at current data, a summary route that only relies on the presence of a localized name will return a summary for every collection for every language.
image.

NateMerritt: Update after video chat: There will be two routes, List and Get (by code).  List will have minimal data from the english parent resource.  Get will join on localization data and include counts for all languages but for only a single parent resource.

src/Aquifer.Common/Services/Caching/CachingLanguageService.cs
```diff
@@ -57,11 +70,113 @@ public async Task<IReadOnlyDictionary<string, int>> GetLanguageIdByCodeMapAsync(
...
+        }
+
+        throw new ArgumentException($"One of {nameof(languageId)} or {nameof(languageCode)} must be non-null.");
+    }
+
+    public async Task<
+            (IReadOnlyList<int> InvalidLanguageIds,
+            IReadOnlyList<string> InvalidLanguageCodes,
+            IReadOnlyList<int>? ValidLanguageIds)>
+        ValidateLanguageIdsOrCodesAsync(
```
NateMerritt, line 122: I moved a couple of validation methods here so they can easily be used by all of the endpoints that deal with language IDs or Codes.

tests/Aquifer.Public.API.UnitTests/ResourcesCollectionsGetEndpointValidatorTests.cs
```diff
@@ -0,0 +1,58 @@
+using Aquifer.Public.API.Endpoints.Resources.Collections.Get;
+using FluentValidation.TestHelper;
+
+namespace Aquifer.Public.API.UnitTests;
+
+public sealed class ResourcesCollectionsGetEndpointValidatorTests
```
NateMerritt, line 6: I tried out the `FluentValidation.TestHelper` with XUnit and am a fan.  It was pretty quick to throw these tests together and they are all tests I would have had to perform manually every time that I changed the validator logic.

I went away from "best practices" and just have two methods, one for bad requests and one for good requests.  This can lend itself to tests that have multiple properties causing a failure but I think it's (a) easier to read because there is much less test code and (b) the pitfalls can be mitigated by only changing one property to invalid at a time.

tests/Aquifer.Common.UnitTests/Aquifer.Common.UnitTests.csproj
```diff
@@ -16,7 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\src\Aquifer.Common\Aquifer.Common.csproj" />
+    <ProjectReference Include="..\..\src\Aquifer.Common\Aquifer.Common.csproj" />
```
NateMerritt, line 19: I accidentally had the Aquifer.Common.UnitTests project in the root folder and moved it to `tests`.

src/Aquifer.Public.API/Endpoints/Resources/Collections/List/Request.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// The optional type of resource to search for, such as "Dictionary". If none specified, will default to None.
+    /// </summary>
+    [DefaultValue(ResourceType.None)]
+    public ResourceType ResourceType { get; init; }
+
+    /// <summary>
+    /// The number of results to return.
+    /// </summary>
+    [DefaultValue(MaximumLimit)]
+    public int Limit { get; init; } = MaximumLimit;
```
NateMerritt, line 20: I decided to go ahead and add offset and limit.  Their default values will allow fetching all values but it's future proofing in case we end up with many more parent resources one day.

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Response.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public sealed class Response
+{
+    public required string Code { get; init; }
+    public required string DisplayName { get; init; }
+    public required string ShortName { get; init; }
+    public required ResourceType ResourceType { get; init; }
+
+    [JsonConverter(typeof(JsonUtilities.RawJsonConverter))]
+    public required string? LicenseInfo { get; init; }
```
NateMerritt, line 15: I only exposed `LicenseInfo` from here in the GET route, not in the List route.  Is that acceptable?

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs
```diff
@@ -0,0 +1,211 @@
...
+            WHERE
+                r.ParentResourceId = @parentResourceId AND
+                {resourceContentCountWhereClause}
+                rcv.IsPublished = 1
+            GROUP BY
+                rc.LanguageId;
+            """;
+
+        // querying separately and joining in memory saves significant time over trying to join in the DB
+        var reader = await dbConnection.QueryMultipleAsync(new CommandDefinition(
```
NateMerritt, line 179: Dapper allows for multiple `SELECT`s in a single call to the DB which can sometimes help with performance to avoid doing a DB JOIN operation.  My method here is kind of like a manual `SplitQuery()` except EF makes two separate DB calls and this is only one.

src/Aquifer.Public.API/Endpoints/ParentResources/List/Request.cs
NateMerritt: Update: I added paging.

src/Aquifer.Public.API/Endpoints/ParentResources/List/Response.cs
NateMerritt: @jwinston-bn, I'm pretty happy with this solution which removed this complexity all together and should also improve performance because instead of deserializing and serializing it keeps the JSON string as is (but still validates it before returning): https://github.com/BiblioNexusStudio/aquifer-server/pull/480/files#diff-b2898cfdfb4c4a326edb830c8f3523312ccfb2a8a18da0432384f980b3e7a6abR37

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs
```diff
@@ -0,0 +1,211 @@
...
+{
+    public override void Configure()
+    {
+        Get("/resources/collections/{code}");
+        Options(EndpointHelpers.UnauthenticatedServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
+        Summary(s =>
+        {
+            s.Summary = "Gets a resource collection with language localization data for the given collection code.";
+            s.Description =
+                "Returns the resource collection matching the collection code including the count of resource items in each available language.";
```
NateMerritt, line 23: Is "resource items" the correct language for public API documentation of what we call "resources"?

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs
```diff
@@ -0,0 +1,211 @@
...
+        string LicenseInfo);
+
+    private static async Task<IReadOnlyList<ParentResourceLocalizationWithResourceContentCount>>
+        GetParentResourceLocalizationWithResourceContentCountAsync(
+            DbConnection dbConnection,
+            int parentResourceId,
+            IReadOnlyList<int>? languageIds,
+            CancellationToken ct)
+    {
+        var parameters = new DynamicParameters(new { parentResourceId });
```
NateMerritt, line 121: `DynamicParameters` are a nice Dapper feature that allows starting with static parameters and then dynamically adding in more in conditional statements without having to deal with ADO.NET pain.

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs
```diff
@@ -0,0 +1,211 @@
...
+        Options(EndpointHelpers.UnauthenticatedServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
+        Summary(s =>
+        {
+            s.Summary = "Gets a resource collection with language localization data for the given collection code.";
+            s.Description =
+                "Returns the resource collection matching the collection code including the count of resource items in each available language.";
+        });
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
```
NateMerritt, line 27: I'm happy with the current performance and long-term sustainability of the performance of this route.  Also, if anything gets expensive we could add caching of the counts.

As it stands right now fetching the counts for all languages for a single parent resource doesn't take longer than ~200ms and the other two queries are fast.

Getting it fast required moving away from the old query used by the Internal API.  See below for more comments about this.

src/Aquifer.Common/Services/Caching/CachingLanguageService.cs
```diff
@@ -57,11 +70,113 @@ public async Task<IReadOnlyDictionary<string, int>> GetLanguageIdByCodeMapAsync(
                 cacheEntry.SlidingExpiration = s_cacheLifetime;
 
                 var LanguageCodeById = await _dbContext.Languages
-                    .ToDictionaryAsync(li => li.Id, li => li.ISO6393Code, ct);
+                    .Select(l => new { l.Id, l.ISO6393Code })
```
NateMerritt, line 73: Interesting performance discovery: EF isn't smart enough to use the `ToDictionary()` properties to inform the columns in the SELECT.  You can force it to only retrieve the ones you care about with a manual `.Select()`.

src/Aquifer.Public.API/Endpoints/Resources/Collections/List/Request.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Aquifer.Public.API.Endpoints.Resources.Collections.List;
+
+public sealed class Request
+{
+    public const int MaximumLimit = 100;
+
+    /// <summary>
+    /// The optional type of resource to search for, such as "Dictionary". If none specified, will default to None.
+    /// </summary>
+    [DefaultValue(ResourceType.None)]
```
NateMerritt, line 13: I copied this from another route but interestingly this shows the default value as `0` in the documentation.  This can be corrected by using the following instead:
```
[DefaultValue(nameof(ResourceType.None))]
```
Has anyone else encountered this?  I'm wondering if there's a tooling solution to automatically `ToString()` any enum in default value documentation.

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs
```diff
@@ -0,0 +1,211 @@
...
+        if (invalidLanguageIds is not [])
+        {
+            ThrowError(r => r.LanguageIds, $"Invalid 'Language Ids': \"{string.Join("\", \"", invalidLanguageIds)}\"");
+        }
+        if (invalidLanguageCodes is not [])
+        {
+            ThrowError(r => r.LanguageCodes, $"Invalid 'Language Codes': \"{string.Join("\", \"", invalidLanguageCodes)}\"");
+        }
+
+        var parentResource = await GetParentResourceAsync(_dbContext.Database.GetDbConnection(), req.Code, ct);
```
jwinston-bn, line 44: What's the reason to pass `_dbContext.Database.GetDbConnection()` as a parameter here, since the private method has access to `_dbContext` anyway?

src/Aquifer.Common/Utilities/JsonUtilities.cs
```diff
@@ -26,4 +26,27 @@ public static string DefaultSerialize(object value)
...
     }
+
+    /// <summary>
+    /// Serializes the contents of a string value as raw JSON.  The string is validated as being an RFC 8259-compliant JSON payload.
+    /// </summary>
+    /// <example>
+    /// [JsonConverter(typeof(JsonUtilities.RawJsonConverter))]
+    /// public required string Data { get; init; } // JSON string
+    /// </example>
+    public sealed class RawJsonConverter : JsonConverter<string>
```
jwinston-bn, line 37: Very nice.

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Response.cs
jwinston-bn: I think it should be okay.

src/Aquifer.Public.API/Endpoints/Resources/Collections/List/Request.cs
jwinston-bn: Maybe could do something in `DefaultParameterOperationProcessor.cs`

src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs
NateMerritt: I intended to only call `_dbContext.Database.GetDbConnection()` only once and then pass the retrieved connection to each of the methods that needs it.  I'll fix that.

If/when we refactor to inject a `SqlConnectionFactory` (which we'd need to create) then the EF `DbContext` won't be necessary for any classes that strictly use Dapper, only a `SqlConnection`.  The `DbContext` has DI magic going on right now that automatically opens and closes `SqlConnection`s under the hood for the lifetime of the DI class (the `Endpoint` in our case) but you can turn off that magic in configuration.  Dapper doesn't have that, so we'd need to manually open/close the connection.  Why move in that direction?  Doing this refactor would be necessary to put in a profiling connection in order to log the SQL from Dapper commands (we can create such a connection from the factory).  The C# would look something like this:
```
using (var dbConnection = await _sqlConnectionFactory.OpenConnectionAsync(ct))
{
    var results = await dbConnection.QueryAsyncWithRetries(...);

    // the Dispose at the end of the using automatically closes the connection
}
```

Interestingly, the `DbContext` can be attached to an existing open `DbConnection` if we want to intermix EF and Dapper when using manual connections, but I'm not sure if it would double log EF queries at that point if we tried to enable SQL profiling logging.

jwinston-bn (approved):

PR #481, opened 2024-10-09
Add retry logic to open ai calls

jwinston-bn:
I couldn't reproduce the 500 not showing up to the front-end. It may be caused by APIM. I hope this just resolves it anyway, because in my own local apps that do translation it seems like OpenAI is prone to random failures, and I've implemented similar retry logic to get around it. Usually the failures are just a one-off and next try is good.

### DIFF
--- a/src/Aquifer.API/Endpoints/AI/Translate/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/AI/Translate/Endpoint.cs
@@ -48,12 +48,33 @@ public partial class Endpoint(
                     sentence += ".&nbsp;";
                 }
 
-                using var clientResponse = await openAiClient.PostChatCompletionsAsync(prompt, sentence, ct);
-                clientResponse.EnsureSuccessStatusCode();
+                await StreamCompletionToResponseBody(prompt, sentence, ct);
+            }
+        }
+    }
 
-                await using var stream = await clientResponse.Content.ReadAsStreamAsync(ct);
-                await stream.CopyToAsync(HttpContext.Response.Body, ct);
-                await HttpContext.Response.Body.FlushAsync(ct);
+    private async Task StreamCompletionToResponseBody(string prompt, string sentence, CancellationToken ct, int retries = 0)
+    {
+        try
+        {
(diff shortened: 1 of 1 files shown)


src/Aquifer.API/Endpoints/AI/Translate/Endpoint.cs
```diff
@@ -48,12 +48,33 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
-                await using var stream = await clientResponse.Content.ReadAsStreamAsync(ct);
-                await stream.CopyToAsync(HttpContext.Response.Body, ct);
-                await HttpContext.Response.Body.FlushAsync(ct);
+            await using var stream = await clientResponse.Content.ReadAsStreamAsync(ct);
+            await stream.CopyToAsync(HttpContext.Response.Body, ct);
+            await HttpContext.Response.Body.FlushAsync(ct);
+        }
+        catch (Exception)
+        {
+            if (++retries < 3)
```
NateMerritt, line 69: It would be good to log the response status code and body if this is not a success status code.  Is there a way to log an info message?

jwinston-bn: Honestly, this happens all the time and it's OpenAI just flat out failing, and then you hit it again a moment later and it's fine. So I just don't think we care unless it gets out of the retry logic and actually fails our translation. It's not a log anyone's gonna go looking for unless it fails the translation, in which case we would get the exception logged.

But I've run many translations locally on huge files watching it go one by one, where it might make 500 calls, and I can just watch them randomly fail here and there, do a quick retry, and all's good.

NateMerritt: Yeah, I'm more concerned about 400s getting swallowed than about logging every 500 from Open API.  I'd still argue for logging 500s at a level where they don't alert us in Slack but are still visible in Azure if you look for them, though.  It's helpful to know what's happening and each failure costs at 500 ms delay in the response time for the user.

NateMerritt (approved):

PR #482, opened 2024-10-09
Add FeedbackType to the FeedbackEntity and BibleWell Feedback endpoint

elisabeth-shakespeare:


src/Aquifer.Data/Entities/FeedbackEntity.cs
```diff
@@ -8,7 +8,14 @@ public class FeedbackEntity
...
     public string Feedback { get; set; } = null!;
+    public FeedbackType FeedbackType { get; set; }
 
     [SqlDefaultValue("getutcdate()")]
     public DateTime Created { get; set; } = DateTime.UtcNow;
+}
+
+public enum FeedbackType
+{
+    WellContactForm = 0,
```
jwinston-bn, line 19: We should have None be 0. We don't want to default to some random value.

src/Aquifer.Data/Entities/FeedbackEntity.cs
```diff
@@ -8,7 +8,15 @@ public class FeedbackEntity
...
 
     [SqlDefaultValue("getutcdate()")]
     public DateTime Created { get; set; } = DateTime.UtcNow;
+}
+
+public enum FeedbackType
+{
+    None = 0,
+    WellContactForm = 1,
+    WellContactPrompt = 2
```
MrGrinst, line 21: I'm not sure I understand the different between these. Should it be something like `WellContactOnly` and `WellFullFeedback` or something to distinguish really clearly?

src/Aquifer.API/Endpoints/Feedback/Resources/Content/Create/Request.cs
```diff
@@ -9,5 +9,6 @@ public record Request
     public string? ContactValue { get; set; }
     public FeedbackContactType? ContactType { get; set; }
     public string? Feedback { get; set; }
+    public FeedbackType FeedbackType { get; set; }
```
MrGrinst, line 12: I think we should default this to whatever one the existing feedback form will go to that way old clients sending data aren't setting it to `None`

jwinston-bn: It gets so little traffic I think it's okay not to put in a default that isn't really desired (we only got our first feedback today actually). But that said, the validator's gonna fail for an old client with the `NotEmpty()`, so I think that needs to be removed for now.

src/Aquifer.Data/Entities/FeedbackEntity.cs
jwinston-bn: My thinking was that one is the form they have to manually go to, while the other one is the prompt that comes up. ContactOnly and FullFeedback seems vague to me also, since we could decide to add more to the prompt if we wanted. I'm not really sure what good names for these are, they're so similar, and could potentially get more overlap as time goes by.

MrGrinst: If we're trying to differentiate auto vs user-triggered rather than what the feedback is inside them, it almost seems like we should call it `FeedbackTriggerType` and have `Auto` and `User`.

src/Aquifer.API/Endpoints/Feedback/Resources/Content/Create/Request.cs
elisabeth-shakespeare: I'm putting in FeedbackType for the existing feedback call in well. Maybe I can put the validator back in once that's deployed

src/Aquifer.Data/Entities/FeedbackEntity.cs
elisabeth-shakespeare: I don't know about the word "Trigger", because a voluntarily filled form is not necessarily triggered. Also, a User fills out both something that's Auto and Volunteered.

jwinston-bn: We're just trying to distinguish which feedback form it came from in the app. One the user has to manually navigate to, and now the new one that Elisabeth is working on is a popup. But aside from their location in the app, we shouldn't assume any differences between them, and we should assume we'll have other feedback forms in the future.

How do you both feel about `WellSettingsFeedbackForm` and `WellPopupFeedbackForm`? This shouldn't hold us up much more though, no one is seeing these enum names except us and there's nothing functional about the names, they can be changed later.

elisabeth-shakespeare: The reqs also have contact type dropdown with a value. In the `ResourceContentVersionFeedback` table we have both of them as fields. 
- Are we okay with `ContactType` and `ContactValue` going in the JSON `Feedback` column? 
- Or would we want them to be new columns on the Feedback Table like the `ResourceContentVersionFeedback` table?
- Would it be better to have some sort of intermediary table handle that info?

src/Aquifer.API/Endpoints/Feedback/BibleWell/Create/Endpoint.cs
```diff
@@ -14,13 +14,18 @@ public override void Configure()
 
     public override async Task HandleAsync(Request req, CancellationToken ct)
     {
+        var emailValue = (req.ContactType == FeedbackContactType.Email) ? req.ContactValue : req.Email;
+        var phoneValue = (req.ContactType == FeedbackContactType.Phone) ? req.ContactValue : req.Phone;
```
elisabeth-shakespeare, line 18: What about this?

jwinston-bn: I think this makes sense, but I would check that Email and Phone are null also. We don't want to overwrite a value that was explicitly entered.

MrGrinst (approved):

PR #485, opened 2024-10-11
Add Dapper retry overloads.

NateMerritt:
Dapper doesn't build in automatic retries.  This PR adds new "WithRetries" extension methods that we should use instead of the primary Dapper extension methods.  The new methods use the EF Core transient SQL Exception detection and Polly to trigger retries.

### DIFF
--- a/src/Aquifer.API/Program.cs
+++ b/src/Aquifer.API/Program.cs
@@ -55,6 +55,8 @@ builder.Services.AddOptions<ConfigurationOptions>().Bind(builder.Configuration);
 
 var app = builder.Build();
 
+StaticLoggerFactory.LoggerFactory = app.Services.GetRequiredService<ILoggerFactory>();
+
 app.UseCors(b => b.AllowAnyOrigin().AllowAnyHeader().AllowAnyMethod());
 app.UseHealthChecks("/_health");
 if (builder.Environment.IsDevelopment())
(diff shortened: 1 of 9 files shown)


src/Aquifer.API/Program.cs
```diff
@@ -55,6 +56,8 @@
 
 var app = builder.Build();
 
+StaticLoggerFactory.LoggerFactory = app.Services.GetRequiredService<ILoggerFactory>();
```
NateMerritt, line 59: Inspired by https://stackoverflow.com/a/50646048 so that I could have logging from extension methods without passing an `ILogger` parameter.

src/Aquifer.Logging/Aquifer.Logging.csproj
jwinston-bn, line 1: I'm hesitant to create a whole new project for one file that appears to be used in only one place. Can this just live in Aquifer.Data and be specific to that? Also, this still goes to app insights?

NateMerritt: > Also, this still goes to app insights?

It should use what AppInsights [wires up here](https://github.com/microsoft/ApplicationInsights-dotnet/blob/e34286b4c8175d7ea47f9227c5b981d661cac470/NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs#L399-L430) (I decompiled `AddApplicationInsightsTelemetry()` to check before creating this PR).  That's why I grabbed the existing `ILoggerFactory` from the app's DI services in the `Program.cs` instead of making a new `LoggerFactory` (which would require reconfiguring App Insights logging on the new factory, too).

jwinston-bn (approved):

PR #488, opened 2024-10-14
Bib 750 add admin help page to aquifer admin

kaseywright:
To keep things concise, yet extensible, I went with a single db table that will store the pertinent information for help documentation. For now, the `Id`, `Type`, `Title`, `Url`, `Created`, and `Updated` columns should suffice.

In the Endpoint, I opted for 2 queries over 1 more complex query because:
- We might want a different number of rows returned for different document types
- Queries should never have performance issues as table should not grow to be too large. If, for some reason this changes, add index(es) and adjust the query(ies) later.
- We can safely have a generous cache setting on this endpoint as the results will change predictably.

In the `HelpDocumentType ` I kept the `None = 0` type for consistency with the codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,22 +48,30 @@ dotnet build --no-incremental /p:WarningsAsErrors=true
 
 ## Database Migrations
 
-First initialize the settings file by copying `appsettings.example.json` as `appsettings.Development.json` and providing
+***Important***: Initialize the settings file in the `src/Aquifer.Migrations` directory by copying `appsettings.example.json` as `appsettings.Development.json` and providing
 the values needed for your instance.
 
 Entity Framework will generate migrations by comparing the C# Entities defined
 in the project and the current state of the database.
 
-To create a new migration:
+### Create a New Migration
+First, create your entity in the `Aquifer.Data/Entities` directory.
+
+Next, add your entitiy definition to the `src/Aquifer.Data/AquiferDbContext.cs` file.
+Entities are listed in an alphabetical order.
+
+Next, to create a new migration, run:
 ```bash
-dotnet ef migrations add --startup-project src/Aquifer.Migrations --project src/Aquifer.Data MigrationNameHere
+dotnet ef migrations add --startup-project src/Aquifer.Migrations --project src/Aquifer.Data <MigrationNameHere>
 ```
+Your new migration will be created in the `src/Aquifer.Data/Migrations` directory along with the `.Designer` file and updated `AquiferDbContextModelSnapshot.cs` file.
+
 
 If you run that command and the new migration file is empty, that means there
 were no changes detected between the C# Entities and the database. You can use
 this to your advantage to create empty migrations and add your own custom code.
 
-To run migrations:
+To run migrations and add your changes to the DB, run:
 ```bash
 dotnet ef database update --startup-project src/Aquifer.Migrations --project src/Aquifer.Data
 ```
--- a/README.md
+++ b/README.md
@@ -54,13 +54,14 @@ the values needed for your instance.
 Entity Framework will generate migrations by comparing the C# Entities defined
 in the project and the current state of the database.
 
-### Create a New Migration
+### Add New Entity/Migration
 First, create your entity in the `Aquifer.Data/Entities` directory.
 
-Next, add your entitiy definition to the `src/Aquifer.Data/AquiferDbContext.cs` file.
+Next, add your entity definition to the `src/Aquifer.Data/AquiferDbContext.cs` file.
 Entities are listed in an alphabetical order.
 
-Next, to create a new migration, run:
+### Create a New Migration
+To create a new migration, run:
 ```bash
 dotnet ef migrations add --startup-project src/Aquifer.Migrations --project src/Aquifer.Data <MigrationNameHere>
 ```
--- a/src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
@@ -1,3 +1,4 @@
+using Aquifer.API.Helpers;
 using Aquifer.Data;
 using Aquifer.Data.Entities;
 using FastEndpoints;
@@ -10,6 +11,7 @@ public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Respo
     public override void Configure()
     {
         Get("/help/aquifer-cms/documents");
+        ResponseCache(EndpointHelpers.OneDayInSeconds);
     }
 
     public override async Task HandleAsync(CancellationToken ct)
--- a/src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
@@ -3,13 +3,13 @@ using Aquifer.Data.Entities;
 using FastEndpoints;
 using Microsoft.EntityFrameworkCore;
 
-namespace Aquifer.API.Endpoints.Help;
+namespace Aquifer.API.Endpoints.Help.Aquifer_cms.Documents;
 
 public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Response>
 {
     public override void Configure()
     {
-        Get("/help");
+        Get("/help/aquifer-cms/documents");
     }
 
     public override async Task HandleAsync(CancellationToken ct)
--- a/src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
@@ -11,7 +11,7 @@ public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Respo
     public override void Configure()
     {
         Get("/help/aquifer-cms/documents");
-        ResponseCache(EndpointHelpers.OneDayInSeconds);
+        ResponseCache(EndpointHelpers.OneHourInSeconds);
     }
 
     public override async Task HandleAsync(CancellationToken ct)
@@ -19,8 +19,8 @@ public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Respo
         var helpDocuments = await GetHelpDocumentsAsync(ct);
         var response = new Response
         {
-            Releases = helpDocuments.FindAll(x => x.Type == HelpDocumentType.Release),
-            HowTos = helpDocuments.FindAll(x => x.Type == HelpDocumentType.HowTo)
+            Releases = helpDocuments.Where(x => x.Type == HelpDocumentType.Release),
+            HowTos = helpDocuments.Where(x => x.Type == HelpDocumentType.HowTo)
         };
         await SendOkAsync(response, ct);
     }
@@ -29,10 +29,10 @@ public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Respo
     {
         return await dbContext
             .HelpDocuments
-            .OrderByDescending(x => x.Updated)
+            .Where(x => x.Enabled)
+            .OrderByDescending(x => x.Created)
             .Select(x => new HelpDocumentResponse
             {
-                Id = x.Id,
                 Title = x.Title,
                 Type = x.Type,
                 Url = x.Url,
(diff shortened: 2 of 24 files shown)


README.md
```diff
@@ -48,22 +48,30 @@ dotnet build --no-incremental /p:WarningsAsErrors=true
...
 the values needed for your instance.
 
 Entity Framework will generate migrations by comparing the C# Entities defined
 in the project and the current state of the database.
 
-To create a new migration:
+### Create a New Migration
+First, create your entity in the `Aquifer.Data/Entities` directory.
+
+Next, add your entitiy definition to the `src/Aquifer.Data/AquiferDbContext.cs` file.
```
MrGrinst, line 60: `entitiy` spelling

README.md
```diff
@@ -48,22 +48,30 @@ dotnet build --no-incremental /p:WarningsAsErrors=true
...
 
-First initialize the settings file by copying `appsettings.example.json` as `appsettings.Development.json` and providing
+***Important***: Initialize the settings file in the `src/Aquifer.Migrations` directory by copying `appsettings.example.json` as `appsettings.Development.json` and providing
 the values needed for your instance.
 
 Entity Framework will generate migrations by comparing the C# Entities defined
 in the project and the current state of the database.
 
-To create a new migration:
+### Create a New Migration
```
MrGrinst, line 57: We may want to call this `### Adding a New Entity/Table` and differentiate from the process of the migration itself. Since I could just update a column on an existing table without needing to mess with the AquiferDbContext. Maybe below could be `### Create a New Migration`?

src/Aquifer.API/Endpoints/Help/Endpoint.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        var response = new Response
+        {
+            Releases = await GetRelesesAsync(ct),
+            HowTos = await GetHowTosAsync(ct)
+        };
+        await SendOkAsync(response, ct);
+    }
+
+    private async Task<List<HelpDocumentEntity>> GetRelesesAsync(CancellationToken ct)
```
MrGrinst, line 25: `Releses` spelling

src/Aquifer.API/Endpoints/Help/Endpoint.cs
```diff
@@ -0,0 +1,44 @@
...
+            .ToListAsync(ct);
+    }
+
+    private async Task<List<HelpDocumentEntity>> GetHowTosAsync(CancellationToken ct)
+    {
+        return await dbContext
+            .HelpDocuments
+            .Where(x => x.Type == HelpDocumentType.HowTo)
+            .OrderByDescending(x => x.Updated)
+            .Take(10)
```
MrGrinst, line 41: Was this in the requirements? For Releases a limit makes sense but wouldn't we want to show all How-Tos?

kaseywright: I'll remove the limit on the how top

src/Aquifer.API/Endpoints/Help/Endpoint.cs
```diff
@@ -0,0 +1,43 @@
...
+using FastEndpoints;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Help;
+
+public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Response>
+{
+    public override void Configure()
+    {
+        Get("/help");
```
jwinston-bn, line 12: I think this might be a little too generic, and there could be other forms of "help" in the future. We might have such a thing for Well in the future, for example. So I would probably go a bit further and say `/help/aquifer-cms/documents`, and have the file folder structure / namespace match that as well.

src/Aquifer.API/Endpoints/Help/Endpoint.cs
```diff
@@ -0,0 +1,43 @@
...
+        await SendOkAsync(response, ct);
+    }
+
+    private async Task<List<HelpDocumentEntity>> GetReleasesAsync(CancellationToken ct)
+    {
+        return await dbContext
+            .HelpDocuments
+            .Where(x => x.Type == HelpDocumentType.Release)
+            .OrderByDescending(x => x.Updated)
+            .Take(10)
```
jwinston-bn, line 31: Where did this requirement to only grab the last 10 come from? I'm not seeing that in the ticket.

src/Aquifer.API/Endpoints/Help/Response.cs
```diff
@@ -0,0 +1,9 @@
+using Aquifer.Data.Entities;
+
+namespace Aquifer.API.Endpoints.Help;
+
+public class Response
+{
+    public required List<HelpDocumentEntity> Releases { get; set; } = [];
+    public required List<HelpDocumentEntity> HowTos { get; set; } = [];
```
jwinston-bn, line 8: We shouldn't be returning entities ever. They have circular references and can easily blow up when serialized.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
```diff
@@ -0,0 +1,22 @@
...
+
+public class HelpDocumentEntity
+{
+    public int Id { get; set; }
+    public required HelpDocumentType Type { get; set; }
+    public required string Title { get; set; }
+    public required string Url { get; set; }
+    
+    [SqlDefaultValue("getutcdate()")]
+    public DateTime Created { get; set; }
```
jwinston-bn, line 11: I'd put `= DateTime.UtcNow` on both of these as well.

src/Aquifer.API/Endpoints/Help/Response.cs
jwinston-bn: In addition to that, while not an issue in this specific case, we could potentially add something sensitive to a table, or otherwise don't want consumers to see, and by sending the whole entity down, any change to the table schema results in the response starting to include that as well. Or vice versa, say we remove a column, and everything continues to work on the back-end but the front-end blows up. If we're mapping to an object we get the compile error that we've removed something required.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
```diff
@@ -0,0 +1,22 @@
+namespace Aquifer.Data.Entities;
+
+public class HelpDocumentEntity
+{
+    public int Id { get; set; }
+    public required HelpDocumentType Type { get; set; }
+    public required string Title { get; set; }
+    public required string Url { get; set; }
```
jwinston-bn, line 8: No need for `required` on entities.

src/Aquifer.API/Endpoints/Help/Endpoint.cs
```diff
@@ -0,0 +1,43 @@
+using Aquifer.Data;
+using Aquifer.Data.Entities;
+using FastEndpoints;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Help;
+
+public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Response>
+{
+    public override void Configure()
```
jwinston-bn, line 10: I would add server and response caching here.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
kaseywright: I would argue that `required` should be in place for the `Title` and `Url` as they are a must and we have no logic for setting a default value.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
kaseywright: Why do we add this if the DB sets the value for us as defined already?

src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+        Get("/help/aquifer-cms/documents");
+    }
+
+    public override async Task HandleAsync(CancellationToken ct)
+    {
+        var helpDocuments = await GetHelpDocumentsAsync(ct);
+        var response = new Response
+        {
+            Releases = helpDocuments.FindAll(x => x.Type == HelpDocumentType.Release),
+            HowTos = helpDocuments.FindAll(x => x.Type == HelpDocumentType.HowTo)
```
jwinston-bn, line 21: We should be using `Where`. The next version of .NET, which comes out in a month, is going to have massive performance improvements to LINQ. Then to avoid calling `ToList()` I would make the properties `IEnumerable` instead of `List`.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
jwinston-bn: Because it's the database's responsibility to enforce this. We don't know how we might want to use and build these entities. The default value for them should be `= null!;`.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
jwinston-bn: The attribute is at the SQL server level and means I don't have to specify the datetime when inserting to the database directly with SQL. But if you instantiate the POCO it will still set a default value of minimum datetime which we would never want. EF may be smart enough to still use the default SQL value when it does the insert, but I think this makes it more explicit that when creating the entity the value has the current utcnow set and doesn't need to be added.

src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Help.Aquifer_cms.Documents;
+
+public class Endpoint(AquiferDbContext dbContext) : EndpointWithoutRequest<Response>
+{
+    public override void Configure()
+    {
+        Get("/help/aquifer-cms/documents");
+        ResponseCache(EndpointHelpers.OneDayInSeconds);
```
jwinston-bn, line 14: Let's do an hour. I would also add a server cache.

```csharp
Options(EndpointHelpers.UnauthenticatedServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
ResponseCache(EndpointHelpers.OneHourInSeconds);
```

src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Response.cs
```diff
@@ -0,0 +1,17 @@
+using Aquifer.Data.Entities;
+
+namespace Aquifer.API.Endpoints.Help.Aquifer_cms.Documents;
+
+public class Response
+{
+    public required List<HelpDocumentResponse> Releases { get; set; } = [];
+    public required List<HelpDocumentResponse> HowTos { get; set; } = [];
```
NateMerritt, line 8:
```suggestion
    public required List<HelpDocumentResponse> Releases { get; init; }
    public required List<HelpDocumentResponse> HowTos { get; init; }
```
You shouldn't need the default empty collection assignment if you switch to `init` instead of `set` which allows an object initializer (rather than only the constructor) to set values.

src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Endpoint.cs
jwinston-bn: Never mind on server cache, it's authenticated.

src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Response.cs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+public class HelpDocumentResponse
+{
+    public required int Id { get; set; }
```
jwinston-bn, line 13: I would remove Id here since it's presumably not used in this case.

src/Aquifer.API/Endpoints/Help/Aquifer-cms/Documents/Response.cs
kaseywright: Good insight. Implemented. TY.

src/Aquifer.Data/Entities/HelpDocumentEntity.cs
```diff
@@ -0,0 +1,23 @@
+namespace Aquifer.Data.Entities;
+
+public class HelpDocumentEntity
+{
+    public int Id { get; set; }
+    public HelpDocumentType Type { get; set; }
+    public string Title { get; set; } = null!;
+    public string Url { get; set; } = null!;
+    public bool Enabled { get; set; } = false;
```
jwinston-bn, line 9: Default value for bool is false, so no need to set this.

jwinston-bn (dismissed):

jwinston-bn (approved):

PR #490, opened 2024-10-15
Use manual swagger tagging in order to get more control over route groupings.

NateMerritt:
The Resources/Collections and Resources/Types sub-groups are now grouped together using manual swagger tags.  I also addressed missing response status code documentation and made naming and punctuation consistent on the endpoint summaries.

Nested groups are [supported in Redocly](https://redocly.com/docs/developer-portal/configuration/sidebar-nav#groups) but [FastEndpoints doesn't support them](https://fast-endpoints.com/docs/swagger-support#swagger-operation-tags).  Swagger only supports hierarchical groups via a plugin but NSwag doesn't allow supplying that plugin (yet): https://github.com/RicoSuter/NSwag/issues/4921 is addressed.

### DIFF
--- a/src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
+++ b/src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
@@ -15,10 +15,13 @@ public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, Response>
     {
         Get("/bibles/{BibleId}/alignments/greek");
         Options(EndpointHelpers.UnauthenticatedServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
-        Description(d => d.ProducesProblemFE());
+        Description(d => d
+            .WithTags("Bibles")
+            .ProducesProblemFE()
+            .ProducesProblemFE(404));
         Summary(s =>
         {
-            s.Summary = "Gets a Bible's Greek alignment information.";
+            s.Summary = "Get a Bible's Greek alignment information.";
             s.Description =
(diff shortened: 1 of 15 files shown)


src/Aquifer.Public.API/Endpoints/Bibles/Alignments/Greek/Endpoint.cs
```diff
@@ -15,10 +15,13 @@ public override void Configure()
     {
         Get("/bibles/{BibleId}/alignments/greek");
         Options(EndpointHelpers.UnauthenticatedServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
-        Description(d => d.ProducesProblemFE());
+        Description(d => d
+            .WithTags("Bibles")
```
NateMerritt, line 19: We _could_ move this into a constant in a shared location but being that the docs will break if it is incorrect and the routes are not using constants this seemed fine.

src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
```diff
@@ -39,6 +42,8 @@ to the same collection. A resource type is the category to which a collection of
                                   title of the collection to which it belongs is "Bible Dictionary (Tyndale)", and the resource type
                                   is "Dictionary".
                                   """;
+                td["Resources/Collections"] = "Endpoints for retrieving collections of resources.";
+                td["Resources/Types"] = "Endpoints for retrieving the different types of resource collections and resources.";
```
NateMerritt, line 46: Suggestions for a better description are welcome.

jwinston-bn (approved):

PR #492, opened 2024-10-17
Bib 750 add admin help page to aquifer admin

kaseywright:


src/Aquifer.Data/Entities/HelpDocumentEntity.cs
```diff
@@ -7,6 +7,7 @@ public class HelpDocumentEntity
     public string Title { get; set; } = null!;
     public string Url { get; set; } = null!;
     public bool Enabled { get; set; }
+    public string? ThumbnailUrl { get; set; } = null!;
```
jwinston-bn, line 10: The `= null!` is to tell the compiler that you promise you'll set a value. In this case, since it's nullable anyway, it's not needed.

jwinston-bn (approved):

PR #494, opened 2024-10-17
Bib 758 add resource history on edit page

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Resources/Content/Versions/History/Endpoint.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Aquifer.API.Endpoints.Resources.Content.Versions.History;
+
+public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, List<Response>>
+{
+    public override void Configure()
+    {
+        Get("/resources/content/versions/history/{VersionId}");
```
jwinston-bn, line 13: This History folder should go inside the Get folder, because the path already has an individual version on it, and now you're asking for something inside that. So `/resources/content/versions/{VersionId}/history`.

I'm not sure if history is quite right in this case though, since it's an assignment and status history, and status doesn't even live on the version, it lives on ResourceContent. Assignment feels like it could be a type of status for the item. I don't know if this is best, but I would lean into something like `status-history` rather than just history.

src/Aquifer.API/Endpoints/Resources/Content/Versions/History/Endpoint.cs
```diff
@@ -0,0 +1,39 @@
...
+                Event = h.Status.GetDisplayName(),
+                DateOfEvent = h.Created
+            })
+            .ToListAsync(ct);
+
+        var userHistory = await dbContext.ResourceContentVersionAssignedUserHistory
+            .Where(h => h.ResourceContentVersionId == request.VersionId)
+            .Select(h => new Response
+            {
+                Event = h.AssignedUser != null ? $"{h.AssignedUser.FirstName} {h.AssignedUser.LastName}" : "",
```
jwinston-bn, line 31: Why don't you just put `h.AssignedUserId != null` in the `Where` clause?

src/Aquifer.API/Endpoints/Resources/Content/Versions/History/Endpoint.cs
jwinston-bn: I forgot the status history is on the version, even though the status itself is on the content.

src/Aquifer.API/Endpoints/Resources/Content/Versions/History/Endpoint.cs
```diff
@@ -0,0 +1,39 @@
...
+                DateOfEvent = h.Created
+            })
+            .ToListAsync(ct);
+
+        userHistory.RemoveAll(h => h.Event.IsNullOrEmpty());
+        Response = [.. statusHistory.Concat(userHistory).OrderBy(h => h.DateOfEvent)];
```
jwinston-bn, line 37: Ticket is updated to make this descending.

src/Aquifer.API/Endpoints/Resources/Content/Versions/History/Endpoint.cs
elisabeth-shakespeare: Because if there's a more complicated way to do it I'm sure I'd find a way.

src/Aquifer.API/Endpoints/Resources/Content/Versions/Get/StatusHistory/Endpoint.cs
```diff
@@ -0,0 +1,39 @@
...
+        var userHistory = await dbContext.ResourceContentVersionAssignedUserHistory
+            .Where(h => h.ResourceContentVersionId == request.VersionId && h.AssignedUser != null)
+            .Select(h => new Response
+            {
+                Event = $"{h.AssignedUser!.FirstName} {h.AssignedUser.LastName}",
+                DateOfEvent = h.Created
+            })
+            .ToListAsync(ct);
+
+        userHistory.RemoveAll(h => h.Event.IsNullOrEmpty());
```
jwinston-bn, line 36: Is this needed?

jwinston-bn: Also, for future reference, you should use `string.IsNullOrEmpty(h.Event)`. You've brought in an extension method from `Microsoft.IdentityModel.Tokens` for this.

jwinston-bn (approved):

PR #495, opened 2024-10-17
Limit chapter calculation to only chapters in the given book.

NateMerritt:
When passages spanned multiple books the start chapter could be greater than the end chapter.  This updates the calculation of the chapter range to use the first or the last chapter of the requested book instead when this happens.

See https://biblionexusworkspace.slack.com/archives/C0662HZ7RKM/p1729175333040109 for the bug report.

### DIFF
--- a/src/Aquifer.API/Modules/Resources/LanguageResources/LanguageResourcesEndpoints.cs
+++ b/src/Aquifer.API/Modules/Resources/LanguageResources/LanguageResourcesEndpoints.cs
@@ -31,6 +31,10 @@ public static class LanguageResourcesEndpoints
             return TypedResults.BadRequest("parentResourceIds query param must be specified");
         }
 
+        var lastChapterInBook = await dbContext.BibleBookChapters
+            .Where(bbc => bbc.BibleBook.BibleId == 1 && bbc.BibleBook.Number == bookId)
+            .MaxAsync(bbc => bbc.Number, cancellationToken);
+
         var parentResourceEntities = await dbContext.ParentResources
             .Where(pr => parentResourceIds.Contains(pr.Id))
             .ToListAsync(cancellationToken);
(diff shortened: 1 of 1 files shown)


src/Aquifer.API/Modules/Resources/LanguageResources/LanguageResourcesEndpoints.cs
```diff
@@ -123,24 +137,32 @@ public static async Task<Results<Ok<ResourceItemsByChapterResponse>, NotFound, B
                 Contents = g
                     .Select(item => item.Content)
                     .DistinctBy(item => item.ContentId)
+                    .OrderBy(item => item.ContentId)
```
NateMerritt, line 140: This isn't necessary for the bug fix but having a consistent ordering here made comparisons easier when testing.

src/Aquifer.API/Modules/Resources/LanguageResources/LanguageResourcesEndpoints.cs
```diff
@@ -31,6 +31,10 @@ public static async Task<Results<Ok<ResourceItemsByChapterResponse>, NotFound, B
             return TypedResults.BadRequest("parentResourceIds query param must be specified");
         }
 
+        var lastChapterInBook = await dbContext.BibleBookChapters
```
NateMerritt, line 34: Fetching this isn't technically necessary unless we have passages returned with a different book ID than the requested `bookId` but it was simpler to just always fetch it here.  I could refactor to conditionally run this if desired.

jwinston-bn: I would make this conditional if we can, because passages with a different start and end book id is going to be extremely rare.

NateMerritt: I updated to conditionally fetch it.

jwinston-bn (approved):

PR #496, opened 2024-10-22
Add Reviewer role

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Users/Create/Validator.cs
```diff
@@ -18,8 +18,8 @@ public Validator()
         RuleFor(x => x.Role).Must(role =>
         {
             var userService = Resolve<IUserService>();
-            return userService.HasPermission(PermissionName.CreateUser) || role == UserRole.Editor;
-        }).WithMessage("Managers can only create editors within their own company.");
+            return userService.HasPermission(PermissionName.CreateUser) || role == UserRole.Editor || role == UserRole.Reviewer;
```
elisabeth-shakespeare, line 21: Need to check the disable endpoint, too

PR #497, opened 2024-10-22
Send aquifer workspace email notifications.

NateMerritt:
This PR uses SendGrid's templating engine to send three different notification emails: https://mc.sendgrid.com/dynamic-templates.

I reworked existing practices for consuming Azure Storage queues which required adding some utility code that should rarely be touched in order to greatly simplify publishing and subscribing (see comments in the PR where I explain new patterns).


src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
```diff
@@ -0,0 +1,174 @@
...
+            return;
+        }
+
+        var commenterUser = await _dbContext.Users
+            .SingleAsync(u => u.Id == resourceCommentData.UserId, ct);
+
+        var templatedEmail = new TemplatedEmail(
+            From: NotificationsHelper.NotificationSenderEmailAddress,
+            // Template Designer: https://mc.sendgrid.com/dynamic-templates/d-ea84b461ed0f439589098053f8810189/version/26393e82-7d17-4b8f-a37c-cb20f62d8802/editor
+            TemplateId: "d-ea84b461ed0f439589098053f8810189",
```
NateMerritt, line 110: See https://biblionexusworkspace.slack.com/archives/C05K5N7F44Q/p1729881033556149 for design review of this email template.

src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
```diff
@@ -0,0 +1,174 @@
...
+            .Include(projectEntity => projectEntity.CompanyLeadUser)
+            .FirstAsync(p => p.Id == message.ProjectId, ct);
+
+        var companyLeadUser = project.CompanyLeadUser;
+        if (companyLeadUser is not null && companyLeadUser.AquiferNotificationsEnabled)
+        {
+            var templatedEmail = new TemplatedEmail(
+                From: NotificationsHelper.NotificationSenderEmailAddress,
+                // Template Designer: https://mc.sendgrid.com/dynamic-templates/d-7760ec3b5ce34b179384d4783cc1bd81/version/e83075a3-ba61-4d42-922f-9fd5df4ee45c/editor
+                TemplateId: "d-7760ec3b5ce34b179384d4783cc1bd81",
```
NateMerritt, line 39: See https://biblionexusworkspace.slack.com/archives/C05K5N7F44Q/p1729881062250939 for design review of this email template.

src/Aquifer.Jobs/Managers/SendResourceAssignmentNotifications.cs
```diff
@@ -0,0 +1,101 @@
...
+            .Select(uh => uh.AssignedUser)
+            .DistinctBy(u => u.Id)
+            .ToDictionary(au => au.Id);
+
+        var templatedEmails = userHistories
+            .GroupBy(uh => uh.AssignedUser.Id)
+            .Select(userGrouping => new TemplatedEmail(
+                From: NotificationsHelper.NotificationSenderEmailAddress,
+                // Template Designer: https://mc.sendgrid.com/dynamic-templates/d-d85f76c6b4d344f5bc8b90b27cc40cc3/version/b6955fec-6f2e-41f7-a9f5-fb695a5b8ed7/editor
+                TemplateId: "d-d85f76c6b4d344f5bc8b90b27cc40cc3",
```
NateMerritt, line 58: See https://biblionexusworkspace.slack.com/archives/C05K5N7F44Q/p1729881075224309 for design review of this email template.

src/Aquifer.API/Endpoints/Resources/Content/Batch/Text/Endpoint.cs
```diff
@@ -46,6 +46,6 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
     public override async Task OnAfterHandleAsync(Request req, List<Response> res, CancellationToken ct)
     {
         const string endpointId = "resources-content-batch-text";
-        await trackingService.TrackAsync(HttpContext, req.Ids, endpointId);
+        await trackingService.TrackAsync(HttpContext, req.Ids, endpointId, cancellationToken: ct);
```
NateMerritt, line 49: I made the resource tracking job conform to the new patterns I introduced.  After this PR merges I'm also planning to move the existing Jobs into the new sub-folders (`Subscribers` and `Managers`).

src/Aquifer.Common/Clients/SendGridClient.cs
```diff
@@ -1,49 +0,0 @@
-using SendGrid;
```
NateMerritt, line 1: I moved the `SendGridClient` to the Aquifer.Jobs project and pretty much rewrote it.  Now only the `EmailSubscriber` is allowed to talk to SendGrid and the `EmailService` will put emails onto a queue.

src/Aquifer.Common/Jobs/MessagesJsonSerializer.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Newtonsoft.Json;
```
NateMerritt, line 1: Newtonsoft.Json correctly serializes and then deserialized anonymous types into an `object`.  System.Text.Json fails at this.  After many workarounds @jwinston-bn convinced me to just use Newtonsoft.Json here.

src/Aquifer.Common/Jobs/QueueClientFactory.cs
```diff
@@ -0,0 +1,56 @@
...
+using Azure.Storage.Queues;
+
+namespace Aquifer.Common.Jobs;
+
+public interface IQueueClientFactory
+{
+    public Task<QueueClient> GetQueueClientAsync(string queueName, CancellationToken ct);
+}
+
+public sealed class QueueClientFactory(QueueConfigurationOptions _queueConfigurationOptions, IAzureClientService _azureClientService)
```
NateMerritt, line 12: The goal of this class is to create a factory that allows for lazily ensuring that a queue exists upon accessing a queue client for publishing purposes.  This will only happen once per queue per app service lifetime and only if we try to publish to that queue.  Previous code did this web request to ensure the queue exists in the class constructor which isn't a best practice.

src/Aquifer.Common/Jobs/Queues.cs
```diff
@@ -0,0 +1,15 @@
...
+    // email
+    public const string SendEmail = "send-email";
+    public const string SendTemplatedEmail = "send-templated-email";
+
+    // notifications
+    public const string SendProjectStartedNotification = "send-project-started-notification";
+    public const string SendResourceCommentCreatedNotification = "send-resource-comment-created-notification";
+
+    // resource tracking
+    public const string TrackResourceContentRequest = "track-resource-content-request";
```
NateMerritt, line 14: This is a breaking change because the live queue currently has `dev` or `prod` in the name.  I'll double check that there are no abandoned queue messages after deployment and will manually replay them on the correct queue if needed.

src/Aquifer.Common/Services/NotificationService.cs
```diff
@@ -0,0 +1,29 @@
...
+
+public sealed class NotificationService(IQueueClientFactory _queueClientFactory) : INotificationService
+{
+    public async Task SendProjectStartedNotificationAsync(int projectId, CancellationToken cancellationToken)
+    {
+        var message = new SendProjectStartedNotificationMessage(projectId);
+
+        var queueClient = await _queueClientFactory.GetQueueClientAsync(Queues.SendProjectStartedNotification, cancellationToken);
+        await queueClient.SendMessageAsync(message, cancellationToken: cancellationToken);
+    }
```
NateMerritt, line 20: Hopefully this simple and easy to use syntax demonstrates why I wrote all of the utility code in this PR for publishing messages.

src/Aquifer.Data/Migrations/20241023135727_AddJobHistoryTable.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
```
NateMerritt, line 1: I already ran this migration in Dev.

src/Aquifer.Data/Migrations/20241023141224_AddAquiferNotificationsEnabledColumnToUsersTable.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
NateMerritt, line 1: I already ran this migration in Dev.

src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
```diff
@@ -3,12 +3,21 @@
 public class ConfigurationOptions
 {
     public bool IsDevelopment { get; init; }
+    public required string AquiferAdminBaseUri { get; init; }
     public required string AquiferApiBaseUri { get; init; }
     public required string KeyVaultUri { get; init; }
+    public required ConnectionStringOptions ConnectionStrings { get; init; }
```
NateMerritt, line 9: Connection strings are already present in the configuration; this allows accessing them via the same patterns as the API projects.

src/Aquifer.Jobs/Managers/SendResourceAssignmentNotifications.cs
```diff
@@ -0,0 +1,101 @@
...
+    IOptions<ConfigurationOptions> _configurationOptions,
+    AquiferDbContext _dbContext,
+    IEmailService _emailService,
+    ILogger<SendResourceAssignmentNotifications> _logger)
+{
+    private const string _everyTenMinutesCronSchedule = "0 */10 * * * *";
+    private const string _tenSecondDelayInterval = "00:00:10";
+
+    [Function(nameof(SendResourceAssignmentNotifications))]
+    [FixedDelayRetry(maxRetryCount: 1, delayInterval: _tenSecondDelayInterval)]
```
NateMerritt, line 22: By default timer jobs don't retry.  There is also an `ExponentialBackoffRetry` (which might be useful for other jobs) but in this case because we run every 10 minutes anyway I decided that a single retry was appropriate to handle the rare transient network error.

src/Aquifer.Jobs/Program.cs
```diff
@@ -1,34 +1,47 @@
...
         services.AddHttpClient<IIpAddressLookupHttpClient, IpAddressLookupHttpClient>();
         services.AddSingleton<IAzureKeyVaultClient, AzureKeyVaultClient>();
-        services.AddSingleton<ISendGridClient, SendGridClient>();
-        services.AddApplicationInsightsTelemetryWorkerService();
-        services.ConfigureFunctionsApplicationInsights();
-        services.AddAzureClient(context.Configuration.Get<ConfigurationOptions>()!.IsDevelopment);
-        services.AddDbContext<AquiferDbContext>(options =>
-            options.UseSqlServer(connectionString, providerOptions => providerOptions.EnableRetryOnFailure(3)));
+        services.AddSingleton<IEmailService, EmailService>();
+        services.AddKeyedSingleton<IEmailService, SendGridEmailService>(nameof(SendGridEmailService));
```
NateMerritt, line 41: The `EmailSubscriber` is the only class that needs to use the `SendGridEmailService` implementation of `IEmailService`.  That said, if we ever need to bypass the queue or want to move away from that strategy the `EmailService` and `SendGridEmailService` both use the same interface.

src/Aquifer.Jobs/Program.cs
```diff
@@ -1,34 +1,47 @@
...
         services.AddOptions<ConfigurationOptions>().Bind(context.Configuration);
 
-        var connectionString = context.Configuration.GetConnectionString("BiblioNexusDb");
+        var configuration = context.Configuration.Get<ConfigurationOptions>()
+            ?? throw new InvalidOperationException($"Unable to bind {nameof(ConfigurationOptions)}.");
+
+        services.AddDbContext<AquiferDbContext>(options => options
+            .UseSqlServer(configuration.ConnectionStrings.BiblioNexusDb, providerOptions => providerOptions.EnableRetryOnFailure(3))
+            .EnableSensitiveDataLogging(sensitiveDataLoggingEnabled: configuration.IsDevelopment)
+            .UseQueryTrackingBehavior(QueryTrackingBehavior.NoTracking));
```
NateMerritt, line 28: I missed adding EF `NoTracking` for Aquifer.Jobs previously.  All but one existing query is read-only and I had already added `AsTracking()` to the write query.

src/Aquifer.Jobs/Services/SendGridEmailService.cs
```diff
@@ -0,0 +1,112 @@
...
+using SendGrid.Helpers.Mail;
+using EmailAddress = Aquifer.Common.Services.EmailAddress;
+
+namespace Aquifer.Jobs.Services;
+
+public class SendGridEmailService : IEmailService
+{
+    private readonly SendGridClient _sendGridClient;
+
+    // TODO move this to DI and don't make a network call in a constructor
```
NateMerritt, line 13: I didn't change this pre-existing strategy of fetching a key vault secret in the constructor in this PR but it looks like we could refactor to do this as part of DI instead.

src/Aquifer.Jobs/Subscribers/EmailSubscriber.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    [Function(nameof(SendTemplatedEmail))]
+    public async Task SendTemplatedEmail(
+        [QueueTrigger(Queues.SendTemplatedEmail)]
+        QueueMessage message,
+        CancellationToken ct)
+    {
+        var templatedEmail = message.Deserialize<TemplatedEmail, EmailSubscriber>(_logger);
+
+        // if `Subject` is present in the dynamic template data then add an environment specific prefix to it
```
NateMerritt, line 49: We're planning to have Local, Dev, and QA email sends so I ensured that the email subject includes the environment (outside of prod).

src/Aquifer.Jobs/Subscribers/EmailSubscriber.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        try
+        {
+            await _emailService.SendEmailAsync(
+                email with { Subject = $"{_configurationOptions.Value.Email.SubjectPrefix}{email.Subject}" },
+                ct);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Unable to send email: \"{Email}\"", message.MessageText);
```
NateMerritt, line 34: Previously failed email sends were logging at error in the telemetry client: https://github.com/BiblioNexusStudio/aquifer-server/blob/ecd3030ae68e27cd9f9cced63886218f60dc3818/src/Aquifer.Jobs/SendNewContentEmails.cs#L86.  Is that necessary here?  It's not clear to me when using a telemetry client is necessary and what we get for free with standard exception and `ILogger<T>`.

src/Aquifer.Jobs/local.settings.example.json
```diff
@@ -1,13 +1,13 @@
 {
     "IsEncrypted": false,
     "ConnectionStrings": {
-        "BiblioNexusDb": "Server=localhost;Database=biblionexus-local-db;User Id=sa;Password=Password!;TrustServerCertificate=true;"
+        "BiblioNexusDb": "Server=localhost;Database=biblionexus-local-db;User Id=sa;Password=Password!;TrustServerCertificate=true;",
+        "AzureStorageAccount": ""
```
NateMerritt, line 5: I set up this connection string and two new environment variables below on both the dev and prod Jobs instances.

src/Aquifer.API/Endpoints/Projects/Start/Endpoint.cs
```diff
@@ -53,6 +58,8 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
 
         await dbContext.SaveChangesAsync(ct);
 
+        await notificationService.SendProjectStartedNotificationAsync(project.Id, ct);
```
jwinston-bn, line 61: I think we should just remove this for now, because the post-processing queue for the project start is going to handle this.

src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
jwinston-bn: This feels like it should be a config value, not hard-coded magic string.

src/Aquifer.Jobs/Managers/SendResourceAssignmentNotifications.cs
```diff
@@ -0,0 +1,101 @@
...
+#pragma warning disable IDE0060 // Remove unused parameter: A (non-discard) TimerInfo parameter is required for correct Azure bindings.
+    public async Task RunAsync([TimerTrigger(_everyTenMinutesCronSchedule)] TimerInfo timerInfo, CancellationToken ct)
+#pragma warning restore IDE0060 // Remove unused parameter
+    {
+        var jobHistory = await _dbContext.JobHistory
+            .AsTracking()
+            .SingleAsync(jh => jh.JobId == JobId.SendResourceAssignmentNotifications, ct);
+
+        // Possible Improvement: Only send notifications when a user is assigned to the *most recent* ResourceContentVersion.
+        var userHistories = await _dbContext.ResourceContentVersionAssignedUserHistory
```
jwinston-bn, line 32: How does this handle the situation where a user shouldn't receive an email on assignment? For example, project start, the company lead shouldn't get an email for the items assigned, because they just get the project started email?

src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
```diff
@@ -0,0 +1,174 @@
...
+            await _emailService.SendEmailAsync(templatedEmail, ct);
+
+            _logger.LogInformation(
+                "Project started notification sent for Project ID {ProjectId} to User ID {UserId}.",
+                project.Id,
+                companyLeadUser.Id);
+        }
+        else
+        {
+            _logger.LogWarning(
```
jwinston-bn, line 59: This should _not_ happen. I'd make this an error so we're getting alerted to it right away.

src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            _logger.LogInformation(
+                "Comment ID {CommentId} was added to Resource Content ID {ResourceContentId} by User ID {CommenterUserId} but no email was sent because only the commenter has previous resource assignment.",
+                resourceCommentData.CommentId,
+                resourceCommentData.ResourceContentId,
+                resourceCommentData.UserId);
+            return;
+        }
+
+        var commenterUser = await _dbContext.Users
```
jwinston-bn, line 104: Would it be more efficient to join into Users from  `GetResourceCommentDataAsync` to avoid the extra database trip here?

src/Aquifer.Jobs/NotificationsHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static readonly EmailAddress NotificationToEmailAddress = new("no-reply@aquifer.bible", "Aquifer");
+
+    public static EmailAddress GetEmailAddress(UserEntity userEntity)
+    {
+        return new EmailAddress(userEntity.Email, GetUserFullName(userEntity));
+    }
+
+    public static string GetUserFullName(UserEntity userEntity)
+    {
+        return $"{userEntity.FirstName} {userEntity.LastName}";
```
jwinston-bn, line 18: It's a little awkward, but there is the `UserDto` class already that does this.

src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
```diff
@@ -0,0 +1,174 @@
...
+        var previouslyAssignedUsers = await _dbContext.ResourceContentVersionAssignedUserHistory
+            .Where(rcvauh =>
+                rcvauh.ResourceContentVersion.ResourceContentId == resourceCommentData.ResourceContentId &&
+                rcvauh.AssignedUserId != null &&
+                rcvauh.AssignedUserId.Value != resourceCommentData.UserId)
+            .Select(rcvauh => rcvauh.AssignedUserId)
+            .Distinct()
+            .Join(
+                _dbContext.Users
+                    .Where(u => u.Id == resourceCommentData.UserId && u.AquiferNotificationsEnabled)
```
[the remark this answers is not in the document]
NateMerritt: Ah, good catch.  I'll add that here and also do the same for the other places I'm grabbing users.

src/Aquifer.Jobs/Subscribers/NotificationSubscriber.cs
NateMerritt: I'll admit that this was a product of my development process.  At one point I was fetching this user when also fetching the resource assignment users (which was the most efficient) but had to stop doing that as things were refactored.

The `GetResourceCommentDataAsync` method uses Dapper (I tried writing it in EF and got stumped even though it seems like it should be easy). I didn't want to make some kind of Dapper/EF interop mapping nor did I want to use the EF `UserEntity` as a destination object for the Dapper query because my assumption when seeing an EF entity is that it was pulled via EF and might have tracking, etc.  I also wanted a `UserEntity` in order to have common code that knows how to generate a full name.

So yes, this could be more efficiently done.  I will update to also fetch the commenter's name in the Dapper call which is really all I want for the email.

src/Aquifer.Jobs/NotificationsHelper.cs
NateMerritt: `UserDto` is in Aquifer.Api and thus can't be referenced by Aquifer.Jobs.  Are you suggesting moving `UserDto` to Aquifer.Common?

src/Aquifer.API/Endpoints/Projects/Start/Endpoint.cs
NateMerritt: Is there a strong objection?  I'd ideally like to ship all of the email notifications at the same time.  After we do the translation work then this endpoint will kick off a different queue and that queue will kick off the notification.  It's only a few lines of code moving around.

jwinston-bn (approved):

PR #498, opened 2024-10-22
Bib 757 add assignment pathway and change status flow

kaseywright:


kaseywright:
Created the PR as draft as this touches the DB by adding a column to the `ResourceContentVersions` table

src/Aquifer.API/Endpoints/Resources/Content/AssignEditor/Endpoint.cs
```diff
@@ -36,6 +36,19 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
             ThrowError("Unable to assign to a user outside your company");
         }
 
+        if (request.AssignedReviewerUserId is not null) {
+            var reviewer = await dbContext.Users 
+                .AsTracking()
```
jwinston-bn, line 41: I think we can remove the `AsTracking()` here since it's a quick check and we don't do anything with the response other than check if it's null or not.

jwinston-bn (approved):

PR #499, opened 2024-10-23
Use Directory.Build.props files for solution wide project settings.

NateMerritt:
Some projects in the solution were running style analysis on build but others weren't.  Some were treating warnings as errors but others weren't.  This caused me to miss [a build error](https://github.com/BiblioNexusStudio/aquifer-server/actions/runs/11481872205/job/31953579431) caught by CI but that Visual Studio didn't notice locally when building.

This PR implements solution-wide settings across projects where possible and moves the settings from the project files into the solution-wide `Directory.Build.props` file(s).

This change required fixing or disabling warnings and style errors that were previously undetected in certain projects.  Note also that I added the `.editorconfig` file to the solution which caused some settings to be added automatically via VS.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -258,9 +258,6 @@ resharper_csharp_continuous_indent_multiplier = 1
 resharper_csharp_max_invocation_arguments_on_line = 6
 resharper_csharp_wrap_list_pattern = chop_if_long
 
-[src/Aquifer.Data/Migrations/*.cs]
-generated_code = true
-
 [*.{cs,vb}]
 dotnet_style_coalesce_expression = true:error
 dotnet_style_null_propagation = true:error
@@ -272,4 +269,3 @@ dotnet_style_operator_placement_when_wrapping = beginning_of_line
 dotnet_style_prefer_collection_expression = when_types_loosely_match:suggestion
 tab_width = 4
 indent_size = 4
-end_of_line = crlf
--- a/.editorconfig
+++ b/.editorconfig
@@ -102,12 +102,12 @@ csharp_style_var_when_type_is_apparent = true:suggestion
 
 # Expression-bodied members
 csharp_style_expression_bodied_accessors = true:suggestion
-csharp_style_expression_bodied_constructors = false
+csharp_style_expression_bodied_constructors = false:silent
 csharp_style_expression_bodied_indexers = true:suggestion
 csharp_style_expression_bodied_lambdas = true:suggestion
-csharp_style_expression_bodied_local_functions = false
-csharp_style_expression_bodied_methods = false
-csharp_style_expression_bodied_operators = false
+csharp_style_expression_bodied_local_functions = false:silent
+csharp_style_expression_bodied_methods = false:silent
+csharp_style_expression_bodied_operators = false:silent
 csharp_style_expression_bodied_properties = true:suggestion
 
 # Pattern matching preferences
@@ -128,8 +128,8 @@ csharp_style_prefer_readonly_struct = true
 csharp_style_prefer_readonly_struct_member = true
 
 # Code-block preferences
-csharp_prefer_braces = true
-csharp_prefer_simple_using_statement = true
+csharp_prefer_braces = true:silent
+csharp_prefer_simple_using_statement = true:suggestion
 csharp_style_namespace_declarations = file_scoped:error
 csharp_style_prefer_method_group_conversion = true:suggestion
 csharp_style_prefer_top_level_statements = true:error
@@ -260,3 +260,16 @@ resharper_csharp_wrap_list_pattern = chop_if_long
 
 [src/Aquifer.Data/Migrations/*.cs]
 generated_code = true
+
+[*.{cs,vb}]
+dotnet_style_coalesce_expression = true:error
+dotnet_style_null_propagation = true:error
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
+dotnet_style_prefer_auto_properties = true:silent
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:silent
+dotnet_style_operator_placement_when_wrapping = beginning_of_line
+tab_width = 4
+indent_size = 4
+end_of_line = crlf
+dotnet_style_prefer_collection_expression = when_types_loosely_match:suggestion
--- a/.editorconfig
+++ b/.editorconfig
@@ -102,12 +102,12 @@ csharp_style_var_when_type_is_apparent = true:suggestion
 
 # Expression-bodied members
 csharp_style_expression_bodied_accessors = true:suggestion
-csharp_style_expression_bodied_constructors = false:silent
+csharp_style_expression_bodied_constructors = false
 csharp_style_expression_bodied_indexers = true:suggestion
 csharp_style_expression_bodied_lambdas = true:suggestion
-csharp_style_expression_bodied_local_functions = false:silent
-csharp_style_expression_bodied_methods = false:silent
-csharp_style_expression_bodied_operators = false:silent
+csharp_style_expression_bodied_local_functions = false
+csharp_style_expression_bodied_methods = false
+csharp_style_expression_bodied_operators = false
 csharp_style_expression_bodied_properties = true:suggestion
 
 # Pattern matching preferences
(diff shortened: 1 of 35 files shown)


.editorconfig
```diff
@@ -260,3 +260,16 @@ resharper_csharp_wrap_list_pattern = chop_if_long
...
+dotnet_style_coalesce_expression = true:error
+dotnet_style_null_propagation = true:error
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
+dotnet_style_prefer_auto_properties = true:silent
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:silent
+dotnet_style_operator_placement_when_wrapping = beginning_of_line
+tab_width = 4
+indent_size = 4
+end_of_line = crlf
```
NateMerritt, line 274: Do we use crlf or lf?  I think this might have just pulled my local settings and populated the `.editorconfig` file.

Directory.Build.props
```diff
@@ -0,0 +1,20 @@
+<Project>
+
+    <PropertyGroup>
+        <LangVersion>latest</LangVersion>
+        <Nullable>enable</Nullable>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <GitHubOrganization>BiblioNexusStudio</GitHubOrganization>
```
NateMerritt, line 10: The settings here and below are primarily for NuGet packaging but they don't hurt to add, especially for an open-source repo.

Directory.Build.props
```diff
@@ -0,0 +1,20 @@
...
+        <RepositoryName>aquifer-server</RepositoryName>
+        <PackageLicenseExpression>MIT</PackageLicenseExpression>
+        <RepositoryUrl>https://github.com/$(GitHubOrganization)/$(RepositoryName).git</RepositoryUrl>
+        <Authors>BiblioNexus</Authors>
+        <Copyright>Copyright $(Authors)</Copyright>
+        <IsPackable>false</IsPackable>
+        <IsTestProject>false</IsTestProject>
```
NateMerritt, line 17: This will be overridden only for test projects (see below).

.editorconfig
jwinston-bn: Mine have this line in it

.editorconfig
```diff
@@ -102,12 +102,12 @@ csharp_style_var_when_type_is_apparent = true:suggestion
 
 # Expression-bodied members
 csharp_style_expression_bodied_accessors = true:suggestion
-csharp_style_expression_bodied_constructors = false
+csharp_style_expression_bodied_constructors = false:silent
```
jwinston-bn, line 105: What's the difference between `:silent` and not having the line at all?

NateMerritt: I'm going to revert the `*:silent` changes here.  I don't know why VS added them and I just realized that it can cause errors to show up in the error list when building but then you don't see the analyzer squiggle in the editor which is confusing.

src/Aquifer.Data/EventHandlers/ResourceStatusChangeHandler.cs
```diff
@@ -22,6 +22,7 @@ public static async Task HandleAsync(DbContextOptions<AquiferDbContext> dbContex
             .ToList()
             .ForEach(x =>
             {
+#pragma warning disable IDE0010 // Add missing cases to switch statement
```
jwinston-bn, line 25: Is this something we can just disable globally? Would we ever care about this if the functionality didn't need it?

src/Aquifer.API/Aquifer.API.csproj
```diff
@@ -2,11 +2,6 @@
 
     <PropertyGroup>
         <TargetFramework>net8.0</TargetFramework>
-        <Nullable>enable</Nullable>
```
jwinston-bn, line 5: Will we still get IDE warnings with this not here any more? Like sometimes I'll get the warning from the IDE before build that something could be null that isn't allowed to be.

src/Aquifer.Data/EventHandlers/ResourceStatusChangeHandler.cs
NateMerritt: I think it's better to explicitly ignore the warning when we are sure it's not needed.  Example:
1. We write code that handles all enum values in a switch statement.
2. Someone adds a new enum value.  The existing switch statement is not updated to handle it.
3. The code explodes at runtime or doesn't have correct execution.

This warning forces you to explicitly add handling for the new enum value.  If we wanted to we could handling here instead of disabling the warning by including all of the other values with an empty case statement (but an empty case statement also can give warnings).

src/Aquifer.API/Aquifer.API.csproj
NateMerritt: Yes, we will still get the warnings because this project still defines `<Nullable>enable</Nullable>`.  All projects in the solution now inherit properties defined in `Directory.Build.Props`, this setting is defined [there](https://github.com/BiblioNexusStudio/aquifer-server/pull/499/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eR5), and thus it is redundant to continue including the setting at the project level (unless we want to override the solution defaults).  And now any new projects in the solution will get this setting for free and won't have to remember to include this line.

jwinston-bn: Okay, I just wanted to make sure that other file (since it has Build in the name), wasn't only applying on build.

.editorconfig
```diff
@@ -260,3 +260,16 @@ resharper_csharp_wrap_list_pattern = chop_if_long
...
+dotnet_style_null_propagation = true:error
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
+dotnet_style_prefer_auto_properties = true
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true
+dotnet_style_operator_placement_when_wrapping = beginning_of_line
+dotnet_style_prefer_collection_expression = when_types_loosely_match:suggestion
+tab_width = 4
+indent_size = 4
+end_of_line = crlf
```
MrGrinst, line 275: We had this in here at one point but it was causing issues cross-platform with git. Removing it lets the committed code live with `lf` which git likes, but then Windows machines will handle it during checkout as `crlf`. That's my understanding of it.

MrGrinst: PR: https://github.com/BiblioNexusStudio/aquifer-server/pull/415

MrGrinst: Also for reference looked at some MS repos to see what they were using for .editorconfig as well as their .gitattributes. Example: https://github.com/dotnet/efcore/blob/main/.editorconfig

NateMerritt: Thanks, I removed the `crlf` setting.  IMO it would be best to update everything to LF rather than have both LF and CRLF depending upon platform (but I'm not going to do that in this PR).  VS will respect this EOL setting (and I'm assuming Rider will as well).

MrGrinst (approved):

src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -20,7 +20,7 @@ public class SendNewContentEmails(
     TelemetryClient telemetryClient)
 {
     [Function(nameof(SendNewContentEmails))]
-    public async Task Run([TimerTrigger("%MarketingEmail:CronSchedule:NewContent%")] TimerInfo timerInfo, CancellationToken ct)
+    public async Task Run([TimerTrigger("%MarketingEmail:CronSchedule:NewContent%")] TimerInfo _, CancellationToken ct)
```
NateMerritt, line 23: This broke Azure Functions bindings.  Fix here: https://github.com/BiblioNexusStudio/aquifer-server/pull/501

src/Aquifer.Data/Entities/ResourceContentVersionEntity.cs
```diff
@@ -32,14 +32,11 @@ public class ResourceContentVersionEntity : IHasUpdatedTimestamp
 
     public ICollection<ResourceContentVersionMachineTranslationEntity> MachineTranslations { get; set; } = [];
 
-    public IEnumerable<ResourceContentVersionStatusHistoryEntity> ResourceContentVersionStatusHistories { get; set; } =
-        new List<ResourceContentVersionStatusHistoryEntity>();
+    public IEnumerable<ResourceContentVersionStatusHistoryEntity> ResourceContentVersionStatusHistories { get; set; } = [];
```
NateMerritt, line 35: This broke resource assignment.  See https://github.com/BiblioNexusStudio/aquifer-server/pull/503 for the fix.

PR #501, opened 2024-10-23
Fix TimerInfo parameter Azure Trigger Function bindings.

NateMerritt:
Azure is dumb and can't bind `_` (the discard parameter) as a parameter name for a `TimerTrigger` parameter.  But it doesn't let you know this until it blows up at runtime and I missed it amongst the other Job logging when testing my previous PR.  See also https://github.com/Azure/azure-functions-dotnet-worker/issues/323.

Logs:
```
[2024-10-23T20:16:27.939Z] The 'SyncApiRequestStatsToStorageTable' function is in error: The binding name _ is invalid. Please assign a valid name to the binding.
[2024-10-23T20:16:27.940Z] The 'SendNewContentEmails' function is in error: The binding name _ is invalid. Please assign a valid name to the binding.
[2024-10-23T20:16:27.940Z] The 'SyncCustomEventsToStorageTable' function is in error: The binding name _ is invalid. Please assign a valid name to the binding.
[2024-10-23T20:16:27.941Z] The 'SyncPageViewsToStorageTable' function is in error: The binding name _ is invalid. Please assign a valid name to the binding.
```

### DIFF
--- a/src/Aquifer.Jobs/SendNewContentEmails.cs
+++ b/src/Aquifer.Jobs/SendNewContentEmails.cs
@@ -20,7 +20,9 @@ public class SendNewContentEmails(
     TelemetryClient telemetryClient)
 {
     [Function(nameof(SendNewContentEmails))]
-    public async Task Run([TimerTrigger("%MarketingEmail:CronSchedule:NewContent%")] TimerInfo _, CancellationToken ct)
+#pragma warning disable IDE0060 // Remove unused parameter: A (non-discard) TimerInfo parameter is required for correct Azure bindings
+    public async Task Run([TimerTrigger("%MarketingEmail:CronSchedule:NewContent%")] TimerInfo timerInfo, CancellationToken ct)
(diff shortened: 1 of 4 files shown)


src/Aquifer.Jobs/SendNewContentEmails.cs
```diff
@@ -20,7 +20,9 @@ public class SendNewContentEmails(
     TelemetryClient telemetryClient)
 {
     [Function(nameof(SendNewContentEmails))]
-    public async Task Run([TimerTrigger("%MarketingEmail:CronSchedule:NewContent%")] TimerInfo _, CancellationToken ct)
+#pragma warning disable IDE0060 // Remove unused parameter: A (non-discard) TimerInfo parameter is required for correct Azure bindings
```
NateMerritt, line 23: I don't like doing this manually for each such parameter but I still think it's preferable to leave IDE0060 enabled.

jwinston-bn (approved):
Wow that's dumb

PR #502, opened 2024-10-24
add limit/offset, resourceCollectionCode and LanguageCode params

elisabeth-shakespeare:


src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,70 @@ public override void Configure()
         {
             s.Summary = "Get resource ids that are new or updated since the given UTC timestamp.";
             s.Description =
-                "For a given UTC timestamp and optional language id, get a list of resource ids that are new or have been updated since the provided timestamp. This is intended for users who are storing Aquifer data locally and want to fetch new content.";
+                "For a given UTC timestamp and optional language id, optional ISO language code, or optional collection code get a list of resource ids that are new or have been updated since the provided timestamp. This is intended for users who are storing Aquifer data locally and want to fetch new content.";
```
jwinston-bn, line 21: Probably can just say "For a given UTC timestamp, get a list..." at this point, since there are multiple parameters.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Request.cs
```diff
@@ -11,4 +13,24 @@ public class Request
...
     public int? LanguageId { get; set; }
+    /// <summary>
+    ///     Optional ISO 639-3 code that is 3 characters in length (e.g. "eng").
+    /// </summary>
+    public string? LanguageCode { get; init; }
+    /// <summary>
+    ///     The number of results to return. Max is 100.
+    /// </summary>
+    [DefaultValue(10)]
+    public int Limit { get; init; } = 10;
```
jwinston-bn, line 24: This is way too low. Let's make it 1000.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Request.cs
```diff
@@ -11,4 +13,24 @@ public class Request
...
+
+    /// <summary>
+    ///     Records to skip if paging through results.
+    /// </summary>
+    [DefaultValue(0)]
+    public int Offset { get; init; }
+    /// <summary>
+    ///     Optional resource collection code to search for. Search by code retrieved from /resources/types endpoint such as
```
jwinston-bn, line 32: Comes from /collections endpoint

jwinston-bn: Also can put ticks around things to get them that code highlighting.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Request.cs
jwinston-bn, line 1: Need to run formatter on this file. Should be line breaks between props and summaries

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,70 @@ public override void Configure()
...
-        Response.Items = await dbContext.ResourceContentVersions
-            .Where(x => (!req.LanguageId.HasValue || x.ResourceContent.LanguageId == req.LanguageId) &&
-                x.IsPublished &&
-                x.Updated >= req.Timestamp)
+        var query = await GetQuery(req, ct);
+        var totalCount = await GetTotalResourceCountAsync(req, query, ct);
+
+        if (totalCount == 0)
+        {
+            await SendOkAsync(new Response(), ct);
```
jwinston-bn, line 32: This would send back 0 for the offset, which may not be the case.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Request.cs
```diff
@@ -11,4 +13,24 @@ public class Request
     ///     Optional LanguageId to search on. If none is provided, updates for all languages will be returned.
     /// </summary>
     public int? LanguageId { get; set; }
```
jwinston-bn, line 15: Should make this and Timestamp `init` only too for consistency

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Validator.cs
```diff
@@ -9,5 +11,12 @@ public Validator()
     {
         RuleFor(x => x.Timestamp).NotEmpty().GreaterThanOrEqualTo(DateTime.UtcNow.AddDays(-90).Date);
         RuleFor(x => x.LanguageId).InclusiveBetween(1, 32);
+        RuleFor(x => x.Offset).GreaterThanOrEqualTo(0);
+        RuleFor(x => x.Limit).InclusiveBetween(0, 100);
+        RuleFor(x => x)
+            .Must(x => (x.LanguageCode is not null && x.LanguageId == default)
+                       || (x is { LanguageId: not null, LanguageCode: null })
+                       || (x.LanguageId == default && x.LanguageCode is null))
+            .WithMessage("Either languageId or languageCode is optional. Cannot use both.");
```
jwinston-bn, line 20: This message is a little confusing. I would just say "Cannot use both languageId and languageCode".

Also, I would be careful about using `default`, because it can cause unexpected behavior sometimes. In this case `default` is null, but if you changed `LanguageId` to `int`, then the default would be `0`. So in this situation it may not matter, but you could potentially create an unexpected error when changing the nullability of a type when you're using `default`.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Validator.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using FastEndpoints;
 using FluentValidation;
 
+// ReSharper disable ArrangeRedundantParentheses
```
jwinston-bn, line 4: What is this for?

elisabeth-shakespeare: On save it takes out the parenthesis, and then it errors saying I need them. And then when I put them in, it takes them out on save, and on and on.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Response.cs
```diff
@@ -11,6 +13,7 @@ public class ResponseContent
     public required ResponseContentUpdateType UpdateType { get; set; }
 
     public required int LanguageId { get; set; }
+    public string LanguageCode { get; set; } = null!;
```
NateMerritt, line 16: Can `LanguageCode` be made `required` and the default `null` assignment be removed?

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,70 @@ public override void Configure()
...
+        }
+
+        return totalCount;
+    }
+
+    private async Task<IQueryable<ResourceContentVersionEntity>> GetQuery(Request req, CancellationToken ct)
+    {
+        var languageId = req.LanguageCode is not null
+            ? await _cachingLanguageService.GetLanguageIdAsync(req.LanguageCode, ct) ?? 0
+            : req.LanguageId;
```
NateMerritt, line 79: I wrote a utility method [here](https://github.com/BiblioNexusStudio/aquifer-server/blob/6007e86325e95b72c915f208d711b77ff78fb060/src/Aquifer.Common/Services/Caching/CachingLanguageService.cs#L82-L83) that is useful for the Public API where both `languageCode` and `languageId` can be passed.  I suggest using it here like this:
```
var (isValidLanguageId, isValidLanguageCode, validLanguageId) = await ValidateLanguageIdOrCodeAsync(req.LanguageId, req.LanguageCode, shouldRequireInput: false, ct);
```

This will validate that either `languageId` or `languageCode` have good data or are both `null` (when `shouldRequireInput` is `false` and will return a `languageId` that is guaranteed to exist (or possibly `null` when `shouldRequireInput` is `false` which is a valid state for this endpoint).

Thus, you could return a bad request if either `isValidLanguageId` or `isValidLanguageCode` is `false` (with a good error message for either case that includes the invalid request parameter) and then use the `int? validLanguageId` in the query below (replacing the current `languageId` variable).

elisabeth-shakespeare: nifty. thanks!

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -1,10 +1,12 @@
...
+﻿using Aquifer.Common.Services.Caching;
+using Aquifer.Data;
+using Aquifer.Data.Entities;
 using FastEndpoints;
 using Microsoft.EntityFrameworkCore;
 
 namespace Aquifer.Public.API.Endpoints.Resources.Updates.List;
 
-public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, Response>
+public class Endpoint(AquiferDbContext dbContext, ICachingLanguageService _cachingLanguageService) : Endpoint<Request, Response>
```
jwinston-bn, line 9: Inconsistent naming with one thing having an underscore and the other not. I don't have a preference either way in these primary constructors, but we should at least be consistent per file.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,69 @@ public override void Configure()
...
-        Response.Items = await dbContext.ResourceContentVersions
-            .Where(x => (!req.LanguageId.HasValue || x.ResourceContent.LanguageId == req.LanguageId) &&
-                x.IsPublished &&
-                x.Updated >= req.Timestamp)
+        var query = await GetQuery(req, ct);
+        var totalCount = await GetTotalResourceCountAsync(req, query, ct);
+
+        if (totalCount == 0)
+        {
+            await SendOkAsync(new Response{Offset = req.Offset}, ct);
```
jwinston-bn, line 32:
```suggestion
            await SendOkAsync(new Response { Offset = req.Offset }, ct);
```

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,69 @@ public override void Configure()
...
+        {
+            ThrowError($"Offset of {req.Offset} equals or exceeds total of {totalCount}");
+        }
+
+        return totalCount;
+    }
+
+    private async Task<IQueryable<ResourceContentVersionEntity>> GetQuery(Request req, CancellationToken ct)
+    {
+        var (_, _, validLanguageId) = await _cachingLanguageService.ValidateLanguageIdOrCodeAsync(req.LanguageId, req.LanguageCode, shouldRequireInput: false, ct);
```
jwinston-bn, line 77: This exceeds 140 line length limit

jwinston-bn: Also, don't think we need `shouldRequireInput:` here.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Request.cs
```diff
@@ -1,14 +1,39 @@
...
+
+    /// <summary>
+    ///     Optional ISO 639-3 code that is 3 characters in length (e.g. "eng").
+    /// </summary>
+    public string? LanguageCode { get; init; }
+
+    /// <summary>
+    ///     The number of results to return. Max is 100.
+    /// </summary>
+    [DefaultValue(10)]
```
jwinston-bn, line 25: Default value needs to be updated

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,73 @@ public override void Configure()
...
+        {
+            ThrowError($"Offset of {req.Offset} equals or exceeds total of {totalCount}");
+        }
+
+        return totalCount;
+    }
+
+    private async Task<IQueryable<ResourceContentVersionEntity>> GetQuery(Request req, CancellationToken ct)
+    {
+        var (_, _, validLanguageId) =
```
NateMerritt, line 80: This will ignore invalid language IDs like they weren't passed at all.  Do we want to return an error instead?  IMO as a consumer I would expect a bad request rather than data not being what I expected.

See https://github.com/BiblioNexusStudio/aquifer-server/blob/48af2303af99152adc2bbfa21676583a59f814d1/src/Aquifer.Public.API/Endpoints/Resources/Collections/Get/Endpoint.cs#L39-L46 for an example of throwing errors (though that route allows multiple language IDs or Codes so it will be simpler here).

elisabeth-shakespeare: Not sure what you mean by "weren't passed at all".  I had thought that we needed this to be open to the possibility of one or both of them being null.

I've tested putting in an invalid language code, and it just returns nothing. In that case it probably would be good for consumer and end user to know that their language code is invalid if they were to pass in something like 'foo', and the validator makes sure that the languageId is between 1 and 32 if it were passed in.

NateMerritt: > I've tested putting in an invalid language code, and it just returns nothing. In that case it probably would be good for consumer and end user to know that their language code is invalid if they were to pass in something like 'foo', and the validator makes sure that the languageId is between 1 and 32 if it were passed in.

The validator doesn't handle all scenarios.  For example, a `LanguageCode` of `TST` will pass validation but will result in no results being returned here which is unexpected IMO.  You could instead do:
```
 if (isInvalidLanguageCode) // from validation method return value
 { 
     ThrowError(r => r.LanguageCode, $"Invalid 'Language Code': {req.LanguageCode}"); 
 } 
```
Likewise, not all language IDs between 1 and 32 are valid.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Request.cs
```diff
@@ -1,14 +1,39 @@
...
-    public int? LanguageId { get; set; }
+    public int? LanguageId { get; init; }
+
+    /// <summary>
+    ///     Optional ISO 639-3 code that is 3 characters in length (e.g. "eng").
+    /// </summary>
+    public string? LanguageCode { get; init; }
+
+    /// <summary>
+    ///     The number of results to return. Max is 100.
```
jwinston-bn, line 23: This is still wrong, should be 1000.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Validator.cs
```diff
@@ -9,5 +9,13 @@ public Validator()
     {
         RuleFor(x => x.Timestamp).NotEmpty().GreaterThanOrEqualTo(DateTime.UtcNow.AddDays(-90).Date);
         RuleFor(x => x.LanguageId).InclusiveBetween(1, 32);
+        RuleFor(x => x.Offset).GreaterThanOrEqualTo(0);
+        RuleFor(x => x.Limit).InclusiveBetween(0, 1000);
```
jwinston-bn, line 13: I assume Limit 0 would just return no results, which I don't think makes sense. Probably should make this at least 1 to 1000.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
NateMerritt: > Sending language is optional for this endpoint, one or the other isn't required. Is that what you were getting at Nate?

Sorry for the confusion.  I mean the scenario where one of language code or language ID is passed but that code or ID is not in the DB.  In that case the current PR will ignore the invalid ID or code instead of erroring.

jwinston-bn: I agree that it should return a relevant error message in that case.

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,83 @@ public override void Configure()
...
+    }
+
+    private async Task<IQueryable<ResourceContentVersionEntity>> GetQuery(Request req, CancellationToken ct)
+    {
+        var (isValidLanguageId, isValidLanguageCode, validLanguageId) =
+            await cachingLanguageService.ValidateLanguageIdOrCodeAsync(req.LanguageId, req.LanguageCode, false, ct);
+
+        if (!isValidLanguageCode)
+        {
+            ThrowError($"Invalid LanguageCode:  '{req.LanguageCode}'");
```
jwinston-bn, line 85:
```suggestion
            ThrowError(x => x.LanguageCode, $"Invalid LanguageCode:  '{req.LanguageCode}'");
```

jwinston-bn: I'd remove the single quotes too

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -16,23 +18,83 @@ public override void Configure()
...
+        }
+
+        if (!isValidLanguageId)
+        {
+            ThrowError($"Invalid LanguageId: '{req.LanguageId}'");
```
jwinston-bn, line 90:
```suggestion
            ThrowError(x => x.LanguageId, $"Invalid LanguageId: {req.LanguageId}");
```

src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
elisabeth-shakespeare: It spits out an error now (just needs a tweak),  so make sure you pull the latest if you're testing.

elisabeth-shakespeare: > The validator doesn't handle all scenarios. For example, a LanguageCode of TST will pass validation but will result in no results being returned here which is unexpected IMO. You could instead do:

> if (isInvalidLanguageCode) // from validation method return value
 { 
     ThrowError(r => r.LanguageCode, $"Invalid 'Language Code': {req.LanguageCode}"); 
 } 
 
I can't find `isInvalidLanguageCode`  being returned in the validator, or anywhere in the `CachingLanguageService` file, but I have used `isValidLanguageCode / Id` for the error messages

NateMerritt: Sorry, I meant `isValid...` not `isInvalid`.

See [my original comment](https://github.com/BiblioNexusStudio/aquifer-server/pull/502#discussion_r1815544627) for getting back an `isValidLanguageCode` and `isValidLanguageId` result (instead of using the discard variable `_`):
```
var (isValidLanguageId, isValidLanguageCode, validLanguageId) = await ValidateLanguageIdOrCodeAsync(req.LanguageId, req.LanguageCode, shouldRequireInput: false, ct);
```

NateMerritt (approved):

PR #504, opened 2024-10-25
Bib 767 update status naming convention

kaseywright:


src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCounts.cs
```diff
@@ -25,20 +25,21 @@ public class ProjectResourceStatusCounts
         ResourceContentStatus.TranslationNotApplicable
     ];
 
-    internal static readonly List<ResourceContentStatus> NotStartedStatuses =
+    internal static readonly List<ResourceContentStatus> AwaitingAiDraftStatuses =
```
MrGrinst, line 28: I'm almost wondering if these should still be called `NotStartedStatuses` and we revert those renames on the frontend as well. The main reason being `AiDraftComplete` and `New` (i.e. Aquiferize not started) aren't really "awaiting AI draft". Thinking about the commonality between the three statuses it feels like they're all considered Not Started from the project's perspective. Thoughts?

src/Aquifer.API/Endpoints/Resources/Content/AssignEditor/Endpoint.cs
```diff
@@ -142,22 +142,21 @@ private static void SetDraftVersionStatus(bool isTakingBackFromReviewPending,
         UserEntity userToAssign,
         ResourceContentVersionEntity draftVersion)
     {
-        if (userToAssign.Role is not UserRole.Manager || !Constants.ManagerReviewStatuses.Contains(originalStatus))
+        if (userToAssign.Role is not UserRole.Manager && !Constants.CompanyReviewStatuses.Contains(originalStatus))
```
MrGrinst, line 145: Was this `||` to `&&` switch intentional?

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
```diff
@@ -23,11 +23,11 @@ public override async Task HandleAsync(CancellationToken ct)
         List<ResourceContentStatus> statuses =
         [
             ResourceContentStatus.New,
-            ResourceContentStatus.AquiferizeInProgress,
-            ResourceContentStatus.TranslationNotStarted,
-            ResourceContentStatus.TranslationInProgress,
-            ResourceContentStatus.AquiferizeManagerReview,
-            ResourceContentStatus.TranslationManagerReview
+            ResourceContentStatus.AquiferizeEditorReview,
+            ResourceContentStatus.TranslationAwaitingAiDraft,
```
MrGrinst, line 27: I think this should be `TranslationAiDraftComplete` since the manager won't have any awaiting draft items assigned to them.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToSelf/List/Endpoint.cs
```diff
@@ -40,7 +40,7 @@ public override async Task HandleAsync(CancellationToken ct)
     {
         var user = await userService.GetUserFromJwtAsync(ct);
         var queryResults = await dbContext.Database
-            .SqlQueryRaw<SqlQueryResult>(Query, user.Id, (int)ResourceContentStatus.TranslationNotStarted).ToListAsync(ct);
+            .SqlQueryRaw<SqlQueryResult>(Query, user.Id, (int)ResourceContentStatus.TranslationAwaitingAiDraft).ToListAsync(ct);
```
MrGrinst, line 43: `TranslationAiDraftComplete` here too

src/Aquifer.API/Endpoints/Resources/Content/NotApplicable/Update/Endpoint.cs
```diff
@@ -20,9 +20,9 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
         var user = await userService.GetUserFromJwtAsync(ct);
         List<ResourceContentStatus> allowedStatuses =
         [
-            ResourceContentStatus.TranslationNotStarted,
-            ResourceContentStatus.TranslationInProgress,
-            ResourceContentStatus.TranslationManagerReview
+            ResourceContentStatus.TranslationAwaitingAiDraft,
```
MrGrinst, line 23: `TranslationAiDraftComplete` here too

src/Aquifer.API/Endpoints/Resources/Content/ToAssign/List/Endpoint.cs
```diff
@@ -30,7 +30,7 @@ FROM ResourceContentVersions AS RCV
                               INNER JOIN ProjectResourceContents PRC ON PRC.ResourceContentId = RC.Id
                               INNER JOIN Projects P ON P.Id = PRC.ProjectId
                      WHERE RCV.IsDraft = 1 AND RCV.AssignedUserId = {user.Id}
-                       AND RC.Status IN ({(int)ResourceContentStatus.New}, {(int)ResourceContentStatus.TranslationNotStarted})
+                       AND RC.Status IN ({(int)ResourceContentStatus.New}, {(int)ResourceContentStatus.TranslationAwaitingAiDraft})
```
MrGrinst, line 33: `TranslationAiDraftComplete` here too

src/Aquifer.API/Endpoints/Resources/Untranslated/List/Endpoint.cs
```diff
@@ -43,7 +43,7 @@ FROM ResourceContents rc2
                              WHERE rc2.ResourceId = rc.ResourceId
                              AND rc2.MediaType = {(int)ResourceContentMediaType.Text}
                              AND rc2.LanguageId = @LanguageId
-                             AND (rc2.Status != {(int)ResourceContentStatus.TranslationNotStarted}
+                             AND (rc2.Status != {(int)ResourceContentStatus.TranslationAwaitingAiDraft}
```
MrGrinst, line 46: Probably want to check for both awaiting and `TranslationAiDraftComplete`

src/Aquifer.API/Endpoints/Resources/Content/AssignEditor/Endpoint.cs
kaseywright: Yes, otherwise, when a item in `CompanyReview` status is kicked back to an editor, it would revert the status to `EditorReview` as well which is not what should happen according to this ticket

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCounts.cs
jwinston-bn: I think that's fair. The next step is EditorReview, so we can be saying that editor review hasn't started, which from a project's point of view, is the relevant bit.

kaseywright: I can see that for the purpose of project count reporting. there were a lot of changes that had to be merged from master and a few things like this were just easier to keep as-is to make sure nothing broke.

src/Aquifer.API/Endpoints/Projects/AssignedToSelf/List/Endpoint.cs
```diff
@@ -40,13 +40,21 @@ private async Task<List<Response>> GetProjectsAsync(CancellationToken ct)
                         : null,
                 Counts = new ProjectResourceStatusCounts
                 {
-                    NotStarted = x.ProjectResourceContents.Count(prc => ProjectResourceStatusCounts.NotStartedStatuses.Contains(prc.ResourceContent.Status)),
-                    InProgress = x.ProjectResourceContents.Count(prc => ProjectResourceStatusCounts.InProgressStatuses.Contains(prc.ResourceContent.Status)),
-                    InManagerReview =
-                        x.ProjectResourceContents.Count(prc => ProjectResourceStatusCounts.InManagerReviewStatuses.Contains(prc.ResourceContent.Status)),
+                    AwaitingAiDraft =
```
MrGrinst, line 43: Could we also rename this and the other occurrences of `AwaitingAiDraft` for project resource counts?

src/Aquifer.API/Endpoints/Resources/Untranslated/List/Endpoint.cs
MrGrinst: @kaseywright bump on this one

src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -109,20 +109,6 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
             ThrowError("Data integrity issue, no resource content version found.");
         }
 
-        if (relevantContentVersion.IsDraft &&
```
kaseywright, line 114: Should this be restored?

MrGrinst: I think yes we do actually want that back since it updates from PendingReview to CompanyReview. Probably need to update that other endpoint to allow it as well.

kaseywright: Which other endpoint?

src/Aquifer.API/Endpoints/Resources/Content/AssignEditor/Endpoint.cs
```diff
@@ -142,22 +142,21 @@ private static void SetDraftVersionStatus(bool isTakingBackFromReviewPending,
...
         {
             var newStatus = isTakingBackFromReviewPending
```
MrGrinst, line 147: Ah here's the part I was thinking of. I think this logic may still be tweaked to get things working. My brain is a bit fuzzy on all the specifics but I think maybe the else if below should be above this conditional, then it should be:
- if isTakingBackFromReviewPending assign to CompanyReview
- else if status is ai draft complete or new, change to EditorReview
- else do nothing

Does that seem right?

MrGrinst: Maybe also need a conditional in there for when a publisher is sending back to a manager?

tl;dr: the user assignment always happens, but the status only changes when being sent back from publisher -> manager, being taken from pending review to manager, or first being assigned to a user.

kaseywright: I think so. I'll refactor it

kaseywright: Github refresh fast enough didn't see your last comment. I'll make sure this works as expected

jwinston-bn (approved):

PR #508, opened 2024-10-28
validate ResourceCollectionCode and throw error 

elisabeth-shakespeare:


src/Aquifer.Public.API/Endpoints/Resources/Updates/List/Endpoint.cs
```diff
@@ -97,4 +98,13 @@ private async Task<IQueryable<ResourceContentVersionEntity>> GetQuery(Request re
                         (req.ResourceCollectionCode == null ||
                          x.ResourceContent.Resource.ParentResource.Code.ToLower() == req.ResourceCollectionCode.ToLower()));
     }
+
+    private void ValidateCollectionCode(Request req)
+    {
+        var exists = dbContext.ParentResources.Any(x => x.Code == req.ResourceCollectionCode);
```
jwinston-bn, line 104: Resource collection code is nullable. I would think we would verify it's not null before calling the database.

elisabeth-shakespeare: good idea

jwinston-bn (approved):

PR #510, opened 2024-10-29
Adding Status Endpoint To Resource Content

carlosmorales125:


src/Aquifer.API/Endpoints/Resources/Content/Status/Get/Validator.cs
```diff
@@ -0,0 +1,18 @@
+using FastEndpoints;
+using FluentValidation;
+
+namespace Aquifer.API.Endpoints.Resources.Content.Status.Get;
+
+public class Validator : Validator<Request>
+{
+    public Validator()
```
jwinston-bn, line 8: I don't think you can actually get into this without providing the value because it's in the path. So I think this file can be removed.

jwinston-bn (approved):

PR #515, opened 2024-10-29
Translate project and resource content via queues and durable functions.

NateMerritt:
This PR accomplishes multiple goals, all of which were required in order to move translations to an async subscriber:
1. Move Tiptap conversion to C#.  [aquifer-tiptap](https://github.com/biblioNexusStudio/aquifer-tiptap/) allows creation of a `TiptapUtilities.js` file which could be copy/pasted and run via Microsoft ClearScript's V8 Engine support (as is currently done in AI.Worker).  Instead of copy/pasting, though, this PR instead pulls in that source code as a node package and builds the `TiptapUtilities.js` file on every build.
2. Handle multi-step FIA Tiptap content conversion (to/from JSON/HTML).
3. Move the OpenAI translation and prompts fully to the back-end and modify it to pull down the entire chat instead of streaming, now using the C# client in the [OpenAI](https://github.com/openai/openai-dotnet) package (which comes with some nice built-in error retry handling).  Also, translate paragraphs in parallel to improve translation speeds.
4. Handle multi-step FIA Tiptap content translation.
5. Create a C# utility for getting the word count of Tiptap HTML content now that `TiptapUtilities.js` no longer supports this functionality.
6. Minify HTML before translating it so that we are less likely to hit translation character limits.  This logic already exists in AI.Worker but it has been updated to be in a utility method with tests.
7. Use Azure durable functions for fan out -> fan in behavior in order to translate all resources for a project and then perform final project related updates.
8. Update the translation flow to create an initial (original language) snapshot before translating and a second translated content snapshot after translating.
9. Update the translation flow to do appropriate resource content version status and assignment changes based upon if the translation began from project start, a publisher using the translation dropdown, or a community reviewer clicking the translate button.

Breaking change:
* The `ResourceContentVersionMachineContentTranslation` `UserId` column is now nullable because a back-end process is creating the translations.  We will probably remove this column altogether in a future change.  Note that the InternalApi now can return `int?` for the `UserId` as well which is a breaking change but which shouldn't have an impact outside of updating translation ratings.

Example translations:
* One resource by Publisher via translation dropdown: https://qa.admin.aquifer.bible/resources/366831
* One resource by Community Reviewer via Translate button: https://qa.admin.aquifer.bible/resources/366301
* Entire project: https://qa.admin.aquifer.bible/projects/1203

Related PRs to use the new `TiptapConverter` from this PR:
* https://github.com/BiblioNexusStudio/ai-worker/pull/1
* https://github.com/BiblioNexusStudio/content-loader/pull/33


src/Aquifer.Common/Aquifer.Common.csproj
```diff
@@ -10,15 +10,22 @@
...
         <PackageReference Include="SendGrid.Extensions.DependencyInjection" Version="1.0.1"/>
     </ItemGroup>
 
+    <ItemGroup>
+        <EmbeddedResource Include="..\Aquifer.TipTap\dist\TiptapUtilities.js" Link="Tiptap\TiptapUtilities.js"/>
+    </ItemGroup>
+
     <ItemGroup>
         <ProjectReference Include="..\Aquifer.Data\Aquifer.Data.csproj"/>
+        <ProjectReference Include="..\Aquifer.TipTap\Aquifer.TipTap.esproj"/>
```
NateMerritt, line 28: This forces `Aquifer.Tiptap` to build first.

src/Aquifer.Common/Aquifer.Common.csproj
```diff
@@ -10,15 +10,22 @@
...
+        <PackageReference Include="Microsoft.ClearScript.V8.Native.win-x64" Version="7.4.5"/>
         <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.2"/>
         <PackageReference Include="Newtonsoft.Json" Version="13.0.1"/>
         <PackageReference Include="ReverseMarkdown" Version="4.6.0"/>
         <PackageReference Include="SendGrid" Version="9.29.3"/>
         <PackageReference Include="SendGrid.Extensions.DependencyInjection" Version="1.0.1"/>
     </ItemGroup>
 
+    <ItemGroup>
+        <EmbeddedResource Include="..\Aquifer.TipTap\dist\TiptapUtilities.js" Link="Tiptap\TiptapUtilities.js"/>
```
NateMerritt, line 23: Using a Link allows us to embed the output of the Aquifer.Tiptap build.

src/Aquifer.TipTap/Aquifer.TipTap.esproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.VisualStudio.JavaScript.Sdk/1.0.1184077">
```
NateMerritt, line 1: Note to self: It should be `Aquifer.Tiptap`.

src/Aquifer.AI/OpenAiTranslationService.cs
```diff
@@ -0,0 +1,194 @@
...
+    /// </summary>
+    public Task<string> TranslateHtmlAsync(
+        string html,
+        (string Iso6393Code, string EnglishName) destinationLanguage,
+        CancellationToken cancellationToken);
+}
+
+public class OpenAiOptions
+{
+    public required string Model { get; init; }
```
NateMerritt, line 32: The OpenAI client requires specifying the AI model in the constructor but we weren't previously passing the model in our homegrown client.  I'm assuming the model is `gpt-4o`.

src/Aquifer.AI/OpenAiTranslationService.cs
```diff
@@ -0,0 +1,194 @@
...
+
+    public OpenAiTranslationService(
+        OpenAiTranslationOptions openAiTranslationOptions,
+        OpenAiOptions openAiOptions,
+        IAzureKeyVaultClient keyVaultClient)
+    {
+        _options = openAiTranslationOptions;
+
+        // TODO Inject a ChatClient instead of building one here. This will require changing how we fetch key vault secrets.
+        const string openAiApiKeySecretName = "OpenAiApiKey";
```
NateMerritt, line 65: Currently the KeyVault secret names are all in a constants file only accessible by Aquifer.API.  I want to look into doing KeyVault secrets differently, anyway, so I didn't bother to refactor those constants to a shared location.

src/Aquifer.API/Endpoints/Projects/Start/Endpoint.cs
```diff
@@ -1,8 +1,10 @@
 using Aquifer.API.Common;
 using Aquifer.API.Services;
-using Aquifer.Common.Services;
+using Aquifer.Common.Jobs.Messages;
+using Aquifer.Common.Jobs.Publishers;
 using Aquifer.Data;
 using Aquifer.Data.Entities;
+using Aquifer.Data.Services;
```
NateMerritt, line 7: I moved the `ResourceHistoryService` into the `Aquifer.Data` project so that I could access it from `Aquifer.Jobs`.

src/Aquifer.Common/Jobs/Publishers/TranslationPublisher.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Aquifer.Common.Jobs.Messages;
+
+namespace Aquifer.Common.Jobs.Publishers;
+
+public interface ITranslationPublisher
```
NateMerritt, line 5: I decided to do a naming change and make all services that publish messages to queues have `Publisher` at the end of the class name and all methods that do the publishing to have `Publish` at the beginning of the method name.  I'm planning to update existing publishers in a separate PR after this one merges (along with some other naming and directory changes).

src/Aquifer.Data/Migrations/20241101195532_Make_ResourceContentVersionMachineTranslation_UserId_Nullable.cs
```diff
@@ -0,0 +1,59 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
NateMerritt, line 1: I already ran against the Dev DB.

src/Aquifer.Jobs/Program.cs
```diff
@@ -12,12 +14,15 @@
...
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 var host = new HostBuilder()
     .ConfigureFunctionsWorkerDefaults()
     .ConfigureServices((context, services) =>
     {
         services.AddOptions<ConfigurationOptions>().Bind(context.Configuration);
+        services.AddSingleton(cfg => cfg.GetService<IOptions<ConfigurationOptions>>()!.Value.OpenAi);
```
NateMerritt, line 24: I moved AI translation into its own project and thus I needed to support options for a class that lives in a separate project from the running app.  To do this I settled on a simple method of:
1. Specifying the class in the dependency project.
2. Loading the options as part of the standard `ConfigurationOptions` object.
3. _Also_ wiring up the specific options as a separate Singleton so that they can be accessed by only the dependency project's class (which doesn't know about `ConfigurationOptions`).

I could also have used `IOptions<T>` but chose this strategy because it is harder to bind to `IOptions<T>` here without doing a rebind or a separate bind for that configuration section (by name) and I don't see a lot of value in the `IOptions<T>` wrapper.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,383 @@
...
+                _logger.LogError(
+                    orchestrationException,
+                    "Error translating resource content for Project ID {ProjectId}. A poison message will be published to \"{PoisonQueueName}\" to enable manual retry. Manual dev intervention is required.",
+                    dto.ProjectId,
+                    dto.PoisonQueueName);
+
+                var translateProjectResourcesPoisonQueueClient = await _queueClientFactory.GetQueueClientAsync(
+                    dto.PoisonQueueName,
+                    ct);
+                await translateProjectResourcesPoisonQueueClient.SendMessageAsync(dto.OriginalQueueMessageText, ct);
```
NateMerritt, line 153: I like this pattern of publishing a message back to the poison queue of the original subscriber function if the orchestration fails.  The original subscriber has no way of tracking when the orchestration completes; it simply triggers the process and completes successfully and then all of the translations run in the background.

src/Aquifer.JsEngine/Aquifer.JsEngine.csproj
```diff
@@ -0,0 +1,22 @@
...
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="JavaScriptEngineSwitcher.V8" Version="3.24.2" />
+        <PackageReference Include="Microsoft.ClearScript.Complete" Version="7.4.5" />
```
NateMerritt, line 11: This includes Windows, Linux, and MacOS nuget packages (as well as multiple CPU variants) so that all dev workstations and the build server (Linux) can properly load JS files.

src/Aquifer.JsEngine/Aquifer.JsEngine.csproj
```diff
@@ -0,0 +1,22 @@
...
+    </ItemGroup>
+
+    <ItemGroup>
+        <EmbeddedResource Include="..\Aquifer.Tiptap\dist\TiptapUtilities.js" Link="Tiptap\TiptapUtilities.js" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="..\Aquifer.Tiptap\Aquifer.Tiptap.esproj" />
```
NateMerritt, line 19: This project reference required me to turn off a build warning but the reference is necessary in order to ensure that `TiptapUtilities.js` is built and available before this project builds.

src/Aquifer.JsEngine/Tiptap/TiptapUtilities.cs
```diff
@@ -0,0 +1,41 @@
+﻿using JavaScriptEngineSwitcher.V8;
+
+namespace Aquifer.JsEngine.Tiptap;
+
+/// <summary>
+/// This class is purely a C# wrapper around TiptapUtilities.js.
+/// See also Aquifer.Tiptap where that file is generated.
+/// </summary>
+public sealed class TiptapUtilities : IDisposable
```
NateMerritt, line 9: I'd prefer that any classes like this have a public API surface area that only exposes the JS methods under the hood.

src/Aquifer.Tiptap/Aquifer.Tiptap.esproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.VisualStudio.JavaScript.Sdk/1.0.1184077">
+  <PropertyGroup>
+    <BuildCommand>npm install &amp;&amp; npx esbuild node_modules/aquifer-tiptap/src/TiptapUtilities.ts --bundle --outfile=dist/TiptapUtilities.js --platform=node --target=node12</BuildCommand>
```
NateMerritt, line 3: The build command is copied from [aquifer-tiptap](https://github.com/BiblioNexusStudio/aquifer-tiptap/blob/e8971fe2a26b4f130b8f9d2cdfdafe8b17fd532c/package.json#L7).  I did have some issues when downgrading the "node" package in which case adding `node update aquifer-tiptap` in here fixed it but I didn't want to add that overhead to every build.  It's probably fine without it if we only increment the package.

src/Aquifer.Tiptap/package.json
```diff
@@ -0,0 +1,22 @@
...
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "build": "tsc --build",
+        "clean": "tsc --build --clean"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "description": "",
+    "dependencies": {
+        "aquifer-tiptap": "https://github.com/BiblioNexusStudio/aquifer-tiptap#1.0.9"
```
NateMerritt, line 15: We'll need to bump this whenever we create a new tag for aquifer-tiptap.  I added a note in the readme about this.

src/Aquifer.Common/Jobs/Publishers/TranslationPublisher.cs
NateMerritt: @MrGrinst, you're suggesting `ITranslationMessagePublisher`?  I don't object though I'd also want to change the subscribers to `MessageSubscriber` for consistency.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,383 @@
...
+using Azure.Storage.Queues.Models;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Aquifer.Jobs.Subscribers;
+
+public sealed class TranslationSubscriber(
```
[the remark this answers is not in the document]
NateMerritt: It could go either way.  There is shared translation code for the invidual resource so it's either one file or three (two for the subscribers and one for the shared code) or possibly four if we want to move the orchestration to another file, too.

As it stands, I chose one because it encapsulates all of the translation logic into one place.  FWIW, I also put all notification queue subscribers in one class for a similar reason.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,383 @@
...
+            message.ProjectId);
+    }
+
+    /// <summary>
+    /// Durable orchestration function.
+    /// This function is single-threaded and thus should not do any DB operations or any substantial processing.
+    /// It should only orchestrate activities.
+    /// </summary>
+    [Function(nameof(OrchestrateProjectResourcesTranslation))]
+    public async Task OrchestrateProjectResourcesTranslation(
```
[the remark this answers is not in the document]
NateMerritt: 1. We only need one queue for project translations instead of three (project start, resource translate, project post-translate).
2. Each of the durable function activities builds in its own retry logic and error handling without me needing to do any coding.
3. And the biggest reason: fan out -> fan in.  The project orchestration triggers the translation of each resource in its own function and then the final project post translation function is run _only once_ after all resources have been translated, all with built in retry and error handling.  We had a proposed solution to try to solve that same concern with just queues but it got tricky because we would have needed the post-project translate queue subscriber to run after every resource translation to see if all resources had been translated in order to do the final work for the project.  That solution also came with potential race conditions and the danger of multiple queue runs happening for the same logic.  At a minimum it required an additional DB call after each resource was translated to see if all resources were in the correct status compared to this durable function fan out -> fan in strategy.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
NateMerritt: Good question.  For a single queue subscriber that fails, yes, the poison queue logic happens automatically.

However, the translate project resources queue subscriber can only kick off the durable functions orchestration.  It doesn't wait for the orchestration to finish (due to how durable functions work).  Thus, so long as it successfully kicks off the orchestration the subscriber will succeed and won't publish to the poison queue, even if the orchestration eventually fails.

Therefore, I manually published to the poison queue if the orchestration fails so we can easily retry the project translation.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,383 @@
...
+    /// <summary>
+    /// Note: Callers are responsible for performing resource assignments and creating original resource content snapshots.
+    /// This method will perform the actual translation, create machine translations,
+    /// update the existing draft resource version with translated content, create a new snapshot for the translation,
+    /// and change the resource content status.
+    /// </summary>
+    private async Task<ResourceContentVersionEntity> TranslateResourceCoreAsync(
+        int resourceContentId,
+        int startedByUserId,
+        TranslationOrigin translationOrigin,
```
NateMerritt, line 238: Move community reviewer post logic here.

src/Aquifer.AI/OpenAiTranslationService.cs
```diff
@@ -0,0 +1,177 @@
...
+
+public interface ITranslationService
+{
+    /// <summary>
+    /// Translates plain text.
+    /// Note: It's up to the caller to guarantee that the text length is not too long.
+    /// </summary>
+    public Task<string> TranslateTextAsync(
+        string text,
+        (string Iso6393Code, string EnglishName) destinationLanguage,
```
jwinston-bn, line 17: Is this ISO code being used anywhere? I don't think we'd ever pass it to OpenAI.

NateMerritt: It's being used to look up the correct prompt for HTML.  I decided to just pass the same language information to both methods (text and HTML) for consistency even though the ISO code isn't needed for straight text translation.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,379 @@
...
+        FunctionContext activityContext)
+    {
+        _logger.LogInformation(
+            "Beginning post-translation project updates for Project ID {ProjectId}.",
+            dto.ProjectId);
+
+        var project = await _dbContext.Projects
+            .SingleAsync(p => p.Id == dto.ProjectId);
+
+        var companyLeadUserId = project.CompanyLeadUserId
```
jwinston-bn, line 180: Looks like this value is never used.

NateMerritt: Fix: https://github.com/BiblioNexusStudio/aquifer-server/pull/515/commits/e0aedde6f1a5abb26579ff5a1f4687e31f5d47c7

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,379 @@
...
+        // if a Community Reviewer requested the translation then the status needs to be updated (again) now that translation has completed
+        if (translationOrigin == TranslationOrigin.CommunityReviewer)
+        {
+            resourceContentVersion.ResourceContent.Status = ResourceContentStatus.TranslationEditorReview;
+            await _resourceHistoryService.AddStatusHistoryAsync(
+                resourceContentVersion,
+                ResourceContentStatus.TranslationEditorReview,
+                changedByUserId: startedByUserId,
+                ct);
+        }
```
jwinston-bn, line 359: I don't think there's any value in running this twice. I would just have this be an if/else.

NateMerritt: I thought about that but I think there's value in always having the status move from `TranslationAwaitingAiDraft` to `TranslationAiDraftComplete` because that shows that this tooling did what it was supposed to do and always does it consistently across the different translation origins.  Any objection to leaving it like this?

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,379 @@
...
+            ?? throw new InvalidOperationException(
+                $"Aborting translation for Resource Content ID {resourceContentId} because it has no Resource Content Version with Text Media Type in the Draft status.");
+
+        if (resourceContentVersion.ResourceContentVersionSnapshots.Count == 0)
+        {
+            throw new InvalidOperationException(
+                $"Aborting translation for Resource Content ID {resourceContentId} because there are no pre-existing Snapshots for Resource Content Version ID {resourceContentVersion.Id}.");
+        }
+
+        if (resourceContentVersion.ResourceContent.Status != ResourceContentStatus.TranslationAwaitingAiDraft)
```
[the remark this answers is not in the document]
NateMerritt: The exception above can only happen if we're in an invalid state because anything that queues a translation message should have already created a snapshot for all related resource content versions.

Being that such a state should never happen, it seems right to validate the data before attempting any action on it, including gracefully skipping translation.

src/Aquifer.Jobs/Subscribers/TranslationSubscriber.cs
```diff
@@ -0,0 +1,379 @@
...
+    /// </summary>
+    [Function(nameof(TranslateProjectResources))]
+    public async Task TranslateProjectResources(
+        [QueueTrigger(Queues.TranslateProjectResources)] QueueMessage queueMessage,
+        [DurableClient] DurableTaskClient durableTaskClient,
+        CancellationToken ct)
+    {
+        var message = queueMessage.Deserialize<TranslateProjectResourcesMessage, TranslationSubscriber>(_logger);
+
+        var projectResourceContentIds = await _dbContext.Projects
```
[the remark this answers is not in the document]
NateMerritt: Good call.  Updated: https://github.com/BiblioNexusStudio/aquifer-server/pull/515/commits/595e7f2395fc4e2608f15c55ab2ed6609b559ee3

jwinston-bn (approved):

PR #518, opened 2024-10-30
set public API docs page title correctly

MrGrinst:


src/Aquifer.Public.API/Program.cs
```diff
@@ -48,7 +48,11 @@
         config.Endpoints.Configurator = ep => ep.AllowAnonymous();
     })
     .UseOpenApi()
-    .UseReDoc(options => options.Path = "/docs");
+    .UseReDoc(options =>
+    {
+        options.Path = "/docs";
+        options.DocumentTitle = "Aquifer API Docs";
```
jwinston-bn, line 54: Is this removing the (1.0.0) piece? Would prefer "Documentation" over "Docs".

MrGrinst: It's setting the page title so it shows right in the browser tab bar:
image

MrGrinst: `Documentation` feels a bit long for the page title, we could also just drop it entirely and do `Aquifer API`. Thoughts?

jwinston-bn: I think it's pretty common for page titles to be that length or longer (I don't have a single tab open right now that isn't as long). If there's a quick and similar way to get rid of that "(1.0.0)" as part of this PR that'd be great.

MrGrinst: Addressed this

jwinston-bn (approved):

PR #521, opened 2024-10-31
Bib 779 update api endpoints for pre translate

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/AI/Translate/Endpoint.cs
elisabeth-shakespeare, line 1: Doesn't seem right to delete the file.
From what I've seen with fastendpoints, the only way to deprecate an endpoint is to use its versioning and set it to deprecate at a version, and then it's not set up for that version. That didn't seem right either.

src/Aquifer.API/Endpoints/Comments/Threads/Create/Endpoint.cs
```diff
@@ -34,10 +34,19 @@ private async Task<Response> CreateResourceContentVersionCommentThreadAsync(Requ
...
         EndpointHelpers.ThrowErrorIfNull<Request>(resourceContentVersion, x => x.TypeId, "No type found for given id.", 404);
 
+        List<ResourceContentStatus> notAllowedStatuses =
+            [ResourceContentStatus.TranslationAiDraftComplete, ResourceContentStatus.TranslationAwaitingAiDraft];
+
+        if (notAllowedStatuses.Contains(resourceContentVersion!.ResourceContent.Status))
+        {
+            ThrowError(x => x.TypeId, $"Resource for type {req.TypeId} is not in correct status");
+        }
+
```
elisabeth-shakespeare, line 49: I'd like to put this in a Helper Method. Would Common be the best place to put it?

src/Aquifer.API/Endpoints/Resources/BibleReferences/Create/Endpoint.cs
MrGrinst, line 1: Looks like there were only formatter changes in this file

src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Create/Endpoint.cs
```diff
@@ -17,6 +18,8 @@ public override void Configure()
 
     public override async Task HandleAsync(Request req, CancellationToken ct)
```
MrGrinst, line 19: As I understand it this endpoint is going away entirely so probably not worth making changes to it

elisabeth-shakespeare: Is Retranslate going away too?

src/Aquifer.API/Endpoints/Resources/BibleReferences/Create/Endpoint.cs
elisabeth-shakespeare: yeah, I kept running format project wide, and figured might as well push up cleaner stuff while I pushed up other little housekeeping items

jwinston-bn (approved):

PR #523, opened 2024-11-01
Bib 788 add appsettings to jobs

kaseywright:
This allows the setting of App Settings from `appSettings.json` and `appSettings.{env}.json` in the Jobs project.

### DIFF
--- a/src/Aquifer.Jobs/Aquifer.Jobs.csproj
+++ b/src/Aquifer.Jobs/Aquifer.Jobs.csproj
@@ -32,6 +32,13 @@
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
             <CopyToPublishDirectory>Never</CopyToPublishDirectory>
         </None>
+        <None Update="appsettings.json">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </None>
+        <None Update="appsettings.Development.json">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+            <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+        </None>
     </ItemGroup>
     <ItemGroup>
         <Using Include="System.Threading.ExecutionContext" Alias="ExecutionContext"/>
--- a/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
+++ b/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
@@ -9,7 +9,6 @@ public class ConfigurationOptions
     public required ConnectionStringOptions ConnectionStrings { get; init; }
     public required AnalyticsOptions Analytics { get; init; }
     public required EmailOptions Email { get; init; }
-    public required MarketingEmailOptions MarketingEmail { get; init; }
 }
 
 public class ConnectionStringOptions
@@ -33,6 +32,7 @@ public class AnalyticsOptions
 public class EmailOptions
 {
     public required string SubjectPrefix { get; init; }
+    public required MarketingEmailOptions Marketing { get; init; }
     public required TransactionalEmailOptions Transactional { get; init; }
 }
 
--- a/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
+++ b/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
@@ -33,6 +33,7 @@ public class AnalyticsOptions
 public class EmailOptions
 {
     public required string SubjectPrefix { get; init; }
+    public required TransactionalEmailOptions Transactional { get; init; }
 }
 
 public class MarketingEmailOptions
@@ -40,4 +41,20 @@ public class MarketingEmailOptions
     public required string Address { get; init; }
     public required string Name { get; init; }
     public required string ResourceLink { get; init; }
+}
+
+public class TransactionalEmailOptions
+{
+    public required SendResourceAssignmentNoticeOptions SendResourceAssignmentNotice { get; init; }
+    public required SubscriberNoticeOptions SubscriberNotice { get; init; }
+}
+
+public class SendResourceAssignmentNoticeOptions
+{
+    public required string TemplateId { get; init; }
+}
+
+public class SubscriberNoticeOptions
+{
+    public required string TemplateId { get; init; }
 }
--- a/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
+++ b/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
@@ -2,13 +2,14 @@
 
 public class ConfigurationOptions
 {
-    public bool IsDevelopment { get; init; }
     public required string AquiferAdminBaseUri { get; init; }
     public required string AquiferApiBaseUri { get; init; }
     public required string KeyVaultUri { get; init; }
     public required ConnectionStringOptions ConnectionStrings { get; init; }
     public required AnalyticsOptions Analytics { get; init; }
     public required EmailOptions Email { get; init; }
+    public required NotificationsOptions Notifications { get; init; }
+    public required MarketingEmailOptions MarketingEmail { get; init; }
 }
 
 public class ConnectionStringOptions
@@ -32,8 +33,6 @@ public class AnalyticsOptions
 public class EmailOptions
 {
     public required string SubjectPrefix { get; init; }
-    public required MarketingEmailOptions Marketing { get; init; }
-    public required TransactionalEmailOptions Transactional { get; init; }
 }
 
 public class MarketingEmailOptions
@@ -43,24 +42,9 @@ public class MarketingEmailOptions
     public required string ResourceLink { get; init; }
 }
 
-public class TransactionalEmailOptions
+public class NotificationsOptions
 {
-    public required SendResourceAssignmentNotificationOptions SendResourceAssignmentNotification { get; init; }
-    public required SendProjectStartedNotificationOptions SendProjectStartedNotification { get; init; }
-    public required SendResourceCommentCreatedNotificationOptions SendResourceCommentCreatedNotification { get; init; }
-}
-
-public class SendResourceAssignmentNotificationOptions
-{
-    public required string TemplateId { get; init; }
-}
-
-public class SendProjectStartedNotificationOptions
-{
-    public required string TemplateId { get; init; }
-}
-
-public class SendResourceCommentCreatedNotificationOptions
-{
-    public required string TemplateId { get; init; }
+    public required string SendResourceAssignmentNotificationTemplateId { get; init; }
+    public required string SendProjectStartedNotificationTemplateId { get; init; }
+    public required string SendResourceCommentCreatedNotificationTemplateId { get; init; }
 }
--- a/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
+++ b/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
@@ -45,16 +45,22 @@ public class MarketingEmailOptions
 
 public class TransactionalEmailOptions
 {
-    public required SendResourceAssignmentNoticeOptions SendResourceAssignmentNotice { get; init; }
-    public required SubscriberNoticeOptions SubscriberNotice { get; init; }
+    public required SendResourceAssignmentNotificationOptions SendResourceAssignmentNotification { get; init; }
+    public required SendProjectStartedNotificationOptions SendProjectStartedNotification { get; init; }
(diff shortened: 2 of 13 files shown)


src/Aquifer.Jobs/Program.cs
```diff
@@ -15,6 +15,15 @@
 
 var host = new HostBuilder()
     .ConfigureFunctionsWorkerDefaults()
+    .ConfigureAppConfiguration((context, config) => 
+    {
+        var envName = context.HostingEnvironment.EnvironmentName;
+        var appSettingsPath = Path.GetFullPath("../../appsettings.json");
```
kaseywright, line 21: This method of determining the location of the appSettings.json file feels terrible but it was the only method that worked. Other attempts using `config.AddJsonFile("appsettings.json", optional: false, reloadOnChange: true);` would look in the `bin/ouput/` directory resulting in a failed state.

src/Aquifer.Jobs/appsettings.example.json
```diff
@@ -0,0 +1,17 @@
...
+        "Marketing": {
+            "Name": {
+                "TemplateId": "sendgrid-id-here"
+            }
+        },
+        "Transactional": {
+            "SendResourceAssignmentNotice": {
+                "TemplateId": "sendgrid-id-here"
+            },
+            "SubscriberNotice": {
```
NateMerritt, line 12: We'll want two sections here, not one (each notification uses a unique template ID):
* `SendProjectStartedNotification`
* `SendResourceCommentCreatedNotification`

src/Aquifer.Jobs/Program.cs
jwinston-bn: Did you look at this at all? https://tharaka-madhusanka.medium.com/how-to-inject-appsettings-environment-json-in-azure-functions-4aea2f10e861

NateMerritt: This won't work after we deploy because the `appsettings.json` will only be in that location on your workstation and won't be included as part of the deployment package.

Try following the steps in [this SO answer](https://stackoverflow.com/a/78120434):
1. Use `config.SetBasePath(context.HostingEnvironment.ContentRootPath)`
2. Ensure the csproj settings are correct for each of the config files (which will copy the new ones into the output directory).

NateMerritt: > Did you look at this at all? https://tharaka-madhusanka.medium.com/how-to-inject-appsettings-environment-json-in-azure-functions-4aea2f10e861

@jwinston-bn, that example is for the old in-process model and thus is obsolete now.  Different code is need for the new isolated model which is what we are using (and is the future).  This bit me with durable functions until I learned the difference.

The SO post I shared above is from someone trying to convert from in-process to isolated and should work.

src/Aquifer.Jobs/Aquifer.Jobs.csproj
```diff
@@ -32,6 +32,13 @@
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
             <CopyToPublishDirectory>Never</CopyToPublishDirectory>
         </None>
+        <None Update="appsettings.json">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </None>
+        <None Update="appsettings.Development.json">
```
NateMerritt, line 38: If you remove your local `appsettings.Development.json` (copy it to a different folder outside the solution) will the project still build?  This file won't be available for CI/CD and we don't want it checked into source control.

NateMerritt: We might need to use a post-build step instead:
```
xcopy /y $(ProjectDir)appsettings.Development.json $(ProjectDir)$(OutDir)
```

src/Aquifer.Jobs/appsettings.json
```diff
@@ -0,0 +1,16 @@
+{
+    "Email": {
+        "Marketing": { },
+        "Transactional": {
+            "SendResourceAssignmentNotification": {
+                "TemplateId": "sendgrid-id-here"
```
NateMerritt, line 6: Will it work to specify the template IDs here and not override them in the environment specific settings?  There's no current need for environment specific overrides.

src/Aquifer.Jobs/Program.cs
```diff
@@ -15,6 +15,14 @@
 
 var host = new HostBuilder()
     .ConfigureFunctionsWorkerDefaults()
+    .ConfigureAppConfiguration((context, builder) => builder
+        .SetBasePath(context.HostingEnvironment.ContentRootPath)
+        .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
```
NateMerritt, line 20: Does adding the new `appsettings.json` stop the old `local.settings.json` from loading in?

src/Aquifer.Jobs/appsettings.example.json
```diff
@@ -0,0 +1,16 @@
+{
```
NateMerritt, line 1: If this PoC works (I'm assuming it does) then we should move to only using the `appsettings.json` paradigm.  Thus, everything in `local.settings.example.json` should move into this new `appsettings.example.json`, instead, as part of this PR.

src/Aquifer.Jobs/Program.cs
```diff
@@ -15,6 +15,14 @@
...
+        .AddJsonFile(
+            $"appsettings.{context.HostingEnvironment.EnvironmentName}.json", 
+            optional: true, 
+            reloadOnChange: false)
```
NateMerritt, line 24: We'll either want to call  `.AddEnvironmentVariables()` at the end of this chain in order to get environment variable overrides _or_ move the `ConfigureFunctionsWorkerDefaults()` method to the end of the chain which might result in that being solved correctly.

To test you can create environment variables locally that match QA, debug a function that pulls in the `ConfigurationOptions` via DI, and check the values.

The generally respected order should be:
1. appsettings.json
2. {environment}.appsettings.json
3. Environment variables

Ideally we use the configure defaults method at the end because that also wires up other overrides that we don't currently use but which would keep things consistent with the other API apps.

NateMerritt (changes requested):
Needed changes:
1. Ensure environment variable overrides work correctly.
2. Fully migrate all settings to `appsettings.json` (including the examples).
3. Ensure that `appsettings.Development.json` is not required by the solution to build.

src/Aquifer.Jobs/Aquifer.Jobs.csproj
kaseywright: It builds when I remove/relocate it.

src/Aquifer.Jobs/appsettings.json
kaseywright: It will, we can move these if we want to. I wasn't sure where we would want them. I'll relocate them

src/Aquifer.Jobs/Program.cs
kaseywright: No, all of the local settings values get populated as previous

src/Aquifer.Jobs/Program.cs
kaseywright: During testing and refactoring, I found that the respected values will put `appsettings.{env}.json` higher than `appsettings.json`. That is to say, the same kv pair will use the `appsettings.{env}.json` value if both are set.

src/Aquifer.Jobs/appsettings.example.json
```diff
@@ -0,0 +1,40 @@
...
+        "ApiRequestStatsTableName": "ApiRequestStatsQa",
+        "AppInsightsResourceId": "/subscriptions/dacfa945-984c-4708-9f2d-7d8358f223d5/resourceGroups/BiblioNexus-Dev/providers/microsoft.insights/components/appinsights-qa",
+        "CustomEventsTableName": "CustomEventsQa",
+        "HoursBetweenRuns": 1,
+        "PageViewsTableName": "PageViewsQa",
+        "StorageAccountUri": "https://aquiferstoragedev.table.core.windows.net"
+    },
+    "Aquifer":{
+        "AdminBaseUri": "Base URL of Aquifer API",
+        "ApiBaseUri": "Base URL of the Aquifer Admin"
```
NateMerritt, line 13: These are reversed.
```suggestion
        "AdminBaseUri": "Base URL of Aquifer Admin website",
        "ApiBaseUri": "Base URL of the Aquifer Internal API"
```

src/Aquifer.Jobs/appsettings.json
```diff
@@ -0,0 +1,40 @@
...
+            },
+            "SendProjectStartedNotification": {
+                "TemplateId": "d-7760ec3b5ce34b179384d4783cc1bd81"
+            },
+            "SendResourceCommentCreatedNotification": {
+                "TemplateId": "d-ea84b461ed0f439589098053f8810189"
+            }
+        }
+    },
+    "KeVaultUri": "https://key-vault-name.vault.azure.net/"
```
NateMerritt, line 39:
```suggestion
    "KeyVaultUri": ""
```

src/Aquifer.Jobs/appsettings.json
```diff
@@ -0,0 +1,40 @@
...
+        "ApplicationInsights": "Application insights connection string. Get from Azure instance.",
+        "AzureStorageAccount": "Storage connection string",
+        "BiblioNexusDb": "DB connection string"
+    },
+    "Email": {
+        "SubjectPrefix": "[Test Email - Local] ",
+        "Marketing": { 
+            "Address": "Email address that the new content email is coming from",
+            "Name": "Name that the new content email is under when email is sent",
+            "ResourceLink": "Link to resources in new content email template"
```
NateMerritt, line 25: Instead of leaving in the examples we should make these all empty instead.  See https://github.com/BiblioNexusStudio/aquifer-server/blob/master/src/Aquifer.API/appsettings.json as an example.

src/Aquifer.Jobs/appsettings.json
```diff
@@ -0,0 +1,40 @@
+{
+    "Analytics": {
+        "ApiManagementResourceId": "/subscriptions/dacfa945-984c-4708-9f2d-7d8358f223d5/resourceGroups/BiblioNexus-Global/providers/Microsoft.ApiManagement/service/aquifer-api-management",
+        "ApiRequestStatsTableName": "ApiRequestStatsQa",
+        "AppInsightsResourceId": "/subscriptions/dacfa945-984c-4708-9f2d-7d8358f223d5/resourceGroups/BiblioNexus-Dev/providers/microsoft.insights/components/appinsights-qa",
+        "CustomEventsTableName": "CustomEventsQa",
+        "HoursBetweenRuns": 1,
+        "PageViewsTableName": "PageViewsQa",
+        "StorageAccountUri": "https://aquiferstoragedev.table.core.windows.net"
```
NateMerritt, line 9: A bunch of these are QA specific values and should be removed (see below comment).

src/Aquifer.Jobs/local.settings.example.json
```diff
@@ -1,32 +1,19 @@
 {
+    "IsDevelopment": true,
```
NateMerritt, line 2: Can this move to appsettings.json instead?

kaseywright: When testing, I could not get it to have any effect using it from appsettings. If we find that it does work in appsettings, I'm happy to move it

src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs
```diff
@@ -33,11 +32,35 @@ public class AnalyticsOptions
 public class EmailOptions
 {
     public required string SubjectPrefix { get; init; }
+    public required MarketingEmailOptions Marketing { get; init; }
```
NateMerritt, line 35: Note: This will require updating the environment variable names in QA and Prod to follow the new naming convention.

kaseywright: Do we want to move it back to where it was then or just make a mention of the needed change in slack?

NateMerritt: I'd put it back and then put a `NotificationsOptions` at the root level named `Notifications` with the three template IDs in it.

src/Aquifer.Jobs/local.settings.example.json
NateMerritt: Does it need to move back into Values, then?  https://github.com/BiblioNexusStudio/aquifer-server/pull/523/files#diff-13c6ac85989c85e36799e1fed24dba5f1e9d5ef74df72bc0a3c976396d0acaebR5 requires it to exist in the settings.

NateMerritt (approved):

PR #525, opened 2024-11-04
rename and normalize Bible tables

MrGrinst:


src/Aquifer.Data/Entities/BibleTextEntity.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Aquifer.Data.Entities;
+
+[Index(nameof(BibleId), nameof(BookId), nameof(ChapterNumber), nameof(VerseNumber))]
+public class BibleTextEntity
+{
+    public int Id { get; set; }
+    public int BibleId { get; set; }
+    public BookId BookId { get; set; }
+    public int ChapterNumber { get; set; }
```
jwinston-bn, line 12: Can this just be `BookChapterId`?

MrGrinst: I went with that initially but the problem is versification. We would end up tying the wrong things together. For instance the Psalms in Russian have most chapters offset by 1. This way we end up with the accurate Book Chapter Verse for each Bible and can handle versification later.

jwinston-bn: Okay, that makes sense. Good thinking.

jwinston-bn (approved):

PR #529, opened 2024-11-07
Make Subscribers, Publishers, and Managers consistent.

NateMerritt:
This PR is almost all renames using automated refactors:
* Name all services that publish messages to a queue `*MessagePublisher`.
* Use similar method names for publishing: `Publish*Message`.
* Ensure all message models have consistent naming: `Verb*Message`.
* Name all classes that listen to a queue `*MessageSubscriber`.
* Name all timer trigger function classes `*Manager`.
* Move things into appropriate folders.

One thing that did change: The `EmailService` and `EmailMessagePublisher` no longer share a model because they are no longer abstracted to be essentially the same thing.  I tested this locally to make sure it still works.

This will change Azure function names because we name them after the class they are in.  I believe that will be fine.


src/Aquifer.Jobs/Services/SendGridEmailService.cs
```diff
@@ -101,7 +125,7 @@ public override string ToString()
         return $"""
                 {Message}
 
-                Email:
```
MrGrinst, line 104: This looks like it may be unintended?

jwinston-bn (approved):

src/Aquifer.Jobs/Services/SendGridEmailService.cs
NateMerritt: Yes.  Too much renaming.  Fixed: https://github.com/BiblioNexusStudio/aquifer-server/pull/529/commits/acd8a9d498c6109824ea2e52a901352630e21e9c

MrGrinst (approved):

PR #537, opened 2024-11-11
use JS-based Tiptap parser for generating HTML on public API

MrGrinst:


src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
```diff
@@ -32,6 +32,8 @@ All endpoints require an API key in the `api-key` header.<br><br>
 
                                  <br><br>
                                  Example applications using the Aquifer API can be found in the <a href="https://github.com/BiblioNexusStudio/aquifer-api-samples" target="_blank">samples repository on GitHub</a>.
+
+                                 <a href="https://cdn.aquifer.bible/training/webpage-search-with-react-demo.mp4" target="_blank">React Sample Application Demo</a>
```
MrGrinst, line 36: Snuck this in as well

src/Aquifer.Common/Tiptap/TiptapModel.cs
```diff
@@ -17,20 +16,28 @@ public class TiptapRoot<TRootContent>
     public List<TRootContent> Content { get; set; } = [];
 }
 
-public class TiptapRootContent<TRootContent>
+public class TiptapNode<TRootContent>
 {
     public virtual string Type { get; set; } = null!;
-    public virtual TiptapAttributes? Attrs { get; set; }
+
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
```
jwinston-bn, line 23: Are these needed since it's the default option for `JsonUtilities.DefaultDeserialize`?

MrGrinst: I also changed some other stuff, but when I tried the first time I was getting a bunch of nulls back. I'll see if my other changes fixed that and I can undo these.

MrGrinst: @jwinston-bn just tested and it does still include the `null` without specifying this attribute:

image

jwinston-bn (approved):

PR #539, opened 2024-11-11
retry IP address lookup if it fails

MrGrinst:


NateMerritt:
If we can instead wire up retries via Polly at the DI level it makes things cleaner and allows us to apply the same retry policy to anywhere that we are using `HttpClient`.  Example: https://github.com/NateMerritt/AvaloniaPrototype/blob/430ccf68e44bf19ae54595a5c97c496843f6dca8/AvaloniaPrototype/AvaloniaPrototype/App.axaml.cs#L78-L94

NateMerritt (approved):

PR #540, opened 2024-11-12
Updated Entities For Review

carlosmorales125:


src/Aquifer.Data/Entities/TranslationPairEntity.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Aquifer.Data.Entities;
+
+[EntityTypeConfiguration(typeof(TranslationPairEntityConfiguration))]
+public class TranslationPairEntity
+{
+    public int Id { get; set; }
+    public required LanguageEntity Language { get; set; }
+    public required int LanguageId { get; set; }
+    public required string Key { get; set; }
+    public required string Value { get; set; }
```
MrGrinst, line 13: I know this was on the Linear ticket, but these names feel kind of generic. Would something like `SourcePhrase` and `TargetPhrase` make more sense?

src/Aquifer.Data/Entities/TranslationPairEntity.cs
```diff
@@ -0,0 +1,28 @@
...
+    [SqlDefaultValue("getutcdate()")]
+    public required DateTime Updated { get; set; }
+}
+
+public class TranslationPairEntityConfiguration : IEntityTypeConfiguration<TranslationPairEntity>
+{
+    public void Configure(EntityTypeBuilder<TranslationPairEntity> builder)
+    {
+        builder.HasOne(x => x.Language).WithMany(l => l.TranslationPairs).HasForeignKey(tp => tp.LanguageId);
+    }
```
MrGrinst, line 27: Does it work without this piece? I think since we're specifying `Language` above and then have `TranslationPairs` in the `LanguageEntity` class it should be able to calculate this on its own.

src/Aquifer.Data/Entities/TranslationPairEntity.cs
carlosmorales125: I think when it's paired with the entity name `TranslationPair.Key` and `TranslationPair.Value`, it makes a lot of sense.

src/Aquifer.Data/Entities/TranslationPairEntity.cs
```diff
@@ -0,0 +1,28 @@
...
+}
```
jwinston-bn, line 28: I don't think you should need all this. Should just be able to add `public LanguageEntity Language { get; set; } = null!;`

src/Aquifer.Data/Entities/TranslationPairEntity.cs
```diff
@@ -0,0 +1,28 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace Aquifer.Data.Entities;
+
+[EntityTypeConfiguration(typeof(TranslationPairEntityConfiguration))]
+public class TranslationPairEntity
+{
+    public int Id { get; set; }
+    public required LanguageEntity Language { get; set; }
```
jwinston-bn, line 10: Need to remove `required` from all these

src/Aquifer.Data/Entities/CompanyLanguageEntity.cs
jwinston-bn, line 1: Need to all all these new entities to AquiferDbContext

PR #549, opened 2024-11-12
update paths for post-merge workflows to account for new projects

MrGrinst:


.github/workflows/post-merge-internal-api.yml
```diff
@@ -6,9 +6,13 @@ on:
             - master
             - 'release-hotfix-*'
         paths:
-            - 'src/Aquifer.Common/**'
+            - 'Directory.Build.*'
```
NateMerritt, line 9: Can just be `Directory.Build.props` instead of wildcard. Also, we need the new `Directory.Packages.props` file included.

MrGrinst: Oops I misread and thought the other was `Directory.Build.packages`

NateMerritt (approved):

PR #555, opened 2024-11-14
Add unique index for translation pair, up EF tools version

jwinston-bn:


MrGrinst (approved):

.config/dotnet-tools.json
```diff
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "dotnet-ef": {
-      "version": "8.0.7",
+      "version": "9.0.0",
```
NateMerritt, line 6: I didn't think to look here for the .net 9 update.  Thanks for fixing.

PR #556, opened 2024-11-15
fix permission logic for fetching a project

MrGrinst:
Also updates some of the Aquifer.Tiptap TS stuff.


src/Aquifer.API/Endpoints/Projects/Get/Endpoint.cs
```diff
@@ -17,7 +17,7 @@ public override void Configure()
 
     public override async Task HandleAsync(Request req, CancellationToken ct)
     {
-        if (userService.HasPermission(PermissionName.ReadProjectsInCompany) && !await HasSameCompanyAsProject(req.ProjectId, ct))
+        if (!userService.HasPermission(PermissionName.ReadProject) && !await HasSameCompanyAsProject(req.ProjectId, ct))
```
MrGrinst, line 20: The logic was reversed here which meant that if the user had both `ReadProjectsInCompany` and `ReadProject` then they would be restricted as if they only had `ReadProjectsInCompany`.

src/Aquifer.Tiptap/Aquifer.Tiptap.esproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.VisualStudio.JavaScript.Sdk/1.0.1184077">
   <PropertyGroup>
-    <BuildCommand>npm upgrade &amp;&amp; npx esbuild node_modules/aquifer-tiptap/src/TiptapUtilities.ts --bundle --outfile=dist/TiptapUtilities.js --platform=node --target=node12</BuildCommand>
+    <BuildCommand>npm install &amp;&amp; npm upgrade aquifer-tiptap &amp;&amp; npx esbuild node_modules/aquifer-tiptap/src/TiptapUtilities.ts --bundle --outfile=dist/TiptapUtilities.js --platform=node --target=node12</BuildCommand>
```
MrGrinst, line 3: This should prevent unwanted package-lock.json updates.

jwinston-bn (approved):

PR #558, opened 2024-11-15
update CI/CD docs

MrGrinst:


.github/workflows/post-merge-jobs.yml
```diff
@@ -20,8 +32,22 @@ permissions:
     contents: write
 
 jobs:
+    check_should_deploy:
```
MrGrinst, line 35: Note: we should do a test deploy to jobs after this merges to make sure this new config works. It should be fine since I copied it over to have parity with Public API and Internal API.

jwinston-bn (approved):

PR #560, opened 2024-11-18
Correctly update the ResourceContent's ContentUpdated column.

NateMerritt:


src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -364,6 +364,7 @@ private async Task TranslateResourceCoreAsync(int resourceContentId,
         resourceContentVersion.DisplayName = translatedDisplayName;
         resourceContentVersion.Content = translatedContentJson;
         resourceContentVersion.ContentSize = Encoding.UTF8.GetByteCount(translatedContentJson);
+        resourceContentVersion.Updated = DateTime.UtcNow;
```
MrGrinst, line 367: I think this should be handled by our middleware: https://github.com/BiblioNexusStudio/aquifer-server/blob/master/src/Aquifer.Data/EventHandlers/UpdatedTimestampHandler.cs#L6

PR #561, opened 2024-11-18
handle AI aquiferizations

MrGrinst:


src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -252,19 +251,38 @@ private async Task TranslateResourceCoreAsync(int resourceContentId,
...
+
+        if (isAquiferization && resourceContentLanguage.Id != _englishLanguageId)
+        {
+            throw new InvalidOperationException(
+                $"Aborting translation for Resource Content ID {resourceContentId} because status indicates Aquiferization but content language is {resourceContentLanguage.EnglishDisplay}.");
+        }
+
+        if (!isAquiferization && resourceContentLanguage.Id == _englishLanguageId)
+        {
+            throw new InvalidOperationException(
```
MrGrinst, line 275: Wondering if this is ok to error on. Before it was being logged as a graceful failure but it seems like something we'd want to actually blow up.

src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -252,19 +251,38 @@ private async Task TranslateResourceCoreAsync(int resourceContentId,
             throw new InvalidOperationException(
                 $"Aborting translation for Resource Content ID {resourceContentId} because it has no Resource Content Version with Text Media Type in the Draft status.");
```
NateMerritt, line 252: These exceptions should say `aquiferization` instead of `translation` when relevant to avoid confusion.

src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -252,19 +251,38 @@ private async Task TranslateResourceCoreAsync(int resourceContentId,
...
             throw new InvalidOperationException(
                 $"Aborting translation for Resource Content ID {resourceContentId} because there are no pre-existing Snapshots for Resource Content Version ID {resourceContentVersion.Id}.");
         }
 
+        var resourceContentLanguage = resourceContentVersion.ResourceContent.Language;
+
+        if (isAquiferization && resourceContentLanguage.Id != _englishLanguageId)
+        {
+            throw new InvalidOperationException(
+                $"Aborting translation for Resource Content ID {resourceContentId} because status indicates Aquiferization but content language is {resourceContentLanguage.EnglishDisplay}.");
```
NateMerritt, line 270:
```suggestion
                $"Aborting aquiferization for Resource Content ID {resourceContentId} because status indicates Aquiferization but content language is {resourceContentLanguage.EnglishDisplay}.");
```

src/Aquifer.AI/OpenAiTranslationService.cs
```diff
@@ -37,6 +37,7 @@ public class OpenAiOptions
 public sealed class OpenAiTranslationOptions
 {
     public required string DefaultLanguageSpecificPromptAppendixFormatString { get; init; }
+    public required string HtmlAquiferizationBasePrompt { get; init; }
```
NateMerritt, line 40: Now that this service does Aquiferization I'm not sold on the existing name for the service and the options but I can't think of anything better that isn't really long or more confusing (`TransmutationService`? 😆)

MrGrinst: Yeah I'd briefly thought about renaming or going through everywhere in the message subscriber and putting `translation/aquiferization` for comments and such. If you squint and view aquiferization as a subset of translation (English -> global English) then it feels like the naming still works. But still open to doing a rename if there's something better.

src/Aquifer.Jobs/appsettings.json
```diff
@@ -29,6 +29,7 @@
...
... the HTML with an altered version of the text therein. Never change the formatting of the HTML elements or attributes. For example, if text has an <em> tag around it or a <span> tag with attributes, that tag and its attributes should stay with the text even if it is simplified or moved within the text. Keep the HTML tags and attributes no matter what. Never add a missing html element (i.e. if something starts with <p> but the end is missing </p>, don't add it). Never remove '&nbsp;'.\n\nImprove the clarity of the provided text for a global English audience. Follow these guidelines:\n- Simplify complex sentences for easier understanding.\n- Replace difficult words or technical jargon with simpler alternatives.\n- Add parenthetical explanations when needed.\n- Keep sentences under 25 words.\n- Break long paragraphs into shorter ones.\n- Avoid idioms or figurative language; use straightforward expressions.\n- Use the active voice whenever natural.\n- Add subheadings to improve readability for longer texts.\n- Spell out abbreviations, except BC and AD.\n- Aim for an 8th-9th grade reading level.\n- Preserve the original meaning and intent.\n- Keep the word count close to the original.",
```
MrGrinst, line 32: Here's what this looks like with the newlines formatted. @jwinston-bn what do you think of this?
```
You receive HTML and then return the HTML with an altered version of the text therein. Never change the formatting of the HTML elements or attributes. For example, if text has an <em> tag around it or a <span> tag with attributes, that tag and its attributes should stay with the text even if it is simplified or moved within the text. Keep the HTML tags and attributes no matter what. Never add a missing html element (i.e. if something starts with <p> but the end is missing </p>, don't add it). Never remove '&nbsp;'.

Improve the clarity of the provided text for a global English audience. Follow these guidelines:
- Simplify complex sentences for easier understanding.
- Replace difficult words or technical jargon with simpler alternatives.
- Add parenthetical explanations when needed.
- Keep sentences under 25 words.
- Break long paragraphs into shorter ones.
- Avoid idioms or figurative language; use straightforward expressions.
- Use the active voice whenever natural.
- Add subheadings to improve readability for longer texts.
- Spell out abbreviations, except BC and AD.
- Aim for an 8th-9th grade reading level.
- Preserve the original meaning and intent.
- Keep the word count close to the original.
```

src/Aquifer.API/Endpoints/Resources/Content/Helpers.cs
```diff
@@ -68,20 +73,36 @@ public static async Task CreateNewDraft(AquiferDbContext dbContext,
         {
             await historyService.AddAssignedUserHistoryAsync(newResourceContentVersion, assignedUserId, user.Id, ct);
         }
-        else
-        {
-            await historyService.AddSnapshotHistoryAsync(newResourceContentVersion, user.Id, ResourceContentStatus.New, ct);
-        }
 
         var resourceContent = await dbContext.ResourceContents.FindAsync([contentId], ct) ?? throw new ArgumentNullException();
```
jwinston-bn, line 77: We should be using `First` or `Single`. .NET 9 introduced significant performance increases to LINQ, and it's much faster than `Find` now (it may not make much difference when constructing a query to the database, but for consistency, I would argue we shouldn't ever use `Find` now).

src/Aquifer.API/Endpoints/Resources/Content/Helpers.cs
```diff
@@ -68,20 +73,36 @@ public static async Task CreateNewDraft(AquiferDbContext dbContext,
...
-            resourceContent.Updated = DateTime.UtcNow;
+            await historyService.AddSnapshotHistoryAsync(newResourceContentVersion, user.Id, ResourceContentStatus.New, ct);
         }
 
-        await historyService.AddStatusHistoryAsync(newResourceContentVersion, ResourceContentStatus.AquiferizeEditorReview, user.Id, ct);
+        if (resourceContent.LanguageId == Constants.EnglishLanguageId)
+        {
+            resourceContent.Status = ResourceContentStatus.AquiferizeAwaitingAiDraft;
+            await historyService.AddStatusHistoryAsync(newResourceContentVersion, ResourceContentStatus.AquiferizeAwaitingAiDraft, user.Id, ct);
+            await dbContext.SaveChangesAsync(ct);
```
jwinston-bn, line 88: Do we need this? It's been left to the caller otherwise, and where you're calling this calls `SaveChangesAsync` after.

src/Aquifer.Jobs/appsettings.json
jwinston-bn: Did you test out a long paragraph to see what would happen with the html/tiptap if it broke it up into multiple paragraphs? I'm wondering if we need to do anything to translate a new line to wrap in `<p>` tags or something.

MrGrinst: Good call, I'll test that

src/Aquifer.API/Endpoints/Resources/Content/Helpers.cs
MrGrinst: From what I can tell it looks like `Find` is actually on `DbSet` and doesn't exist on Enumerables in LINQ. Seems nice to keep using for simple primary key lookups so I don't see much downside unless I'm missing something.

src/Aquifer.API/Endpoints/Resources/Content/Helpers.cs
MrGrinst: Yeah I don't love this either, but since we publish the message right after this I think we need to make sure it's saved so it will always be in the right state when the job gets it. An alternative would be to move the publish message to each of the two usages after the save happens. Thoughts?

src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -399,7 +411,7 @@ private async Task TranslateResourceCoreAsync(int resourceContentId,
         {
             await _resourceHistoryService.AddSnapshotHistoryAsync(resourceContentVersion,
                 startedByUserId,
-                ResourceContentStatus.TranslationAwaitingAiDraft,
+                awaitingStatus,
```
NateMerritt, line 414: The comment above also needs to be updated.

NateMerritt (approved):
I mostly reviewed the `OpenAiTranslationService` and `TranslationMessageSubscriber` and they look good.

NateMerritt (approved):

PR #565, opened 2024-11-19
Apply translation pair replacements after minifying HTML.

NateMerritt:
Translation pairs were swapping out HTML attribute text.  This won't perfectly solve the problem but it will remove all span contents before doing translation pair swaps.

The only way to do it without any risk of messing with HTML would be to parse the HTML using something like HtmlAgilityPack, only do replacements on HTML text content, and then `ToString()` it back to HTML.

### DIFF
--- a/src/Aquifer.AI/OpenAiTranslationService.cs
+++ b/src/Aquifer.AI/OpenAiTranslationService.cs
@@ -108,7 +108,7 @@ public sealed partial class OpenAiTranslationService : ITranslationService
 
         return isFullReplacement
             ? textWithReplacements
-            : await TranslateTextAsync(prompt, text, cancellationToken);
+            : await TranslateTextAsync(prompt, textWithReplacements, cancellationToken);
     }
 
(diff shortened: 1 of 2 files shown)


src/Aquifer.AI/OpenAiTranslationService.cs
```diff
@@ -154,6 +150,42 @@ public async Task<string> TranslateHtmlAsync(
...
+                return (pair.Value, true);
+            }
+
+            text = Regex.Replace(text, $"""\b(?:{pair.Key})\b""", pair.Value, RegexOptions.IgnoreCase);
+        }
+
+        return (text, false);
+    }
+
+    private async Task<string> TranslateTextAsync(string prompt, string text, CancellationToken cancellationToken)
```
NateMerritt, line 168: This could probably be merged into the method below but I'm not going to worry about that right now.

MrGrinst (approved):

PR #571, opened 2024-11-19
Update existing snapshot during retranslation instead of creating a new one.

NateMerritt:


src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -393,12 +393,28 @@ private async Task TranslateResourceCoreAsync(int resourceContentId,
...
+        {
+            await _resourceHistoryService.AddSnapshotHistoryAsync(resourceContentVersion,
+                startedByUserId,
+                ResourceContentStatus.TranslationAwaitingAiDraft,
+                ct);
+        }
+        // If a retranslation then update the existing snapshot to not create too many snapshots.
+        else
+        {
+            var existingAiTranslationSnapshot = resourceContentVersion.ResourceContentVersionSnapshots
```
MrGrinst, line 408: Do we need to handle if this is null or will `Last` blow up if it can't find one?

NateMerritt: It will blow up if it can't find one.  If that happens we want it to blow up because a dev put a bad retranslate message on the queue but we could admittedly have a better error message.

PR #574, opened 2024-11-21
remove old Bible tables and deprecated endpoint

MrGrinst:


src/Aquifer.API/Endpoints/Resources/Content/AssignPublisherReview/Endpoint.cs
```diff
@@ -12,11 +12,9 @@ namespace Aquifer.API.Endpoints.Resources.Content.AssignPublisherReview;
 
 public class Endpoint(AquiferDbContext dbContext, IResourceHistoryService historyService, IUserService userService) : Endpoint<Request>
 {
-    // TODO: Deprecate/Delete in favor of  SendForPublisherReview after next deploy
```
MrGrinst, line 15: @elisabeth-shakespeare can you take a look at this part? I removed this comment because I'm pretty sure this is doing something different than `SendForPublisherReview` is, but wanted to confirm.

elisabeth-shakespeare: It's a little different from SendForPublisherReview in that in can do a subset of what SendForPublisher can do. SendForPublisher will also sanitize the tiptap content and has ChangedByPublisher in its response. 

SendForPublisher does all that this can do and more, this is an ambiguous endpoint.

kaseywright (approved):

PR #579, opened 2024-11-21
Add handler for bad translation pair keys

jwinston-bn:
Originally was going to use an interceptor (hence the branch name), but just adding to the saving event was simpler. Noticed that the other method was misnamed, and it runs after the records are already saved (so that was a nice waste of time).

### DIFF
--- a/src/Aquifer.Data/AquiferDbContext.cs
+++ b/src/Aquifer.Data/AquiferDbContext.cs
@@ -18,7 +18,8 @@ public class AquiferDbContext : DbContext
         // a wrapper around it. https://www.jetbrains.com/help/resharper/VirtualMemberCallInConstructor.html
         // It's likely irrelevant anyway, because the events are additive.
         ChangeTracker.StateChanged += OnStateChange;
-        SavedChanges += async (s, e) => await OnSavingChanges(s, e);
+        SavedChanges += async (s, e) => await OnSavedChanges(s, e);
+        SavingChanges += OnSavingChanges;
     }
 
     public DbSet<AssociatedResourceEntity> AssociatedResources { get; set; }
@@ -100,9 +101,15 @@ public class AquiferDbContext : DbContext
         UpdatedTimestampHandler.Handle(e.Entry);
     }
 
-    private async Task OnSavingChanges(object? sender, SavedChangesEventArgs e)
+    private async Task OnSavedChanges(object? sender, SavedChangesEventArgs e)
     {
         var entries = ChangeTracker.Entries();
         await ResourceStatusChangeHandler.HandleAsync(_options, entries);
+    }
+
+    private void OnSavingChanges(object? sender, SavingChangesEventArgs e)
+    {
+        var entries = ChangeTracker.Entries();
+        BadTranslationPairHandler.Handle(entries);
     }
 }
--- a/src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
+++ b/src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
@@ -7,7 +7,11 @@ namespace Aquifer.Data.EventHandlers;
 
 public static class BadTranslationPairHandler
 {
-    private static readonly IReadOnlyList<string> s_KeyBlacklist = ["span", "div"];
+    private static readonly IReadOnlySet<string> s_KeyBlacklist = new HashSet<string>
+    {
+        "span",
+        "div"
+    };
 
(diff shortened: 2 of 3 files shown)


src/Aquifer.Data/AquiferDbContext.cs
```diff
@@ -18,7 +18,8 @@ public AquiferDbContext(DbContextOptions<AquiferDbContext> options) : base(optio
         // a wrapper around it. https://www.jetbrains.com/help/resharper/VirtualMemberCallInConstructor.html
         // It's likely irrelevant anyway, because the events are additive.
         ChangeTracker.StateChanged += OnStateChange;
-        SavedChanges += async (s, e) => await OnSavingChanges(s, e);
+        SavedChanges += async (s, e) => await OnSavedChanges(s, e);
```
NateMerritt, line 21:
```suggestion
        SavedChanges += OnSavedChanges;
```

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Aquifer.Data.Entities;
+using Aquifer.Data.Exceptions;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+
+namespace Aquifer.Data.EventHandlers;
+
+public static class BadTranslationPairHandler
+{
+    private static readonly IReadOnlyList<string> s_KeyBlacklist = ["span", "div"];
```
NateMerritt, line 10: Others to consider: `p`, `h1`

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public static void Handle(IEnumerable<EntityEntry> entityEntries)
+    {
+        entityEntries.Where(entry => entry is { Entity: TranslationPairEntity, State: EntityState.Modified or EntityState.Added })
+            .Select(x => x.Entity as TranslationPairEntity)
+            .ToList()
```
NateMerritt, line 16: Is `ToList()` needed?

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
NateMerritt, line 10: Optimization: An `IReadOnlySet<T>` will be faster for `Contains` lookups.

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+            .ForEach(x =>
+            {
+                if (s_KeyBlacklist.Contains(x?.Key))
+                {
+                    throw new TranslationPairKeyNotAllowedException();
```
NateMerritt, line 21: I suggest putting the `key` into the message so we know why it exploded if we see an alert.

src/Aquifer.Data/AquiferDbContext.cs
jwinston-bn: That doesn't work.

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
jwinston-bn: I think we're gonna disallow anything below two characters. I'll just blanket disallow it.

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
jwinston-bn: Yup

NateMerritt (approved):

src/Aquifer.Data/EventHandlers/BadTranslationPairHandler.cs
NateMerritt: Oh, interesting.  It only works on a `List<T>` and not an `IEnumerable<T>`.  I have historically avoided `ForEach()` like the plague and prefer to just do:
```
foreach (var item in Chain().Of().Linq().Commands())
{
    DoTheThing();
}
```
This makes it clear that a mutation is happening (otherwise we'd use a `Select()`).  It also allows for the removal of the `ToList()` call as a micro-optimization.

PR #580, opened 2024-11-21
Catching Translation Pair Not Allowed Exception

carlosmorales125:


src/Aquifer.API/Endpoints/TranslationPairs/Patch/Endpoint.cs
```diff
@@ -43,7 +44,16 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+            await dbContext.SaveChangesAsync(ct);
+        }
+        catch (TranslationPairKeyNotAllowedException)
+        {
+            AddError("Keys must be at least 3 characters long. Some keywords are not allowed.");
+            await SendErrorsAsync(400, ct);
+            return;
+        }
 
         await SendNoContentAsync(ct);
```
jwinston-bn, line 58: I would move this into the `try` block and then you can remove the `return` from the `catch`.

jwinston-bn (approved):

PR #585, opened 2024-12-02
Fix resource assignment notifications when a large number of resources have been assigned.

NateMerritt:


src/Aquifer.Jobs/Managers/SendResourceAssignmentNotificationsManager.cs
```diff
@@ -68,8 +82,10 @@ public async Task RunAsync([TimerTrigger(CronSchedules.EveryTenMinutes)] TimerIn
                     {
                         [EmailMessagePublisher.DynamicTemplateDataSubjectPropertyName] = "Aquifer Notification: Resources Assigned",
                         ["aquiferAdminBaseUri"] = _configurationOptions.Value.AquiferAdminBaseUri,
-                        ["resourceCount"] = userGrouping.Count(),
+                        ["resourceCount"] = countOfResourcesAssignedToUser,
+                        ["additionalResourceCount"] = Math.Max(countOfResourcesAssignedToUser - _maxNumberOfResourcesToDisplayInNotificationContent, 0),
                         ["parentResources"] = userGrouping
+                            .Take(_maxNumberOfResourcesToDisplayInNotificationContent)
```
NateMerritt, line 88: We could potentially improve this to spread out resources displayed in order to include one for each parent resource but the number of emails where that would be valuable are likely very small.

PR #589, opened 2024-12-03
Separate translation and text improvement and implement post-processing stubs.

NateMerritt:
Note that Aquiferization is now part of the English "text improvement" step and follows the same flow as the other languages.


src/Aquifer.Jobs/Services/TranslationPostProcessingService.cs
```diff
@@ -0,0 +1,73 @@
...
+        var languageSpecificTranslationPostProcessingService =
+            _languageSpecificTranslationPostProcessingServiceByIso6393CodeMap.GetValueOrDefault(languageIso6393Code);
+
+        return languageSpecificTranslationPostProcessingService == null
+            ? html
+            : await languageSpecificTranslationPostProcessingService.PostProcessHtmlAsync(html, cancellationToken);
+    }
+}
+
+public sealed class EnglishTranslationPostProcessingService : ILanguageSpecificTranslationPostProcessingService
```
NateMerritt, line 55: We could split out these multiple classes into separate files if desired.  Right now they are just stubs so I left them here to make things easier to discover.

src/Aquifer.Jobs/Services/TranslationPostProcessingService.cs
```diff
@@ -0,0 +1,73 @@
...
+    public Task<string> PostProcessHtmlAsync(string html, CancellationToken cancellationToken);
+}
+
+public static class TranslationPostProcessingServicesRegistry
+{
+    public static IServiceCollection AddTranslationPostProcessingServices(this IServiceCollection services)
+    {
+        services.AddSingleton<ITranslationPostProcessingService, TranslationPostProcessingService>();
+
+        foreach (var implementationType in AppDomain.CurrentDomain.GetAssemblies().SelectMany(s => s.GetTypes())
```
NateMerritt, line 23: Any new language-specific post-processing services will automatically be included without requiring any explicit DI registration due to this reflection.

I also tried out a keyed services DI strategy (.net 8+) but I like this one better.

src/Aquifer.Jobs/Services/TranslationPostProcessingService.cs
NateMerritt: We might also want to delete them but the performance impact is negligible.

src/Aquifer.Jobs/appsettings.json
```diff
@@ -28,19 +28,20 @@
...
+        "HtmlBasePrompt": "You receive HTML and then return the HTML with an altered version of the text therein. Never change the formatting of the HTML elements or attributes. For example, if text has an <em> tag around it or a <span> tag with attributes, that tag and its attributes should stay with the text even if it is simplified or moved within the text. Keep the HTML tags and attributes no matter what. Never add a missing html element (i.e. if something starts with <p> but the end is missing </p>, don't add it). Never remove '&nbsp;'.",
+        "LanguageSpecificTextImprovementPromptAppendixByLanguageIso6393CodeMap": {
+            "ENG": "Simplify the text for a global audience. Aim for an 8th grade reading level. Add parenthetical explanations if needed. Keep sentences under 25 words. Use the active voice whenever natural. Spell out abbreviations, except BC and AD.",
```
jwinston-bn, line 33: Can we add "Avoid using contractions." somewhere in here please.

src/Aquifer.Jobs/appsettings.json
```diff
@@ -28,19 +28,20 @@
...
         },
+        "PlainTextTranslationPromptFormatString": "You receive a string and then return that string in the {0} language. Translate the exact string that you receive. Do not interpret it as anything other than the exact string that it is.",
         "Temperature": 0.2,
-        "TextTranslationPromptFormatString": "You receive a string and then return that string in the {0} language. Translate the exact string that you receive. Do not interpret it as anything other than the exact string that it is."
+        "TextImprovementPromptFormatString": "Improve the grammar and clarity of the provided {0} text. Never echo the original content, only respond with the alteration.",
+        "TranslationPromptFormatString": "Never echo the original content, only respond with the translation. Then translate to {0}."
```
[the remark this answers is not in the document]
jwinston-bn: Or maybe even "only respond with the alteration" is better here too.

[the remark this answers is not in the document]
NateMerritt: The current ordering in this PR is consistent with the way that they were previously because the old [HTML translation base prompt](https://github.com/BiblioNexusStudio/aquifer-server/blob/a5aae8cbfb5345a548d1b6077cacab708e227e48/src/Aquifer.Jobs/appsettings.json#L33) was followed by the [default translate prompt](https://github.com/BiblioNexusStudio/aquifer-server/blob/a5aae8cbfb5345a548d1b6077cacab708e227e48/src/Aquifer.Jobs/appsettings.json#L31).

I split out the "HTML" part so that it could be used for both translation and text improvement (including Aquiferization where it was previously mostly copy/pasted).

Do you still want to swap them?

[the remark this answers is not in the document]
NateMerritt: I went ahead [and swapped them](https://github.com/BiblioNexusStudio/aquifer-server/pull/589/commits/c3ca98abdac57da8f10bcf461a7b5930d0140adb) because that's consistent with the text improvement prompt.

src/Aquifer.Jobs/appsettings.json
NateMerritt: [Done](https://github.com/BiblioNexusStudio/aquifer-server/pull/589/commits/5f4a2e193c67884892c4feac73b71f4288445b39).

jwinston-bn (approved):

PR #590, opened 2024-12-04
Improve manager error handling.

NateMerritt:
* Use Manager base class with default exception handling and logging.
* Use explicit retry policies for Azure clients (it looks like there is a default policy already).
* Use retry policy for SendGridClient (the default is to not retry).


src/Aquifer.Jobs/Managers/ManagerBase.cs
```diff
@@ -0,0 +1,125 @@
...
+/// 
+///     protected override async Task RunCoreAsync(CancellationToken cancellationToken)
+///     {
+///         Logger.LogInformation("Function implementation goes here");
+///     }
+/// }
+/// </code>
+/// </example>
+/// </remarks>
+[SuppressMessage("ReSharper", "StaticMemberInGenericType")]
```
NateMerritt, line 32: This is fine because there is one static instance per `T` which is desired.

src/Aquifer.Jobs/Managers/SendNewContentEmailsManager.cs
```diff
@@ -7,20 +7,27 @@
...
-    TelemetryClient telemetryClient)
+    TelemetryClient telemetryClient,
+    ILogger<SendNewContentEmailsManager> logger)
+    : ManagerBase<SendNewContentEmailsManager>(logger)
 {
     [Function(nameof(SendNewContentEmailsManager))]
-#pragma warning disable IDE0060 // Remove unused parameter: A (non-discard) TimerInfo parameter is required for correct Azure bindings
-    public async Task Run([TimerTrigger(CronSchedules.FirstOfMonthAtNoon)] TimerInfo timerInfo, CancellationToken ct)
-#pragma warning restore IDE0060 // Remove unused parameter
+    [FixedDelayRetry(maxRetryCount: 1, Timings.TenSecondDelayInterval)]
```
NateMerritt, line 24: I added a single ten second delay retry to all of the existing timer functions.

jwinston-bn (approved):

PR #591, opened 2024-12-05
See if this fixes informational logging not showing in functions

jwinston-bn:


src/Aquifer.Jobs/Program.cs
```diff
@@ -21,14 +21,9 @@
...
-    .ConfigureAppConfiguration((context, builder) => builder
-        .SetBasePath(context.HostingEnvironment.ContentRootPath)
-        .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
-        .AddJsonFile(
-            $"appsettings.{context.HostingEnvironment.EnvironmentName}.json",
-            optional: true,
-            reloadOnChange: true)
-    )
+    .ConfigureAppConfiguration((context, builder) => builder.SetBasePath(context.HostingEnvironment.ContentRootPath)
+        .AddJsonFile("appsettings.json", false, true)
```
NateMerritt, line 25: Did your formatter remove the argument names or was that manual?  I'm a fan of explicit argument names for literal types because I've seen too many bugs where `Method(true, false)` should have been `Method(false, true)` and showing the argument names makes this bug much more detectable.

src/Aquifer.Jobs/host.json
```diff
@@ -1,60 +1,53 @@
 {
     "version": "2.0",
-
     "logging": {
         "logLevel": {
             "default": "Information",
-            "Azure.Core": "Error" // suppress noisy SDK warnings
+            "Azure.Core": "Error"
+            // suppress noisy SDK warnings
```
NateMerritt, line 7: This comment applies to the preceding line and should be moved above it if the formatter doesn't like same line comments.

src/Aquifer.Jobs/host.json
```diff
@@ -1,60 +1,53 @@
...
+        },
+        "ApplicationInsights": {
+            "LogLevel": {
+                "Default": "Information"
```
NateMerritt, line 11: 👍

NateMerritt (approved):

src/Aquifer.Jobs/Program.cs
jwinston-bn: It did.

PR #592, opened 2024-12-05
Add tables for versification

elisabeth-shakespeare:


src/Aquifer.Data/Entities/VersificationMappingEntity.cs
```diff
@@ -0,0 +1,21 @@
+using Aquifer.Data.EventHandlers;
+
+namespace Aquifer.Data.Entities;
+
+public class VersificationMappingEntity : IHasUpdatedTimestamp
+{
+    public int Id { get; private set; }
```
elisabeth-shakespeare, line 7: I kept thinking about making a composite key with this, but I started thinking it may have too many pieces to it to keep it unique, so I thought maybe an Id was best... not sure if that was the right choice though

src/Aquifer.Data/Entities/VersificationExclusionEntity.cs
```diff
@@ -0,0 +1,17 @@
...
+    public int Id { get; set; }
+    public int BibleId { get; set; }
+    public int BibleVerseId { get; set; }
+    public BibleEntity Bible { get; set; } = null!;
+    public VerseEntity Verse { get; set; } = null!;
+    [SqlDefaultValue("getutcdate()")]
+    public DateTime Created { get; set; } = DateTime.UtcNow;
+
+    [SqlDefaultValue("getutcdate()")]
+    public DateTime Updated { get; set; } = DateTime.UtcNow;
```
jwinston-bn, line 16: I don't think these are going to get updated really. Don't think we need this on either one.

src/Aquifer.Data/Entities/VersificationExclusionEntity.cs
```diff
@@ -0,0 +1,17 @@
+using Aquifer.Data.EventHandlers;
+
+namespace Aquifer.Data.Entities;
+
+public class VersificationExclusionEntity : IHasUpdatedTimestamp
+{
+    public int Id { get; set; }
```
jwinston-bn, line 7: I don't think there's ever a situation where we look this up by `Id`. I'd just make a dual primary key on `BibleId` and `BibleVerseId`.

src/Aquifer.Data/Entities/VersificationMappingEntity.cs
jwinston-bn: Yeah, I agree that this one has a lot of pieces to it to make a unique item. @NateMerritt Do you have any thoughts on this? The combination of `BibleId`, `BibleVerseId`, and `VerseIdPart` should be unique. Is there any issue putting a unique index on those three with one of them being nullable char, or a triple primary key?

elisabeth-shakespeare: I think it would also be the `BaseVerseId` and maybe even `BaseVerseIdPart` as well, because of the many to many relationships.

NateMerritt: The PK column value cannot be `null`.  I haven't tested that out with multi-column PKs but I'm expecting it to still apply.  A unique constraint should work, though.

I've generally avoided multi-column PKs for a few reasons:
1. Insert performance: An auto-incrementing `int` or `bigint` PK column performs better on inserts because multi-column inserts can be anywhere in the primary index but for auto-incrementing PKs any new row is always added at the end of the primary index.  Inserting into the middle of the index (B-tree) eventually requires rebalancing as part of the insert operation which can take many seconds for larger tables (I've seen it take over 30 seconds for a table with multiple millions of rows).  That said, this data should never have inserts/updates, only possible re-loads, so that's not really an issue.
2. FKs to the PK require multiple columns which makes queries more complex because all columns must be in the WHERE clause.
3. API usage: Generally you want a single unique ID for Get by ID requests when a resource is exposed over an API.

I do think they make sense in some usages, though.

jwinston-bn (approved):

PR #593, opened 2024-12-05
Add Public API integration tests.

NateMerritt:
This PR implements integration testing of the Public API via `FastEndpoints.Testing` and `WebApplicationFactory` which create an in-memory API service which uses an `HttpClient` that doesn't actually make any network calls in order to test the `Endpoint`s using the existing `Request` and `Response` objects.

The build server won't pass the tests until we add environment variables as GitHub Secrets for the needed `appsettings.json` overrides (see "Edit Workflow" [here](https://medium.com/@gabreguarnieri/keep-your-keys-safe-while-testing-and-deploying-your-application-414a9e15b32f) for an example).  If I get good feedback on this PR I'll work on adding those.

### DIFF
--- a/tests/Aquifer.Public.API.IntegrationTests/AppFixture.cs
+++ b/tests/Aquifer.Public.API.IntegrationTests/AppFixture.cs
@@ -1,0 +1,66 @@
+﻿using FastEndpoints.Testing;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Aquifer.Public.API.IntegrationTests;
+
+/// <summary>
+/// This is a FastEndpoints <see cref="AppFixture"/> that sits on top of a <see cref="WebApplicationFactory{TEntryPoint}"/>
+/// which allows for in-memory web requests to be sent and received without actual network traffic.
+/// Details:
+/// * https://fast-endpoints.com/docs/integration-unit-testing
+/// * https://learn.microsoft.com/en-us/aspnet/core/test/integration-tests
+/// </summary>
+public sealed class AppFixture : AppFixture<Program>
+{
+    /// <summary>
+    /// Configure Clients here.
+    /// The default Client is anonymous and has no API Key header value.
+    /// That is sufficient for Public API tests because all routes are anonymous and no actual web requests are sent via this fixture.
+    /// </summary>
+    /// <remarks>
+    /// Example of adding a default header to the existing client:
+    /// <example>
+    /// <code>
+    /// Client.DefaultRequestHeaders.Add("api-key", "TODO");
+    /// </code>
+    /// </example>
+    /// An entirely new client could be defined in this class and set up in this method as well (e.g. an authenticated client).
+    /// </remarks>
+    protected override Task SetupAsync()
+    {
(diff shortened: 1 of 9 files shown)


tests/Aquifer.Public.API.IntegrationTests/BiblesListTestFixture.cs
```diff
@@ -0,0 +1,25 @@
...
+using Aquifer.Public.API.Endpoints.Bibles.List;
+using FastEndpoints.Testing;
+using FastEndpoints;
+
+namespace Aquifer.Public.API.IntegrationTests;
+
+public sealed class BiblesListTestFixture(AppFixture _appFixture) : TestBase<AppFixture>
+{
+    [Fact]
+    public async Task ValidRequest_ShouldReturnSuccess()
```
NateMerritt, line 11: This isn't the most ideal test but my goal in this PR is to demonstrate that the integration tests work.  I can come back later and write some better tests for the API endpoints if desired.

tests/Aquifer.Public.API.IntegrationTests/AppFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Aquifer.Public.API.IntegrationTests;
+
+/// <summary>
+/// This is a FastEndpoints <see cref="AppFixture"/> that sits on top of a <see cref="WebApplicationFactory{TEntryPoint}"/>
+/// which allows for in-memory web requests to be sent and received without actual network traffic.
+/// Details:
+/// * https://fast-endpoints.com/docs/integration-unit-testing
+/// * https://learn.microsoft.com/en-us/aspnet/core/test/integration-tests
+/// </summary>
+public sealed class AppFixture : AppFixture<Program>
```
NateMerritt, line 16: I added a bunch of documentation to this class assuming that we'll copy/paste it for the Internal API (which will need an authenticated client) and eventual Mobile API.

tests/Directory.Build.props
```diff
@@ -8,16 +8,10 @@
...
     <ItemGroup>
-        <PackageReference Update="coverlet.collector">
-            <PrivateAssets>all</PrivateAssets>
-            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-        </PackageReference>
+        <PackageReference Include="coverlet.collector" />
         <PackageReference Include="Microsoft.NET.Test.Sdk" />
         <PackageReference Include="xunit" />
-        <PackageReference Update="xunit.runner.visualstudio">
-            <PrivateAssets>all</PrivateAssets>
```
NateMerritt, line 18: Using `PrivateAssets` here (which I think I added when I switched to `Directory.Build.props`) was preventing test discovery so `dotnet test` was "passing" with `0` discovered tests.  🤦‍♂️

tests/Aquifer.Public.API.IntegrationTests/AppFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// Configure Clients here.
+    /// The default Client is anonymous and has no API Key header value.
+    /// That is sufficient for Public API tests because all routes are anonymous and no actual web requests are sent via this fixture.
```
jwinston-bn, line 21: Probably worth adding the point that the API key is enforced by Azure API Management, and that's why it's not needed locally.

jwinston-bn (approved):

PR #594, opened 2024-12-06
on delete no action for bibleVerseId

elisabeth-shakespeare:


src/Aquifer.Data/Migrations/AquiferDbContextModelSnapshot.cs
```diff
@@ -2718,7 +2718,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.HasOne("Aquifer.Data.Entities.VerseEntity", "BibleVerse")
                         .WithMany()
                         .HasForeignKey("BibleVerseId")
-                        .OnDelete(DeleteBehavior.Cascade)
+                        .OnDelete(DeleteBehavior.NoAction)
```
elisabeth-shakespeare, line 2721: should I do this for other Verse keys just in case?

jwinston-bn: This is a generated file, shouldn't be updating it directly. Need to do it like in `ResourceContentVersionAssignedUserHistoryEntity.cs`, and redo the migrations.

kaseywright (dismissed):

jwinston-bn (approved):

PR #601, opened 2024-12-10
Generate Clients for Public API.

NateMerritt:
There are now four new routes under `/docs/index.html?url=/swagger/v1/swagger.json#tag/Clients`, each of which downloads a zip file containing client source code for C#, TypeScript, Python, and Java, respectively.  For example, to download the C# client perform a GET request to `/clients/cs`.

This PR also adds substantial documentation to each of those routes about how to use the client.  The C# documentation is fully featured with a working example including how to inject the API Key into the header of every request; if we like this then examples for other languages could also be added.  See the screenshot below.  We could also consider moving this example into a standalone GitHub repo with examples for each language.

Concerns I found when consuming the generated C# client:
* The current OpenAPI spec is somewhat lacking in terms of schema names for response objects so the generated clients will use `Response1`, `Response2`, etc. (and the same for requests).  We could solve this by renaming the C# classes with descriptive names.
* We have many response properties that are `object`s in C# which get serialized to JSON but have no spec about how to deserialize them.  The ideal solution to this is to provide strongly typed objects for C# usage instead of sending JSON straight from the DB to the client (this would potentially solve `LicenseInfo` properties assuming there is a standard schema).  However, properties like a `resource.Content` have multiple representations and thus can never be strongly typed.  Therefore, I provided a method in the example C# code to convert the `UntypedNode` representation of these objects in the generated client models back to the original JSON string.  Consumers will need to write their own deserialization scheme for such properties.

Screenshot of C# client documentation:
image

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,7 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <WarningsNotAsErrors>NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
+        <WarningsNotAsErrors>NU1608;NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
         <NoWarn>MSB3245</NoWarn>
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <GitHubOrganization>BiblioNexusStudio</GitHubOrganization>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,8 +6,7 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <WarningsNotAsErrors>NU1608;NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
-        <NoWarn>MSB3245</NoWarn>
+        <WarningsNotAsErrors>MSB3245;NU1608;NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <GitHubOrganization>BiblioNexusStudio</GitHubOrganization>
         <RepositoryName>aquifer-server</RepositoryName>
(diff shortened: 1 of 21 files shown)


Directory.Build.props
```diff
@@ -6,7 +6,7 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <WarningsNotAsErrors>NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
+        <WarningsNotAsErrors>NU1608;NU1901;NU1902;NU1903;NU1904</WarningsNotAsErrors>
```
NateMerritt, line 9: This worked for my local VS build but the build server is still failing.  I'll figure it out tomorrow.

Motivation: We need to use `System.Text.Json` 9.0.0 due to .net 9 targeting but the new Kiota dependencies want < 9.0.0 so there is a build warning that we can just ignore.

src/Aquifer.Public.API/OpenApi/ClientGenerationSettings.cs
```diff
@@ -0,0 +1,202 @@
...
+    {
+        app.MapApiClientEndpoint(
+            $"/{_clientsRouteName}/cs",
+            c =>
+            {
+                c.SwaggerDocumentName = swaggerDocumentName;
+                c.Language = GenerationLanguage.CSharp;
+                c.ClientNamespaceName = "BiblioNexus.Aquifer.API.Client";
+                c.ClientClassName = "AquiferClient";
+                c.OutputPath = Path.Combine(c.OutputPath, "cs", Path.GetRandomFileName());
```
NateMerritt, line 28: I think there's a bug in FastEndpoints where if you specify `CleanOutput = true` the client generation blows up if you use a non-default `OutputPath`.  I worked around this by just using a different temp folder every time we generate a client (which never gets erased until the Temp folder is cleared by the OS).

Directory.Build.props
NateMerritt: Fix: [commit](https://github.com/BiblioNexusStudio/aquifer-server/pull/601/commits/5708803049260106a50e5402827afa43321464ec).  I _don't_ want the warnings totally hidden from the build output (via `NoWarn`) because they do deserve to eventually be fixed by updating the impacted NuGet packages but right now there's nothing we can do to fix them because we are on the latest versions.  This commit ensures that they still warn but that they don't fail the build.

jwinston-bn (approved):

PR #602, opened 2024-12-12
Add French language post processor

jwinston-bn:


src/Aquifer.Common/Utilities/HtmlUtilities.cs
```diff
@@ -93,6 +93,37 @@ static string ExpandHtml(string minifiedHtml, IReadOnlyDictionary<int, string> r
         }
     }
 
+    /// <summary>
+    ///     Extracts the text of each HTML node and runs the processor on the text, replacing the extracted text with the result.
+    ///     Thus, processing should only act on the human-readable content, not directly elements or attribute markdown.
```
NateMerritt, line 98: Pre-existing:
```suggestion
    ///     Thus, processing should only act on the human-readable content, not directly on elements or attribute markdown.
```

src/Aquifer.Common/Utilities/HtmlUtilities.cs
```diff
@@ -30,25 +32,23 @@ public static int GetWordCount(string html)
...
                 .Where(n => n.NodeType == HtmlNodeType.Text && !s_nodeNamesToSkip.Contains(n.ParentNode.Name))
                 .Select(n => WebUtility.HtmlDecode(n.InnerText).Trim())
                 .Where(t => !string.IsNullOrWhiteSpace(t)));
     }
 
     /// <summary>
-    /// Minifies HTML element attributes before processing and then reinstates the original element attributes after processing.
-    /// Thus, processing should only act on the human-readable content, not directly elements or attribute markdown.
+    ///     Minifies HTML element attributes before processing and then reinstates the original element attributes after processing.
+    ///     Thus, processing should only act on the human-readable content, not directly elements or attribute markdown.
```
NateMerritt, line 51: Pre-existing:
```suggestion
    ///     Thus, processing should only act on the human-readable content, not directly on elements or attribute markdown.
```

tests/Aquifer.Common.UnitTests/HtmlUtilitiesTests.cs
```diff
@@ -7,33 +7,25 @@ public sealed class HtmlUtilitiesTests
...
-                            <!-- This is a comment.-->
-                                Test  &nbsp;   *HTML.
-                            </p>
-                        </div>
-                    </body>
-                    """,
-                PlainText: "Test  \u00a0   *HTML.",
-                WordCount: 2
-            ),
+            (Html: /* lang=html */ """
```
NateMerritt, line 10: These whitespace changes are going to cause conflicts with my refactoring PR: https://github.com/BiblioNexusStudio/aquifer-server/pull/603

I also moved the test class into a `Utilities` folder.

src/Aquifer.Jobs/Services/TranslationProcessors/FrenchTranslationProcessingService.cs
```diff
@@ -0,0 +1,148 @@
...
+    private static readonly IReadOnlySet<string> s_RegexItemsThatHaveNnbspBeforeSelf = new HashSet<string>
+    {
+        "»",
+        "!",
+        @"\?",
+        ":",
+        ";"
+    };
+
+    private readonly string _nnbspAfterPattern = @$"({string.Join("|", s_RegexItemsThatHaveNnbspAfterSelf)})(?=[A-Za-zà-üÀ-Ü0-9\x20])";
```
NateMerritt, line 31: Can a regular space be used instead of `\x20`?

src/Aquifer.Jobs/Services/TranslationProcessors/FrenchTranslationProcessingService.cs
```diff
@@ -0,0 +1,148 @@
+﻿using System.Text.RegularExpressions;
+using Aquifer.Common.Utilities;
+
+namespace Aquifer.Jobs.Services.TranslationProcessors;
+
+public sealed partial class FrenchTranslationProcessingService : ILanguageSpecificTranslationProcessingService
+{
+    private const char NarrowNonBreakingSpace = '\u202f';
+    private const char NonBreakingSpace = '\u00a0';
```
NateMerritt, line 9: Changing these to `string`s would save doing a bunch of unnecessary string interpolation below.  They are only consumed as `string`s, anyway.

jwinston-bn: There are a couple `Replace` that are using them as `char`, but you're right that overall it's causing more hassle.

src/Aquifer.Jobs/Services/TranslationProcessors/FrenchTranslationProcessingService.cs
```diff
@@ -0,0 +1,148 @@
...
+            .Replace("J.-C.", "J.‑C.");
+
+        return Task.FromResult(text);
+    }
+
+    private Task<string> SetCorrectBibleBookAbbreviation(string text)
+    {
+        text = BibleBookRegex()
+            .Replace(text,
+                m => m.Value.ToLower().Replace(NarrowNonBreakingSpace, ' ') switch
```
NateMerritt: I generally suggest a `Dictionary<string, string>` for something like this though pattern matching does make it concise at least.  I believe `switch` pattern matching will ultimately write all of this out in the IL as conditional `if`/`else`/`elseif` logic whereas a `Dictionary` would use `O(N)` lookup time of a data structure.

src/Aquifer.Jobs/Services/TranslationProcessors/FrenchTranslationProcessingService.cs
jwinston-bn: Oh, for clarity. I think it's easy to miss a trailing space like that and think it's unintentional. This way it's explicit what's intended.

src/Aquifer.Jobs/Services/TranslationProcessors/FrenchTranslationProcessingService.cs
```diff
@@ -0,0 +1,148 @@
...
+                    "matt" => "Mt",
+                    "mk" or "marc" => "Mc",
+                    "tob" => "Tb",
+                    "1 macc" => "1\u202fM",
+                    "2 macc" => "2\u202fM",
+                    "sa" => "Sg",
+                    "sir" => "Si",
+                    "bar" => "Ba",
+                    "lk" or "luc" => "Lc",
+                    "jean" or "john" => "Jn",
```
NateMerritt, line 118: tl;dr: Concern: Bible book name detection is difficult. If we are detecting a string that is already an abbreviation then it seems pretty safe to do the conversion.  I'm thinking it would be best to remove the full words and full proper names and catch them with manual review.

Some of these might be too catch all, especially with the `ToLower()` comparison.
Examples:
* If any of these are an actual French word then they'll be replaced (I don't know if any of these apply).  This could potentially be solved by not doing a `ToLower()` assuming the abbreviations usually are capitalized but it still wouldn't help for usage of the word at the beginning of a sentence.
* In this case, `Jean` is a proper name which we don't always want to abbreviate (e.g. `St. Jean Baptiste` would become `St. Jn Baptiste` which isn't desired).  The same is potentially true for `Esdras` above.

jwinston-bn: The regex accounts for this. There has to be a number followed by a space or period for it to match.

jwinston-bn: image

NateMerritt (approved):

PR #607, opened 2024-12-16
bib-808-discovery-versification2

elisabeth-shakespeare:


src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
```diff
@@ -0,0 +1,34 @@
...
+public sealed class CachingVersificationService(AquiferDbContext _dbContext) : ICachingVersificationService
+{
+    private const string VersificationMappingsCacheKey = $"{nameof(CachingVersificationService)}VersificationMappings";
+    private const string VersificationBiblesCacheKey = $"{nameof(CachingVersificationService)}VersificationBibles";
+    // private static readonly TimeSpan s_cacheLifetime = TimeSpan.FromMinutes(30);
+
+    public async Task<int> GetVersificationMapping(int bibleId, int verseId, CancellationToken ct)
+    {
+        var englishMapping =
+            await _dbContext.VersificationMappings.FirstOrDefaultAsync(vm => vm.BibleId == 1 && vm.BibleVerseId == verseId, ct);
```
jwinston-bn, line 21: Ideally we'd load everything in on startup.

src/Aquifer.Common/Utilities/BibleTextUtilities.cs
```diff
@@ -0,0 +1,42 @@
+using Aquifer.Common.Services.Caching;
+using Aquifer.Data;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.Common.Utilities;
+
+public static class BibleTextUtilities
+{
+    public static async Task<IEnumerable<Chapter>> GetTextsByBibleId(int bookNumber, int chapter, int verse,
```
jwinston-bn, line 9: I think this would ideally use the VerseId from the Verses table, or there'd be an overload of this that took it.

src/Aquifer.Common/Utilities/BibleTextUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+        int bibleId, AquiferDbContext dbContext, ICachingVersificationService cachingService, CancellationToken ct)
+    {
+        var verseId = BibleUtilities.GetVerseId(bookNumber, chapter, verse);
+
+        var mappedVerseId = await cachingService.GetVersificationMapping(bibleId, verseId, ct);
+        var (mappedBookNumber, mappedChapter, mappedVerse) = BibleUtilities.TranslateVerseId(mappedVerseId);
+
+        var bibleTexts = await dbContext.BibleTexts
```
jwinston-bn, line 17: I think you need a method that specifically handles transformations, because that may be all we care about. I think getting the transformations and getting texts are two separate things.

elisabeth-shakespeare: makes sense. I'll split out the transformation part and have this call that

PR #610, opened 2024-12-17
Fix machine translation updates to allow any user to edit.

NateMerritt:
See https://biblionexusworkspace.slack.com/archives/C0662HZ7RKM/p1734458335235189

### DIFF
--- a/src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
@@ -19,13 +19,12 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService) : En
         var existingMt = await dbContext.ResourceContentVersionMachineTranslations.AsTracking()
             .FirstOrDefaultAsync(x =>
                     x.Id == req.Id &&
-                    x.ResourceContentVersion.AssignedUserId == user.Id &&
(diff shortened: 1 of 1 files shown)


src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
```diff
@@ -19,13 +19,12 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
         var existingMt = await dbContext.ResourceContentVersionMachineTranslations.AsTracking()
             .FirstOrDefaultAsync(x =>
                     x.Id == req.Id &&
-                    x.ResourceContentVersion.AssignedUserId == user.Id &&
```
jwinston-bn, line 22: The problem with this is now anyone can update it. It's not a realistic scenario, but someone could go through and set a rating for everything in this status regardless of if they've worked on it or looked at it.

NateMerritt: Do we need to lock it down to specific roles/permissions?

jwinston-bn: Assigned editor is who sets it, so everyone has the permission.

jwinston-bn: There's some sort of timing issue where the front-end reassigns before this gets set. I think the issue is on the front-end.

PR #612, opened 2024-12-18
GET resource/content/{contentId} performance fix

jwinston-bn:


src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -150,19 +157,22 @@ await dbContext.ResourceContentVersionAssignedUserHistory
...
-            UserId = mt.UserId,
-            UserRating = mt.UserRating,
-            ImproveClarity = mt.ImproveClarity,
-            ImproveConsistency = mt.ImproveConsistency,
-            ImproveTone = mt.ImproveTone,
-            HadRetranslation = mt.RetranslationReason != null,
-            Created = mt.Created
-        })
-            .ToList();
+        // Can this move into the IsDraft check below?
```
jwinston-bn, line 160: Need to check what the front-end is doing here, and since I want to hotfix this, will save for another time.

NateMerritt, line 160: Did you intend to commit this comment?

NateMerritt (approved):

PR #614, opened 2024-12-18
Add Versification Cache and Bible Texts utility

elisabeth-shakespeare:


elisabeth-shakespeare:
Unit tests coming soon

src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    private const string VersificationDictionariesCacheKey = $"{nameof(CachingVersificationService)}:VersificationDictionaries";
+    private static readonly TimeSpan s_cacheLifetime = TimeSpan.FromMinutes(30);
+
+    public async Task<Dictionary<(int bibleVerseId, char? bibleVersePart), (int baseVerseId, char? baseVersePart)>>
+        GetVersificationsByBibleIdAsync(int bibleId, CancellationToken cancellationToken)
+    {
+        var dictionaries = await GetDictionariesFromCacheAsync(cancellationToken);
+
+        var versificationDictionary = dictionaries?.GetValueOrDefault(bibleId) ?? throw new KeyNotFoundException();
```
NateMerritt, line 23: If this code is assuming that all Bibles will have versification data in the DB (a reasonable assumption) then I suggest throwing an `InvalidOperationException()` with a good error message stating that assumption.

elisabeth-shakespeare: So far, I've been under the impression that it's not assuming that there will be versification for every Bible

src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
```diff
@@ -0,0 +1,42 @@
+using Aquifer.Data;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace Aquifer.Common.Services.Caching;
+
+public interface ICachingVersificationService
+{
+    Task<Dictionary<(int bibleVerseId, char? bibleVersePart), (int baseVerseId, char? baseVersePart)>> GetVersificationsByBibleIdAsync(
```
NateMerritt, line 9: I suggest returning a `ReadOnlyDictionary`.  Callers should never modify this return value because if they do it will modify the cached dictionary and impact all future callers.  Using a `ReadOnlyDictionary` incurs a small performance cost but protects against that.

elisabeth-shakespeare: good catch.

src/Aquifer.Common/Utilities/BibleTextUtilities.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        const int fromBibleId = 1;
+        var bibleTexts = new List<FlatBibleText>();
+        try
+        {
+            var verses = await dbContext.Verses.Where(v => v.Id >= startVerseId && v.Id <= endVerseId).Select(v => v.Id).ToListAsync(ct);
+            var fromBibleToBaseMapping = await versificationService.GetVersificationsByBibleIdAsync(fromBibleId, ct);
+            var toBibleToBaseMapping = await versificationService.GetVersificationsByBibleIdAsync(toBibleId, ct);
+
+            var fromBibleToBibleMapping = GetMergedMappingsOnBaseId(fromBibleToBaseMapping, toBibleToBaseMapping);
```
elisabeth-shakespeare, line 22: I think these can be handled by VersificationCaching, where we just send it a fromBibleId, toBibleId and get the dictionary back.

elisabeth-shakespeare: On second thought, should probably make a `VersificationUtilities` because it only merges what comes out of the cache and doesn't cache the `fromBibleIdToBibleId` map.......

src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
jwinston-bn: Definitely wouldn't assume they'll all have it.

jwinston-bn: Seems like we'd just return an empty dictionary here rather than blow up.

src/Aquifer.Common/Utilities/BibleTextUtilities.cs
```diff
@@ -0,0 +1,125 @@
...
+            if (mappedVerseId.Equals(default))
+            {
+                mappedVerseId = verse;
+            }
+
+            var (mappedBookId, mappedChapter, mappedVerseNumber) = BibleUtilities.TranslateVerseId(mappedVerseId);
+
+            var (_, tChapter, tVerse) = BibleUtilities.TranslateVerseId(verse);
+
+            var mappedText = await dbContext.BibleTexts.Where(bt =>
```
jwinston-bn, line 82: We should never be going to the database in a loop like this. What if I pass the whole book of Genesis? How many hits to the database is that?

elisabeth-shakespeare: Sorry, I knew better than to do that

src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
```diff
@@ -0,0 +1,109 @@
...
+    public async Task<ReadOnlySet<int>> GetExclusionsByBibleIdAsync(int bibleId, CancellationToken cancellationToken)
+    {
+        var dictionaries = await GetExclusionListsFromCacheAsync(cancellationToken);
+        var exclusions = dictionaries?.GetValueOrDefault(bibleId) ?? new ReadOnlySet<int>(new HashSet<int>());
+
+        return exclusions;
+    }
+
+    private async Task<ReadOnlyDictionary<int,
+            ReadOnlyDictionary<int, int>>?>
```
NateMerritt, line 62: Nit: I'd put the return type all on one line above for better readability.

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public async Task<List<BibleText>> GetBibleTextsForBibleIdAsync(int toBibleId, int startVerseId, int endVerseId, CancellationToken ct)
+    {
+        const int fromBibleId = 1;
+        var mappedStartVerseId =
+            await VersificationUtilities.GetVersificationAsync(startVerseId, fromBibleId, toBibleId, versificationService, ct);
+        var mappedEndVerseId =
+            await VersificationUtilities.GetVersificationAsync(endVerseId, fromBibleId, toBibleId, versificationService, ct);
+
+        var (bookId, startChapter, startVerse) = BibleUtilities.TranslateVerseId(mappedStartVerseId);
```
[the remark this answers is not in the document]
NateMerritt: That still won't solve it for edge cases where a verse is significant out of order.  Again, though, I don't think we should change this logic until we see examples in practice of such mappings.  I just wanted to call it out here so we know the limitations of the current algorithm.

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,58 @@
+using Aquifer.Common.Services.Caching;
+using Aquifer.Common.Utilities;
+using Aquifer.Data;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.Common.Services;
+
+public interface IBibleTextService
+{
+    Task<List<BibleText>> GetBibleTextsForBibleIdAsync(int toBibleId, int startVerseId, int endVerseId, CancellationToken ct);
```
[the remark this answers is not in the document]
jwinston-bn: I'm not sure there's anything existing that's necessarily the BSB. That just happens to be our default English. But the content that we import we have no idea what specific Bible their references are based on. I would lean toward this remaining the generic `startVerseId`, because that aligns with our database tables that have no knowledge of Bible ids.

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+public sealed class BibleTextService(AquiferDbContext dbContext, ICachingVersificationService versificationService) : IBibleTextService
+{
+    public async Task<List<BibleText>> GetBibleTextsForBibleIdAsync(int toBibleId, int startVerseId, int endVerseId, CancellationToken ct)
+    {
+        const int fromBibleId = 1;
+        bool skipMapping = toBibleId == fromBibleId;
```
jwinston-bn, line 18: Right or wrong, our editorconfig does prefer `var` here.

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,57 @@
...
+        var mappedStartVerseId = skipMapping ? startVerseId : await VersificationUtilities.GetVersificationAsync(startVerseId, fromBibleId, toBibleId, versificationService, ct);
+        var mappedEndVerseId = skipMapping ? endVerseId : await VersificationUtilities.GetVersificationAsync(endVerseId, fromBibleId, toBibleId, versificationService, ct);
```
jwinston-bn, line 20: Check line length here. I recommend setting your IDE to automatically break lines past the editor config setting.

src/Aquifer.Common/Services/BibleTextService.cs
elisabeth-shakespeare: Sounds good. It was a suggested Rider change to change it to const for me for some reason

src/Aquifer.Common/Services/BibleTextService.cs
jwinston-bn: I think this method is weird, and I maintain that getting Bible text shouldn't care about the versification. If this method takes a BibleId, a start verse id, and an end verse id, it might be completely unexpected what I get from this. I guess this is the reasoning for the above comment about making it `bsbStartVerseId`, but it's still this weird method that does translation and fetching, and I think those are two separate concerns. This should take a bible id, and two verse ids, and then that's what you get. This is not the outcome I expect from method called `GetBibleTextsForBibleId`.

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,57 @@
...
+public sealed class BibleTextService(AquiferDbContext dbContext, ICachingVersificationService versificationService) : IBibleTextService
+{
+    public async Task<List<BibleText>> GetBibleTextsForBibleIdAsync(int toBibleId, int startVerseId, int endVerseId, CancellationToken ct)
+    {
+        var fromBibleId = 1;
+        var skipMapping = toBibleId == fromBibleId;
+        var mappedStartVerseId = skipMapping ? startVerseId : await VersificationUtilities.GetVersificationAsync(startVerseId, fromBibleId, toBibleId, versificationService, ct);
+        var mappedEndVerseId = skipMapping ? endVerseId : await VersificationUtilities.GetVersificationAsync(endVerseId, fromBibleId, toBibleId, versificationService, ct);
+
+        var (bookId, startChapter, startVerse) = BibleUtilities.TranslateVerseId(mappedStartVerseId);
```
jwinston-bn, line 22: This method isn't giving any indication that it only works with a single book id. I could pass a range spanning multiple books and this would break.

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,62 @@
...
+public interface IBibleTextService
+{
+    Task<List<BibleText>> GetBibleTextsForBibleIdAsync(int toBibleId, int startVerseId, int endVerseId, CancellationToken ct);
+}
+
+public sealed class BibleTextService(AquiferDbContext dbContext, ICachingVersificationService versificationService) : IBibleTextService
+{
+    public async Task<List<BibleText>> GetBibleTextsForBibleIdAsync(int toBibleId, int startVerseId, int endVerseId, CancellationToken ct)
+    {
+        var fromBibleId = 1;
```
jwinston-bn, line 17: This should be const

src/Aquifer.Common/Services/BibleTextService.cs
```diff
@@ -0,0 +1,62 @@
...
+                    {
+                        VerseNumber = bti.VerseNumber,
+                        Text = bti.Text
+                    }).ToList()
+            })
+            .ToListAsync(ct);
+    }
+}
+
+public class BibleText
```
jwinston-bn, line 52: I would call these `BibleTextChapter` and `BibleTextVerse` to be more explicit.

src/Aquifer.Common/Services/BibleTextService.cs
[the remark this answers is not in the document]
elisabeth-shakespeare: Going to resolve this, but if anyone feels like this needs more conversation feel free to open it back up

src/Aquifer.Common/Services/BibleTextService.cs
elisabeth-shakespeare: Then maybe the method name should change to something like `GetVersifiedTextsForBibleId`?

elisabeth-shakespeare: And then, to mitigate confusion with the `fromBibleId = 1`, we can put in a parameter `sourceBibleId` that defaults to 1

src/Aquifer.Common/Services/BibleTextService.cs
elisabeth-shakespeare: Our endpoints currently get texts within the book. would it be better to rename the method name and add a check or add support for multiple books?

src/Aquifer.Common/Services/BibleTextService.cs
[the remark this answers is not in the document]
elisabeth-shakespeare: Okay, resolving this then

src/Aquifer.Common/Services/BibleTextService.cs
elisabeth-shakespeare: Alright, took out all versification stuff and made it only responsible for getting texts in booknumber, start/endChapter, start/endVerse

src/Aquifer.Common/Services/BibleTextService.cs
elisabeth-shakespeare: add bookNumber as a parameter

NateMerritt (approved):

PR #616, opened 2024-12-18
Use LicenseInfo strongly typed models and make DB column non-nullable.

NateMerritt:
This is a step on the way to using relational DB tables for the license info JSON.  I confirmed that all public API routes match production/QA and the internal API routes all load correctly.  The main difference is that `null`s are now populated:
image

### DIFF
--- a/src/Aquifer.API/Endpoints/Bibles/BibleLicenseInfo.cs
+++ b/src/Aquifer.API/Endpoints/Bibles/BibleLicenseInfo.cs
@@ -1,0 +1,33 @@
+﻿namespace Aquifer.API.Endpoints.Bibles;
+
+public sealed class BibleLicenseInfo
+{
+    public required string Title { get; init; }
+    public required LicenseCopyright Copyright { get; init; }
+    public required IReadOnlyList<LicenseDetails> Licenses { get; init; }
+
+    public sealed class LicenseCopyright
+    {
+        public string? Dates { get; set; }
+        public required LicenseCopyrightHolder Holder { get; init; }
+    }
+
+    public sealed class LicenseCopyrightHolder
+    {
+        public required string Name { get; init; }
+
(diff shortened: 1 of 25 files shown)


src/Aquifer.API/Endpoints/Bibles/BibleLicenseInfo.cs
jwinston-bn, line 1: Why not put this in Common since it's duplicated?

jwinston-bn: Or Aquifer.Data since it's representing something from the db?

src/Aquifer.API/Endpoints/Resources/ResourceLicenseInfo.cs
jwinston-bn, line 1: Same question here

src/Aquifer.API/Endpoints/Bibles/BibleLicenseInfo.cs
NateMerritt: I could put it wherever.  I considered it a response type which is why I duplicated it and put it in a shared location for endpoints to access.  Eventually if we make a DB entity then I'd expect we would map that to this existing response type.

NateMerritt: Oh, an important thing to note is that this shared object's name will show up in the Open API spec and in the auto-generated client for the Public API.  That's one of the reasons I made this PR because the existing `object?` get turns into a very-hard-to-work-with `UntypedNode` in the auto-generated clients.  Therefore having a suffix like `*Entity` in the name would be weird for an API spec and thus I'd argue for specific response types instead of serializing DB models directly.

jwinston-bn (approved):

PR #617, opened 2024-12-19
Bib 809 machine translation scoring db changes

kaseywright:


src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
```diff
@@ -0,0 +1,22 @@
+namespace Aquifer.Data.Entities;
+
+public class ResourceContentVersionSimilarityScore
+{
+    public int Id { get; set; }
```
kaseywright, line 5: Wanted to use a composite primary key on `BaseVersionId` + `ComparedVersionId` but we are able to compare different content from different tables. So, a composite key will not work.

src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
```diff
@@ -0,0 +1,22 @@
...
+    public int BaseVersionId { get; set; }
+    public ResourceContentVersionTypes BaseVersionType { get; set; }
+    public int ComparedVersionId { get; set; }
+    public ResourceContentVersionTypes ComparedVersionType { get; set; }
+    public ResourceContentStatus Status { get; set; }
```
kaseywright, line 10: Not 100% on using the `ResourceContentStatus` to identify the status for the comparison but using a custom status seemed inconsistent.

src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
```diff
@@ -0,0 +1,22 @@
...
+    public double SimilarityScore { get; set; }
+    
+    [SqlDefaultValue("getutcdate()")]
+    public DateTime Created { get; set; } = DateTime.UtcNow;
+}
+
+public enum ResourceContentVersionTypes
+{
+    Base,
```
kaseywright, line 19: We can compare ant two `ResourceContentVersions` , `...MachineTranslations` and `...Snapshots` and have to be able to identify which type of resource we are comparing. Since it can be any of these three, we cannot have a foreign key here.

src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
jwinston-bn: I'm not really sure what goes here. If it's against a MachineTranslation type it probably doesn't have a meaning, and snapshot to snapshot can join on those to get their status at the time the snapshot was created. So unless there's something I'm not thinking of, I would drop this.

src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
jwinston-bn: I like to have `None` as 0, since the default value for an enum is 0. That way it doesn't take on a value if someone doesn't set it.

src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
kaseywright: Okay, I can see dropping it. Below is just a note on how this works in theory for posterity.

From what Frank had described, I got the impression of wanting to know the similarity score between a resource at two different stages (i.e. from machine to published, from editor review to company review, etc.). 
Here is a list to infer the status of a resource comparison:
 - machine to Rcv = published 'complete' content
 - machine to snapshot = whatever status the snapshot is
 - snapshot to snapshot = pick the status you are concerned with
 - Rcv to snapshot = snapshot status

src/Aquifer.Data/Entities/ResourceContentVersionSimilarityScore.cs
kaseywright: This should never have 'none'. It has to be either a base `ResourceContentVersion`, a `ResourceContentVersionMachineTranslation`, or `ResourceContentVersionSnapshot`. Otherwise, the comparison cannot happen.

jwinston-bn: That's my point though. If someone doesn't set it it has a default value that has meaning. So the default shouldn't have meaning.

kaseywright: Ah, bc Mssql doesn't have an enum type. It makes it an int...and that could cause what you described.

jwinston-bn (approved):

PR #618, opened 2024-12-19
Add integration tests to Internal API.

NateMerritt:
I didn't get time to get auth working but we can still add least add tests for the anonymous routes and do auth'd routes later in a separate PR.

I also tried out [FluentAssertions](https://fluentassertions.com/) which I'm liking a lot so I added it to the test project global usings.


tests/Aquifer.Public.API.IntegrationTests/Endpoints/Bibles/List/EndpointTests.cs
```diff
@@ -19,7 +19,8 @@ public async Task ValidRequest_ShouldReturnSuccess()
                 HasGreekAlignment = true,
             });
```
NateMerritt, line 21: This file is a good `FluentAssertions` diff.

jwinston-bn (approved):

PR #620, opened 2024-12-20
Bib 809 machine translation scoring

kaseywright:
This PR addresses the need to evaluate the effectiveness of machine translations and human edits by determining the similarity between two versions of a resource. To achieve this, we will utilize the Levenshtein Distance algorithm, a measure of the minimum number of single-character edits (insertions, deletions, or substitutions) required to change one string into another.

The PR includes a Levenshtein Distance implementation and the necessary utility methods to effectively score long resources. A new queue, generate-resource-content-similarity-score, has been created, along with a corresponding message, publisher, and subscriber. The message will contain information about the resource content versions being compared, including the type of comparison being performed, as designated by one of the values in the ResourceContentVersionSimilarityComparisonTypes enum. This allows the subscriber to run the appropriate logic for the resource content types being compared.


src/Aquifer.API/Endpoints/Resources/Content/Publish/Endpoint.cs
```diff
@@ -86,10 +94,35 @@ await Helpers.CreateNewDraft(dbContext,
...
+                similarityScoreMessages.Add(
+                    new ScoreResourceContentVersionSimilarityMessage(
+                        machineTranslation.Id,
+                        mostRecentContentVersion.Id,
+                        ResourceContentVersionSimilarityComparisonTypes.MachineTranslationToResourceContentVersion)
+                );
+            }
+            else
+            {
+                logger.LogInformation($"No machine translation found for published content version {mostRecentContentVersion.Id}.");
```
kaseywright, line 114: Open to suggestions on how to best log this as we do not want to interrupt the request but we also might be interested in which published resources do not have machine translations.

src/Aquifer.API/Endpoints/Resources/Content/Publish/Endpoint.cs
```diff
@@ -86,10 +94,35 @@ await Helpers.CreateNewDraft(dbContext,
...
+            }
         }
 
         await dbContext.SaveChangesAsync(ct);
 
+        foreach (var similarityScoreMessage in similarityScoreMessages)
+        {
+            await scoreResourceContentVersionSimilarityMessagePublisher
+                    .PublishScoreResourceContentVersionSimilarityMessageAsync(similarityScoreMessage, ct);
```
NateMerritt, line 123: At this point we should probably use `CancellationToken.None` instead of passing through the `ct` because we need to publish these messages even if cancellation is requested due to having already saved to the DB.

src/Aquifer.Common/Utilities/StringSimilarityUtilities.cs
```diff
@@ -0,0 +1,117 @@
+using System.Diagnostics;
+
+namespace Aquifer.Common.Utilities;
+
+public static class StringSimilarityUtilities
+{
+    public static (double similarity, double executionTime) ComputeLevenshteinSimilarity(string textA, string textB, int limit = 15000)
+    {
+        var stopwatch = Stopwatch.StartNew();
```
NateMerritt, line 9: Can the stopwatch part be removed now that you're done testing?

src/Aquifer.Jobs/Subscribers/ScoreResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,246 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Aquifer.Jobs.Subscribers;
+
+public sealed class ScoreResourceContentVersionSimilarityMessageSubscriber(
+    AquiferDbContext _dbContext,
+    ILogger<ScoreResourceContentVersionSimilarityMessageSubscriber> _logger)
+{
+    [Function(nameof(ScoreResourceContentVersionSimilarity))]
+    public async Task ScoreResourceContentVersionSimilarity(
```
NateMerritt, line 19: I renamed all of the subscriber functions to end with `MessageSubscriber` to bring clarity when we see the names in the Azure Portal.
```suggestion
    public async Task ScoreResourceContentVersionSimilarityMessageSubscriber(
```
When you make this change you'll need to also rename the class (or choose a different root name for the Function).

src/Aquifer.Jobs/Subscribers/ScoreResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,246 @@
...
+        await queueMessage.ProcessAsync<ScoreResourceContentVersionSimilarityMessage, ScoreResourceContentVersionSimilarityMessageSubscriber>(
+            _logger,
+            nameof(ScoreResourceContentVersionSimilarity),
+            ProcessAsync,
+            ct);
+    }
+
+    private async Task ProcessAsync(QueueMessage queueMessage, ScoreResourceContentVersionSimilarityMessage message, CancellationToken ct)
+    {
+        _logger.LogInformation("Scoring machine translation...");
```
NateMerritt, line 33: I suggest adding more data from the `message` into the log so we know what exactly was being processed.

src/Aquifer.Jobs/Subscribers/ScoreResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        var machineContentResource = await _dbContext
+            .ResourceContentVersionMachineTranslations
+            .AsTracking()
+            .SingleOrDefaultAsync(x => x.Id == machineId, ct) 
+                ?? throw new InvalidOperationException($"Machine translation with id {machineId} not found");
+        
+        var machineText = HtmlUtilities.GetPlainText(machineContentResource.Content);
+        
+
```
NateMerritt, line 79: Nit: Extra newline.

tests/Aquifer.Common.UnitTests/Utilities/StringSimilarityUtilitiesTests.cs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+public class StringPair
+{
+    public required string First { get; set; }
+    public required string Second { get; set; }
+    public int Limit { get; set; }
+}
+
+public class StringDistanceTestData : TheoryData<StringPair, int>
```
NateMerritt, line 58: Each of these classes are only used by a single test method and thus can be moved to `InlineData` attributes on the test method.

src/Aquifer.Common/Messages/Models/ResourceContentVersionSimilarityComparisonTypes.cs
```diff
@@ -0,0 +1,9 @@
+namespace Aquifer.Common.Messages.Models;
+
+public enum ResourceContentVersionSimilarityComparisonTypes
```
NateMerritt, line 3: `Type` is singular for other `enum`s like this.
```suggestion
public enum ResourceContentVersionSimilarityComparisonType
```

src/Aquifer.API/Endpoints/Resources/Content/Publish/Endpoint.cs
NateMerritt: This seems reasonable though I'd add something like `"Skipping similarity scoring."` at the end.

src/Aquifer.API/Endpoints/Resources/Content/Publish/Endpoint.cs
```diff
@@ -86,10 +94,35 @@ await Helpers.CreateNewDraft(dbContext,
 
                 await historyService.AddStatusHistoryAsync(mostRecentContentVersion, ResourceContentStatus.Complete, user.Id, ct);
             }
+            
+            var machineTranslation = await dbContext
```
jwinston-bn, line 98: I would argue that this isn't really the responsibility of the Publish endpoint, and that we would just always add these to the queue, and the message handler can decide if there's anything to actually process or not. @NateMerritt what do you think? A big benefit of the queue is being able to offload processing to a background job, and this is an extra database call now on something that a user is actively waiting for. In addition, because this is in a loop, if someone were to bulk publish 1000 items, that's 1000 extra database calls the user is now waiting on.

src/Aquifer.API/Program.cs
```diff
@@ -47,6 +47,7 @@
     .AddSingleton<IResourceContentRequestTrackingMessagePublisher, ResourceContentRequestTrackingMessagePublisher>()
     .AddSingleton<INotificationMessagePublisher, NotificationMessagePublisher>()
     .AddSingleton<ITranslationMessagePublisher, TranslationMessagePublisher>()
+    .AddSingleton<IResourceContentVersionSimilarityMessagePublisher, ResourceContentVersionSimilarityMessagePublisher>()
```
jwinston-bn, line 50: I think we should pull all these into a separate publisher registration utility, similar to AquiferHttpServices.cs.

src/Aquifer.Common/Utilities/StringSimilarityUtilities.cs
```diff
@@ -0,0 +1,108 @@
...
+    public static double ComputeLevenshteinSimilarity(string textA, string textB, int limit = 15000)
+    {
+        var textASubstrings = GetSubstrings(textA, limit);
+        var textBSubstrings = GetSubstrings(textB, limit);
+        var distance = ComputeTotalStringDistance(textASubstrings, textBSubstrings);
+
+        return 1 - ((double)distance / Math.Max(textA.Length, textB.Length));
+    }
+    
+    public static int LevenshteinDistance(string textA, string textB)
```
jwinston-bn, line 14: This and the ones below should be private methods, but I understand they're made public for the sake of being called in the unit tests project. I would generally expect the private methods to be tested implicitly through the public method's testing. Ultimately I'm arguing against changing the correctness of code to accommodate a unit test. Thoughts? cc: @NateMerritt

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,164 @@
...
+        await queueMessage.ProcessAsync<ScoreResourceContentVersionSimilarityMessage, ResourceContentVersionSimilarityMessageSubscriber>(
+            _logger,
+            nameof(ScoreResourceContentVersionSimilarityMessageSubscriber),
+            ProcessAsync,
+            ct);
+    }
+
+    private async Task ProcessAsync(QueueMessage queueMessage, ScoreResourceContentVersionSimilarityMessage message, CancellationToken ct)
+    {
+        var similarityScore = new ResourceContentVersionSimilarityScore()
```
jwinston-bn, line 33: Nit: Don't need parens here.

jwinston-bn: Looks like we have a naming problem with this object too that got missed before. It should be `ResourceContentVersionSimilarityScoreEntity` to be consistent with all the other database objects.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,164 @@
...
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        return resourceType switch
+        {
+            ResourceContentVersionTypes.Base => await GetBaseVersionContentText(versionId, dbContext, ct),
+            ResourceContentVersionTypes.Snapshot => await GetSnapshotVersionText(versionId, dbContext, ct),
+            ResourceContentVersionTypes.MachineTranslation => await GetMachineTranslationVersionText(versionId, dbContext, ct),
+            ResourceContentVersionTypes.None => throw new InvalidOperationException($"Invalid similarity comparison type: {resourceType}"),
+            _ => throw new InvalidOperationException($"Invalid similarity comparison type: {resourceType}"),
```
jwinston-bn, line 111:
```suggestion
ResourceContentVersionTypes.None or _ => throw new InvalidOperationException(
                $"Invalid similarity comparison type: {resourceType}")
```

src/Aquifer.API/Endpoints/Resources/Content/Publish/Endpoint.cs
NateMerritt: I agree.  @kaseywright and I talked through the implications of this at one point.  I suggested that it should be possible to look up all of the data that we need from just a resource content version ID on the subscriber.  The question we had was if we could still do that lookup correctly if the message wasn't processed immediately (imagine the function was temporarily disabled for a couple hours or a message got put into the poison queue for a week before we noticed) because DB data could change in the meantime.  That needed some further investigation but I think that a `ResourceContentVersionId` would be sufficient whereas a `ResourceContentId` would not be.

One advantage of passing the ID here is that it allows us to have a single subscriber with a single message DTO for all of the difference scenarios but we could still accomplish that by making one of the IDs nullable (or having two Subscribers with shared core logic if necessary).

jwinston-bn: ResourceContentId lookup could potentially change the data, but ResourceContentVersionId lookup would not. Once published, the ResourceContentVersion can't be changed. So let's move forward with that change please.

src/Aquifer.Common/Utilities/StringSimilarityUtilities.cs
NateMerritt: I noticed this in review and think that `public` is acceptable for this method because it is the true Levenshtein computation and thus is still a valid utility method that could be useful for a different caller.  The other method that does the chunking is just our shim on the top.

However, the methods below are truly implementation detail methods (`GetSubstrings` and `ComputeTotalStringDistance` and should be `private`).

For the general question of changing visibility for unit testing, note that the [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-9.0) in a `csproj` definition can allow specifying that a unit test project can see a given project's internal members.  Then any methods or classes under test could be marked as `internal` if necessary.  I don't necessarily like changing visibility just for tests but this is better than making them `public` IMO.  The only other way to test would be to use reflection to look up `private` methods by string name (I've done this before but eww).

jwinston-bn: I've done that before where they're `internal`, but for practical purposes I don't think it makes any real difference (it's not like this is some public library project that's being used externally). And agreed that reflection would be a worse outcome. So I'd go back to my original thing, the public methods should implicitly be testing the private ones.

NateMerritt: > So I'd go back to my original thing, the public methods should implicitly be testing the private ones.

💯

src/Aquifer.API/Endpoints/Resources/Content/Publish/Endpoint.cs
kaseywright: When the ResourceContentVersionId being supplied has more than one MachineTranslation, do we want to default to scoring against the first one, last one, etc? ( in prod, there are currently 24682 records like this )

jwinston-bn: Should be able to assume it's the last one.

src/Aquifer.Common/Utilities/StringSimilarityUtilities.cs
kaseywright: Another reason these were marked as public was because this is a utility class that could be used differently than how the ResourceContentVersionSimilarityMessageSubscriber currently uses it. I'll refactor to make the methods private

kaseywright:
Bump @jwinston-bn @NateMerritt

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+
+    private async Task ProcessAsync(QueueMessage queueMessage, ScoreResourceContentVersionSimilarityMessage message, CancellationToken ct)
+    {
+        var similarityScoreEntity = await GenerateResourceContentVersionSimilarityScoreEntity(message, _dbContext, ct);
+        
+        await _dbContext.ResourceContentVersionSimilarityScores.AddAsync(similarityScoreEntity, ct);
+        await _dbContext.SaveChangesAsync(ct);
+    }
+    
+    private async Task<ResourceContentVersionSimilarityScoreEntity> GenerateResourceContentVersionSimilarityScoreEntity(
```
jwinston-bn, line 39: I generally would like to have the `Async` suffix on any async methods for consistency (like `ProcessAsync` above).

And personal preference, I wouldn't typically include types in method names where it can be implied by the return type. This is kind of like if the score were an int, and the method was `GenerateSimilarityScoreInt`. I would just call it `GenerateSimilarityScoreAsync`.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+    
+    private static async Task<ResourceContentVersionMachineTranslationEntity> GetMachineTranslationForSnapshot(
+        int versionId, 
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        return await dbContext
+                       .ResourceContentVersionMachineTranslations
+                       .AsTracking()
+                       .Join(
```
jwinston-bn, line 184: Would like to see if we can get this using the object references, rather than using this Join operator.

```csharp
x => x.ResourceContentVersion.ResourceContentVersionSnapshots
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
kaseywright: Makes perfect sense. `GenerateResourceContentVersionSimilarityScoreAsync` or get much shorter with it `GenerateSimilarityScoreAsync`?

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
kaseywright: I can give this another shot but there is an issue I was running into. We are given the snapshot id. So, to use the object references, we have to get to the machine translation from the snapshot. The issue is that snapshots have FKs to RCVs but there is no FK to machine translations _from_ RCVs. Rather, machine translations have the FK _to_ RCVs. So, I was not able to `Snapshot.RCV.Machine` or `Machine.RCV.Snapshot`. If that makes sense.
If LINQ can somehow bridge that gap without a join, I'm interested.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+                    machineTranslationEntity = await GetMachineTranslationForResourceContentVersion(
+                        message.BaseVersionId,
+                        dbContext,
+                        ct);
+                    
+                    similarityScoreEntity.ComparedVersionId = machineTranslationEntity.Id;
+                }
+                else
+                {
+                    similarityScoreEntity.ComparedVersionId = (int)message.CompareVersionId;
```
NateMerritt, line 66: Don't cast a nullable to non-nullable.  For `Nullable<T>` value types you can use the `Value` property instead.
```suggestion
                    similarityScoreEntity.ComparedVersionId = message.CompareVersionId.Value;
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+                    machineTranslationEntity = await GetMachineTranslationById(similarityScoreEntity.ComparedVersionId, dbContext, ct);
+                }
+                
+                compareVersionText = GetMachineTranslationVersionText(machineTranslationEntity.Content);
+                baseVersionText = await GetBaseVersionContentText(similarityScoreEntity.BaseVersionId, dbContext, ct);
+
+                _logger.LogInformation(
+                    "Scoring machine translation {CompareVersionId} vs published resource content version {BaseVersionId}...",
```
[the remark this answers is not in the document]
NateMerritt: Similar suggestion for the other logging below: Use the actual ID names as a log message parameter.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
NateMerritt: (same below)

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+                    similarityScoreEntity.ComparedVersionId);
+
+                break;
+            case ResourceContentVersionSimilarityComparisonType.SnapshotToSnapshot:
+                similarityScoreEntity.BaseVersionType = ResourceContentVersionTypes.Snapshot;
+                similarityScoreEntity.ComparedVersionType = ResourceContentVersionTypes.Snapshot;
+
+                similarityScoreEntity.ComparedVersionId = message.CompareVersionId 
+                                                          ?? throw new InvalidOperationException(
+                                                              $"Snapshot version {message.CompareVersionId} not found.");
```
NateMerritt, line 130:
```suggestion
                                                              $"{nameof(message.CompareVersionId)} (snapshot ID) is required for {nameof(ResourceContentVersionSimilarityComparisonType.SnapshotToSnapshot)} similarity scoring.");
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        return await dbContext
+                   .ResourceContentVersionMachineTranslations
+                   .AsTracking()
+                   .Where(x => x.ResourceContentVersionId == versionId)
+                   .OrderBy(x => x.Id)
+                   .LastOrDefaultAsync(ct) 
+               ?? throw new InvalidOperationException( $"No Machine translation for ResourceContentVersion with id {versionId} found");
```
NateMerritt, line 173:
```suggestion
               ?? throw new InvalidOperationException($"No Machine translation for ResourceContentVersion with id {versionId} found");
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+    {
+        return await dbContext
+                .ResourceContentVersionMachineTranslations
+                .AsTracking()
+                .FirstOrDefaultAsync(m => m.Id == machineTranslationId, ct)
+            ?? throw new InvalidOperationException($"Machine translation with id {machineTranslationId} not found");
+    }
+    
+    private static async Task<ResourceContentVersionMachineTranslationEntity> GetMachineTranslationForResourceContentVersion(
+        int versionId,
```
NateMerritt, line 163:
```suggestion
        int resourceContentVersionId, 
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+                   .ResourceContentVersionMachineTranslations
+                   .AsTracking()
+                   .Where(x => x.ResourceContentVersionId == versionId)
+                   .OrderBy(x => x.Id)
+                   .LastOrDefaultAsync(ct) 
+               ?? throw new InvalidOperationException( $"No Machine translation for ResourceContentVersion with id {versionId} found");
+    }
+    
+    private static async Task<ResourceContentVersionMachineTranslationEntity> GetMachineTranslationForSnapshot(
+        int versionId,
```
NateMerritt, line 177:
```suggestion
        int resourceContentVersionId, 
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+    {
+        var similarityScoreEntity = await GenerateResourceContentVersionSimilarityScoreEntity(message, _dbContext, ct);
+        
+        await _dbContext.ResourceContentVersionSimilarityScores.AddAsync(similarityScoreEntity, ct);
+        await _dbContext.SaveChangesAsync(ct);
+    }
+    
+    private async Task<ResourceContentVersionSimilarityScoreEntity> GenerateResourceContentVersionSimilarityScoreEntity(
+        ScoreResourceContentVersionSimilarityMessage message,
+        AquiferDbContext dbContext,
```
NateMerritt, line 41: I suggest using the class member's `_dbContent` instead of passing it here.  The same suggestion applies to the `private static` methods below doing the same thing but they'll need to be made non-static as well.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+                           rcvs => rcvs.ResourceContentVersionId,
+                           (rcvmt, rcvs) => new {rcvmt, rcvs}
+                       )
+                       .Where(joined => joined.rcvs.Id == versionId)
+                       .Select(x => x.rcvmt)
+                       .OrderBy(x => x.Id)  
+                       .LastOrDefaultAsync(ct) 
+                   ?? throw new InvalidOperationException(
+                       $"No Machine translation for ResourceContentVersionSnapshot with id {versionId} found"
+                   );
```
NateMerritt, line 196:
```suggestion
                       $"No Machine translation for ResourceContentVersionSnapshot with id {versionId} found");
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+    }
+    
+    private static async Task<ResourceContentVersionMachineTranslationEntity> GetMachineTranslationById(
+        int machineTranslationId,
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        return await dbContext
+                .ResourceContentVersionMachineTranslations
+                .AsTracking()
```
NateMerritt, line 157: It looks like none of the EF queries in this class need `AsTracking()` because none of the entities retrieved are used in any DB write operations.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+        
+        return HtmlUtilities.GetPlainText(
+            string.Join(string.Empty, contentHtmlItems)
+        );
+    }
+    
+    private static string GetMachineTranslationVersionText(string resourceHtmlStringContent)
+    {
+        return HtmlUtilities.GetPlainText(resourceHtmlStringContent);
+    }
```
NateMerritt, line 238: These comparisons will be incorrect for FIA content with multiple steps.  Each step is translated independently and has a separate machine translation DB entry (see [the translation code](https://github.com/BiblioNexusStudio/aquifer-server/blob/8645bb5f179ed4eb5a8ac1a5a1b68b7346e2efb3/src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs#L763C21-L763C34)).  When the `contentHtmlItems` only has a single entry then there will only be a single machine translation entry (with `ContentIndex` == `0`) but if it has multiple then there should be one machine translation record for each of them with `ContentIndex` matching the `contentHtmlItems` index.

Options when there are multiple steps:
1. Compare and score each step independently and save multiple DB records.
2. Fetch all of the machine translation records, join them together, and compare that with the joined `contentHtmlItems`, resulting in a single comparison/record.

@jwinston-bn, do you have an opinion?

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+                           .ResourceContentVersions
+                           .AsTracking()
+                           .SingleOrDefaultAsync(x => x.Id == resourceContentVersionId, ct)
+                       ?? throw new InvalidOperationException($"Content version with id {resourceContentVersionId} not found");
+        
+        var resourceContentHtmlItems = TiptapConverter.ConvertJsonToHtmlItems(resource.Content);
+        
+        return HtmlUtilities.GetPlainText(
+            string.Join(string.Empty, resourceContentHtmlItems)
+        );
```
NateMerritt, line 214: C# styling generally prefers parentheses to not be on their own line and instead be at the end of the preceding line.  I very rarely make exclusions to this rule but usually only for tuples where the parentheses are acting more like braces.

All that said, I'd just put this on a single line (but see my question below on if we should even be doing this string join):
```suggestion
        return HtmlUtilities.GetPlainText(string.Join(string.Empty, resourceContentHtmlItems));
```

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,239 @@
...
+    
+    private static async Task<ResourceContentVersionMachineTranslationEntity> GetMachineTranslationById(
+        int machineTranslationId,
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        return await dbContext
+                .ResourceContentVersionMachineTranslations
+                .AsTracking()
+                .FirstOrDefaultAsync(m => m.Id == machineTranslationId, ct)
```
NateMerritt, line 158: Below you used `SingleOrDefaultAsync`.  Any reason to not use that here as well when we know there should only be at most one match?

There are performance arguments for using `FirstOrDefaultAsync()` so my main concern is consistency.

kaseywright: Nope. Good catch. This was probably an autocomplete from Rider.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
kaseywright: Good catch. Whichever way we want this to be handled, I'll refactor for it.

jwinston-bn: The machine translation records have step number on them (looks like it's called ContentIndex), so I'd just grab all of them and do a single comparison. We don't need to get so granular where we have 6 scores for a single resource if we can help it.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,278 @@
...
+        string machineTranslationText;
+        
+        if (machineTranslations.Count == 1)
+        {
+            return (
+                machineTranslations.First().Id, 
+                HtmlUtilities.GetPlainText(machineTranslations.First().Content));
+        }
+
+        // We want to ensure that we get the correct set of MTs, whether it is 1 or many ( as in the case of FIA content )
```
kaseywright, line 241: I left some comments to guide the thinking behind the logic. Let me know if you find value in them or if I should just remove them.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,278 @@
...
+
+        if (machineTranslationId is null)
+        {
+            // Assume we are using the latest MT
+            contentList = machineTranslations
+                .Where(x => x.Id >= referenceMachineTranslation.Id)
+                .Select(x => x.Content)
+                .ToList();
+            
+            machineTranslationText = HtmlUtilities.GetPlainText(string.Join(string.Empty, contentList));
```
NateMerritt, line 257: It's possible that `contentList` could be an empty list here so I suggest throwing an exception instead of operating on an empty string.

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -0,0 +1,278 @@
...
+        if (machineTranslationId is not null && !machineTranslations.Any(m => m.Id != machineTranslationId))
+        {
+            throw new InvalidOperationException(
+                $"Machine translation with id {machineTranslationId} not found for snapshot version {resourceContentVersionSnapshotId}");
+        }
+        
+        return GetMachineTranslationIdAndText(machineTranslations, machineTranslationId);
+    }
+
+    private (int, string) GetMachineTranslationIdAndText(
```
NateMerritt, line 227: 👍 I reviewed this method and it looks like it will handle the FIA steps correctly.

NateMerritt (dismissed):

jwinston-bn (approved):

PR #621, opened 2024-12-27
Update `/resources/content/machine-translation/{Id}` sql 

kaseywright:
Update endpoint to allow users that have been assigned the resource at all to update machine translation scores.

This resolves the error messages when a user clicks the machine translation score and then the `Send to Review` too quickly.

Screenshot 2024-12-27 at 12 45 38 PM

### DIFF
--- a/src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
@@ -1,6 +1,5 @@
 ﻿using Aquifer.API.Services;
 using Aquifer.Data;
-using Aquifer.Data.Entities;
 using FastEndpoints;
 using Microsoft.EntityFrameworkCore;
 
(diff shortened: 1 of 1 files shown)


src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
```diff
@@ -16,38 +16,47 @@ public override void Configure()
     public override async Task HandleAsync(Request req, CancellationToken ct)
     {
         var user = await userService.GetUserFromJwtAsync(ct);
-        var existingMt = await dbContext.ResourceContentVersionMachineTranslations.AsTracking()
```
jwinston-bn, line 19: Can we just change `x.ResourceContentVersion.AssignedUserId == user.Id` to `x.ResourceContentVersion.ResourceContentVersionAssignedUserHistories.Any(h => h.AssignedUserId == user.Id)` without any other changes and it work?

I think the status check needs to go away too, because with the delay, it's presumably in CompanyReview when it fails.

jwinston-bn (changes requested):
I think can be simplified per comment

src/Aquifer.API/Endpoints/Resources/Content/MachineTranslation/Update/Endpoint.cs
kaseywright: Yes, that should work. It's a much more elegant approach to it. I'm still getting used to being able to access other tables in these queries like that.

PR #622, opened 2025-01-06
Reduce calls to 3rd party IP address service and handle rate limiting.

NateMerritt:
1. Only allow one simultaneous request to the IP address service for a given IP address.  This is accomplished with DB read row locking because Azure functions don't necessarily share the same memory caching instance so a `MemoryCache`/`HybridCache` won't work and the only other way to do distributed locking would be to stand up a distributed redis cache instance.  Using a DB transaction isn't the most ideal way to lock because the transaction is held open during the duration of the IP address service lookup (including Polly retries with delays) but being that the lock is for a single row (including an empty row) it should be acceptable.
2. Retry with Polly on 429 (Too Many Requests) responses with a delay.
3. Record IP lookup data _even if_ it's null.  Right now we don't record it which means we'll keep looking that IP up indefinitely which causes us to hit limits.

### DIFF
--- a/src/Aquifer.Data/Entities/IpAddressData.cs
+++ b/src/Aquifer.Data/Entities/IpAddressData.cs
@@ -8,13 +8,13 @@ public class IpAddressData
     public string IpAddress { get; set; } = null!;
 
     [MaxLength(256)]
-    public string City { get; set; } = null!;
+    public string? City { get; set; }
 
     [MaxLength(256)]
-    public string Region { get; set; } = null!;
+    public string? Region { get; set; }
 
     [MaxLength(256)]
-    public string Country { get; set; } = null!;
+    public string? Country { get; set; }
 
     [SqlDefaultValue("getutcdate()")]
     public DateTime Created { get; set; } = DateTime.UtcNow;
--- a/src/Aquifer.Jobs/Program.cs
+++ b/src/Aquifer.Jobs/Program.cs
@@ -1,3 +1,4 @@
+using System.Net;
 using Aquifer.AI;
 using Aquifer.Common.Clients;
 using Aquifer.Common.Clients.Http.IpAddressLookup;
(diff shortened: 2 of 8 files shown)


src/Aquifer.Jobs/Program.cs
```diff
@@ -54,7 +55,8 @@
         services.AddSingleton<IAquiferApiManagementClient, AquiferApiManagementClient>();
         services.AddHttpClient<IIpAddressLookupHttpClient, IpAddressLookupHttpClient>()
             .AddPolicyHandler(HttpPolicyExtensions.HandleTransientHttpError()
-                .WaitAndRetryAsync(Backoff.DecorrelatedJitterBackoffV2(TimeSpan.FromSeconds(0.5), 2)));
+                .OrResult(res => res.StatusCode == HttpStatusCode.TooManyRequests)
+                .WaitAndRetryAsync(Backoff.DecorrelatedJitterBackoffV2(TimeSpan.FromSeconds(5), 2)));
```
NateMerritt, line 59: The only thing calling the IP service is an Azure function so a longer median wait timeout of 5 seconds seemed reasonable.

src/Aquifer.Jobs/Subscribers/TrackResourceContentRequestMessageSubscriber.cs
```diff
@@ -77,19 +96,11 @@ private async Task LookupIpAddressAsync(TrackResourceContentRequestMessage track
                         Country = ipData.Country
                     };
 
-                    try
-                    {
-                        await dbContext.IpAddressData.AddAsync(newRecord, ct);
-                        await dbContext.SaveChangesAsync(ct);
-                    }
-                    catch (DbUpdateException)
```
NateMerritt, line 85: It should no longer be possible to have duplicate key exceptions with this change.

src/Aquifer.Data/Entities/IpAddressData.cs
jwinston-bn, line 1: This file should have `Entity` suffix. Can you add as part of this please. Shouldn't need any migrations run since it won't affect the table name itself.

NateMerritt: Added.

jwinston-bn (approved):

PR #623, opened 2025-01-06
Log if HTML is invalid before/after translation.

NateMerritt:


src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -727,9 +727,33 @@ private async Task TranslateResourceCoreAsync(
...
+
+                // validate source HTML and log if there are any issues
+                var sourceHtmlValidationErrors = HtmlUtilities.GetHtmlValidationErrors(contentHtmlItem);
+                if (sourceHtmlValidationErrors.Count > 0)
+                {
+                    _logger.LogError(
+                        "HTML validation errors found in source content for Resource Content ID {ResourceContentId} and Resource Content Version ID {ResourceContentVersionId} during translation. Source HTML Validation errors: {HtmlValidationErrors}.",
+                        resourceContentId,
+                        resourceContentVersion.Id,
+                        sourceHtmlValidationErrors);
```
jwinston-bn, line 739: Did you verify this doesn't just get `ToString` or something called on it? It actually lists out all the errors?

PR #629, opened 2025-01-14
Enable versification to work across ranges, apply exclusions, and support base verse parts in mappings.

NateMerritt:
This PR includes the addition of integration tests for Aquifer.Common with `appsettings.json` and DI support (the required environment variables for GHA are already present).  There's an available NuGet package to provide full test class constructor DI with XUnit but it also provided a lot of other functionality I didn't want, had some abstractions that made you learn its specific strategies instead of standard `Builder` extension methods, and didn't do exactly what I wanted for DI, so I rolled my own app fixture which requires manually getting a required service.  I think it's pretty clean for what it does but welcome other opinions.

### DIFF
--- a/src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
+++ b/src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
@@ -8,9 +8,9 @@ namespace Aquifer.Common.Services.Caching;
 
 public interface ICachingVersificationService
 {
-    Task<ReadOnlyDictionary<int, int>> GetBaseVerseIdByBibleVerseIdMapAsync(int bibleId, CancellationToken cancellationToken);
+    Task<ReadOnlyDictionary<int, string>> GetBaseVerseIdWithOptionalPartByBibleVerseIdMapAsync(int bibleId, CancellationToken cancellationToken);
 
-    Task<ReadOnlyDictionary<int, int>> GetBibleVerseIdByBaseVerseIdMapAsync(int bibleId, CancellationToken cancellationToken);
+    Task<ReadOnlyDictionary<string, int>> GetBibleVerseIdByBaseVerseIdWithOptionalPartMapAsync(int bibleId, CancellationToken cancellationToken);
 
     Task<ReadOnlySet<int>> GetExcludedVerseIdsAsync(int bibleId, CancellationToken cancellationToken);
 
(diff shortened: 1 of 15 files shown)


src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
```diff
@@ -1,108 +1,216 @@
...
+using Aquifer.Data.Enums;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Caching.Memory;
 
 namespace Aquifer.Common.Services.Caching;
 
 public interface ICachingVersificationService
 {
-    Task<ReadOnlyDictionary<int, int>> GetBaseVerseIdByBibleVerseIdMapAsync(int bibleId, CancellationToken cancellationToken);
+    Task<ReadOnlyDictionary<int, string>> GetBaseVerseIdWithOptionalPartByBibleVerseIdMapAsync(int bibleId, CancellationToken cancellationToken);
```
NateMerritt, line 11: Updated to `string` to support base verse parts like `1002003002a` as a value (or key for the inverted dictionary).

jwinston-bn (approved):

src/Aquifer.Common/Utilities/DictionaryExtensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Aquifer.Common.Utilities;
+
+public static class DictionaryExtensions
+{
+    public static TValue? GetValueOrNull<TKey, TValue>(this IDictionary<TKey, TValue> source, TKey key)
```
NateMerritt, line 5: `GetValueOrDefault()` doesn't support returning `null` for a `Dictionary<int, int>` so I wrote this extension method to allow null chaining.

src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
```diff
@@ -1,108 +1,216 @@
...
+                            grp.Key.ChapterNumber,
+                            MaxVerseNumber = grp.Max(x => x.VerseNumber),
+                        })
+                        .ToListAsync(cancellationToken))
+                    .GroupBy(x => x.BookId)
+                    .ToDictionary(
+                        grp => grp.Key,
+                        grp =>
+                        (
+                            MaxChapterNumber: grp.Max(x => x.ChapterNumber),
```
NateMerritt, line 117: There's an argument that the max chapter number should come from `BibleBookContents` but being that I have to do this grouping to get the max verse number, anyway, I decided to pull it from `BibleTexts` in this single query instead.

src/Aquifer.Common/Utilities/VersificationUtilities.cs
```diff
@@ -1,19 +1,258 @@
...
+    /// <c>null</c> will be returned if it's not possible to map between the versification schemes.
+    /// </summary>
+    public static async Task<int?> ConvertVersificationAsync(
+        int sourceBibleId,
+        int sourceVerseId,
+        int targetBibleId,
+        ICachingVersificationService versificationService,
+        CancellationToken ct)
+    {
+        if (!await IsValidVerseIdAsync(sourceBibleId, sourceVerseId, versificationService, ct))
```
NateMerritt, line 162: I decided to have these utilities throw an exception if the provided source verse IDs are not in the given Bible.  We could change this if desired but it seems to me that any caller should already be doing the right thing regarding verse IDs.

We'll also have the `VersificationUtilities.IsValidVerseIdAsync()` method available to any endpoints that want to validate  API consumer input.

src/Aquifer.Common/Utilities/VersificationUtilities.cs
```diff
@@ -1,19 +1,258 @@
...
+        // 3. Default to the base verse ID without a part (no explicit mapping).
+        // Rules when a part is not present on base verse ID:
+        // 1. Use the base verse ID without a part if present in the target inverse mapping (the direct map).
+        // 2. If not found, then try to map using the base verse ID with part 'a' (in case the target mapping uses parts).
+        // 3. Default to the base verse ID without a part (no explicit mapping).
+        var targetVerseId = targetBibleVerseIdByBaseVerseIdWithOptionalPartMap.GetValueOrNull(baseVerseIdWithOptionalPart)
+               ?? targetBibleVerseIdByBaseVerseIdWithOptionalPartMap.GetValueOrNull(
+                   baseVerseIdWithOptionalPart.Length > 10
+                       ? baseVerseId
+                       : $"{baseVerseId}a")
```
NateMerritt, line 212: Interestingly, there are multiple mappings where the first `baseVerseIdPart` is `b`.  In those cases this lookup will miss that first mapping (because we append `a`) and the lookup will then miss any explicit mapping and use the default of `value = key`.  Looking at the data where `b` is the only mapping this logic still seems like the best move (and I added integration tests for a couple of them).

tests/Aquifer.Common.IntegrationTests/Utilities/VersificationUtilitiesTests.cs
```diff
@@ -0,0 +1,133 @@
+﻿using Aquifer.Common.Services.Caching;
+using Aquifer.Common.Utilities;
+
+namespace Aquifer.Common.IntegrationTests.Utilities;
+
+public sealed class VersificationUtilitiesTests(App _app) : TestBase<App>
+{
+    private readonly ICachingVersificationService _versificationService = _app.GetRequiredService<ICachingVersificationService>();
```
NateMerritt, line 8: Per the PR description, this amount of manual DI is acceptable to me because getting class constructor level DI injection with third-party packages isn't worth it.

tests/Aquifer.Common.IntegrationTests/Utilities/VersificationUtilitiesTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    private const int _engVersificationSchemeBibleId = 0;
+    private const int _bsbBibleId = 1;
+    private const int _ls1910BibleId = 6;
+    private const int _rsbBibleId = 9;
+    private const int _gltBibleId = 10;
+
+    [Theory]
```
NateMerritt, line 16: For clarity, these integration tests are all hitting the Dev/QA DB, loading actual Bible versifications, exclusions, and max chapter/verse numbers, and using the actual `CachingVersificationService` code.

jwinston-bn (approved):

PR #632, opened 2025-01-16
Allow for graceful succeed when no Machine Translation exists

kaseywright:
There should not be an error thrown when a machine translation does not exist for a Resource Content Version or Snapshot when an attempt was made to score similarity. There are many resources that are in a working state that are in this situation. Instead, we should log a warning and allow the graceful exit of the scoring job.

### DIFF
--- a/src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
+++ b/src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
@@ -50,6 +50,8 @@ public sealed class ResourceContentVersionSimilarityMessageSubscriber(
         var similarityScoreEntity = new ResourceContentVersionSimilarityScoreEntity { BaseVersionId = message.BaseVersionId };
         string baseVersionText;
         string compareVersionText;
+        int? machineTranslationId;
+        string? machineTranslationText;
         IReadOnlyList<string> resourceContentHtmlItems;
         IReadOnlyList<string> rcvSnapshotHtmlItems;
         ResourceContentVersionEntity resourceContentVersion;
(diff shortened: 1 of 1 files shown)


jwinston-bn (approved):

src/Aquifer.Jobs/Subscribers/ResourceContentVersionSimilarityMessageSubscriber.cs
```diff
@@ -186,8 +201,11 @@ private async Task<ResourceContentVersionSnapshotEntity> GetSnapshotVersionAsync
 
         if (machineTranslations.Count == 0)
         {
-            throw new InvalidOperationException(
-                $"No Machine translation for ResourceContentVersion with id {resourceContentVersionId} found");
+            _logger.LogWarning(
+                "No Machine translation for ResourceContentVersion with id {ResourceContentVersionId} found. Skipping similarity score generation.",
+                resourceContentVersionId);
+            
+            return (-1, string.Empty);
```
NateMerritt, line 208: Sentinel values like `-1` make code harder to understand and consume because callers of this method need to read the method code to understand what sentinel are returned (or we need good method documentation).  Therefore, I suggest one of the following:
1. Return a `null` tuple for this state (with `private async Task<(int, string)>`) and callers will be forced to handle the `null` state by the compiler, or
2. Rename the method `Try...Async` and return a `bool IsSuccess` value as part of the tuple returned.  Then callers can check that `IsSuccess` value and this method can assign `default` to the other return values when `IsSuccess` is `false`.

kaseywright: So, something like this where the returned tuple is `(null, null)`

```
var (result, text) = await GetMachineTranslationIdAndTextForResourceContentVersionAsync(
      similarityScoreEntity.BaseVersionId,
      message.CompareVersionId,
      ct);

  if (result is null || text == null)
  {
      return null;
  }

  similarityScoreEntity.ComparedVersionId = result.Value; 
  compareVersionText = text;
```
With better variable names ofc

NateMerritt: I should have clarified.  You can return a `null` tuple itself.  In this case:
```
private async Task<(int, string)?> GetMachineTranslationIdAndTextForResourceContentVersionAsync
```
It's also best practice to include tuple property names in return value tuples:
```
private async Task<(int MachineTranslationId, string Text)?> GetMachineTranslationIdAndTextForResourceContentVersionAsync
```

kaseywright: Thanks for the clarification!

PR #635, opened 2025-01-27
Refactor ResourceContentSearch into new service and support filtering on audio and unresolved comments.

NateMerritt:
This took me a couple of rewrites in order to get acceptable performance.  From some ad-hoc testing the new code seems about the same with regards to performance as the old code while also adding the required new search filtering.

### Philosophy

* Build out one (complicated) resource content search service that is easy to consume and hides complexity.
* Remove some of the existing joins and fetch them independently when warranted (e.g. users and languages).
* For expensive operations, callers must explicitly request sub-DTOs on the search result class and they will only be populated if requested.
* Improve total count calculations by not running the total query if there is only one page.
* For the new audio and resolved comment filtering, use `CROSS APPLY` and `OUTER APPLY` instead of `JOIN` and `OUTER JOIN`, respectively.  This dramatically improved the new filtering performance.
* JOIN on a single `ResourceContentVersions` whenever possible (e.g. when doing `IsDraft` and `IsPublished` filtering) because it is faster than aggregating all `ResourceContentVersions` for a `ResourceContent`.  This complicated the code but made the dashboards must faster (they were previously doing this).
* If not possible, use some data rules to fetch less data (e.g. if the `ResourceContent` has any `ResourceContentVersions` in the `IsDraft` status then that `ResourceContentVersion` must be the latest one).
* Comment threads are only examined if the latest `ResourceContentVersion` is a draft.  Thus, old comment threads for previous versions are ignored.
* Switch to using Dapper and use its parameters bag and a multi-object query (this also had the benefit of fixing SQL injection attack vectors).

This PR will also unlock the ability to have server-side paging for the dashboard tabs.  However, it doesn't not yet support filtering by `LastAssigned` (required by "My Work" and "Manage") or sorting by `LastAssigned` (required by "My Work") which is currently not part of the search query (it would make it much more expensive in terms of performance).  If we need that in the back-end search then I propose adding an `int? LastAssignedUserId` to the `ResourceContentVersion` table.

### Performance

Regarding performance, I'd like to see this tested for users with many items on their dashboards.  The resource search performance is acceptable but if you remove the verse range filtering (currently required in the CMS) it is very slow.  For example, filtering only to English resource contents (as the only filter) takes > 2 seconds and filtering to only published resources contents (as the only filter) is > 4 seconds on my local.

**Ideas for improving performance further:**

1. Add summary columns at the `ResourceContent` level:
* Create summary copies of the `ResourceContentVersion`s `IsDraft` and `IsPublished` at the parent `ResourceContent` level.
* Move (or just copy if that's hard) `ResourceContentVersion`s `AssignedUserId` to the parent `ResourceContent` level.
* Add summary `HasAudio` at the `ResourceContent` level.
* Add summary `HasUnresolvedComments` at the `ResourceContent` level.

The pain with those ideas is that we have to keep the summary columns always up-to-date but it might still be worth it.

2. Use indexed views.

Create a new `ResourceContentSummaries` [indexed view](https://learn.microsoft.com/en-us/sql/relational-databases/views/views?view=sql-server-ver15#indexed-views) (or possibly multiple indexed views) that fetches all needed data and "caches it" in a view.  The pain with that idea is that any insert/update to component tables triggers a view refresh.  Splitting into multiple views and joining them together in a caller's query might make that more palatable.

3. Move searching to in-memory.

* Load in all needed `ResourceContent` and associated data on a timed cache.
* Refresh the cache every 15 min.
* Search within memory.
* Can get better fuzzy search.

The pain with this idea is that searches will hit stale data and be incorrect.

4. Use Azure's search solution.

* Requires a loader job for full loads.
* If we want close to real-time search accuracy, also requires a subscriber for incremental updates.
* More money.
* Can get better fuzzy search.
* Could also implement search within resource content text.

Pain: Cost both in terms of implementation time and maintenance time.


.editorconfig
```diff
@@ -271,8 +277,8 @@ dotnet_style_null_propagation = true:error
 dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
 dotnet_style_prefer_auto_properties = true
 dotnet_style_object_initializer = true:suggestion
-dotnet_style_collection_initializer = true
+dotnet_style_collection_initializer = true:suggestion
```
NateMerritt, line 280: It looks like I duplicated things after accepting Visual Studio's suggestions in a previous commit.  I'll address this in another commit later.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
```diff
@@ -1,101 +1,121 @@
...
+                    ResourceContentStatus.TranslationAiDraftComplete,
+                    ResourceContentStatus.TranslationCompanyReview,
+                    ResourceContentStatus.TranslationEditorReview,
+                ],
+                HasAudio = req.HasAudio,
+                HasUnresolvedCommentThreads = req.HasUnresolvedCommentThreads,
+            },
+            ResourceContentSearchSortOrder.ProjectProjectedDeliveryDate,
+            offset: 0,
+            limit: null,
```
NateMerritt, line 55: This will fetch all results in a single page to keep current behavior.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToSelf/List/Endpoint.cs
```diff
@@ -1,90 +1,133 @@
...
     public override void Configure()
     {
         Get("/resources/content/assigned-to-self");
         Permissions(PermissionName.ReadResources);
     }
 
-    public override async Task HandleAsync(CancellationToken ct)
+    /// ---------------------------------------------------------------------------------------------------
+    /// ---------------------------------------------------------------------------------------------------
+    /// IMPORTANT: WHEN THIS SEARCH FILTER IS UPDATED THE /next-up ENDPOINT NEEDS TO BE UPDATED ACCORDINGLY
```
NateMerritt, line 29: I did not update the `next-up` endpoint because the results returned from the search service should match what was previously returned for this endpoint.

src/Aquifer.API/Endpoints/Resources/Content/Get/Response.cs
```diff
@@ -43,7 +43,8 @@ public class Response
                 IsComplete = ProjectEntity.ActualPublishDate is not null
             };
 
-    public CommentThreadsResponse CommentThreads { get; set; } = null!;
+    public bool HasUnresolvedCommentThreads => CommentThreads?.Threads.Any(t => !t.Resolved) ?? false;
```
NateMerritt, line 46: I did not add `HasAudio` to the `GET /resources/content/{id}` route because it would make the route slower by adding an extra DB query and I'm not sure if it would even be used.  I can add if needed.

src/Aquifer.Common/Services/Caching/CachingLanguageService.cs
```diff
@@ -69,13 +85,19 @@ public async Task<IReadOnlyDictionary<string, int>> GetLanguageIdByCodeMapAsync(
...
                 cacheEntry.SlidingExpiration = s_cacheLifetime;
 
-                var LanguageCodeById = await _dbContext.Languages
-                    .Select(l => new { l.Id, l.ISO6393Code })
-                    .ToDictionaryAsync(l => l.Id, l => l.ISO6393Code, ct);
+                var languageEntityByIdMap = (await _dbContext.Languages
+                        .ToDictionaryAsync(l => l.Id, l => l, ct))
+                    .AsReadOnly();
+
+                var languageCodeByIdMap = languageEntityByIdMap
```
NateMerritt, line 92: I don't think there should be any issues with caching entities in memory because we're not tracking them but I wanted to call it out here in case something comes up.

src/Aquifer.Data/Aquifer.Data.csproj
```diff
@@ -6,6 +6,7 @@
 
     <ItemGroup>
         <PackageReference Include="Dapper" />
+        <PackageReference Include="Microsoft.Data.SqlClient" />
```
NateMerritt, line 9: This was a transient dependency of EF that I made explicit in order to update to the latest version.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -0,0 +1,672 @@
+﻿using Aquifer.Data;
+using Aquifer.Data.Entities;
+using Dapper;
+using Microsoft.EntityFrameworkCore;
+using static Aquifer.Common.Services.ResourceContentSearchResult;
+
+namespace Aquifer.Common.Services;
+
+public interface IResourceContentSearchService
```
NateMerritt, line 9: If it would be easier for review then I could do a pairing session to walk through this class.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -0,0 +1,672 @@
...
+
+    All = Project | ResourceContentVersions | HasAudioForLanguage | HasUnresolvedCommentThreads,
+}
+
+public enum ResourceContentSearchSortOrder
+{
+    ResourceContentId = 0,
+
+    /// <summary>
+    /// Sort: ParentResource.DisplayName, Resource.SortOrder, Resource.EnglishLabel, ResourceContent.LanguageId
```
NateMerritt, line 62: I added `LanguageId` here and below for consistent ordering results when multiple languages are present with the same `EnglishLabel`.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -0,0 +1,672 @@
...
+        var whereSql = whereClausesSql.Count > 0
+            ? $"WHERE{Environment.NewLine}    {string.Join($" AND{Environment.NewLine}    ", whereClausesSql)}"
+            : "";
+
+        var hasPaging = offset != 0 || limit != null;
+
+        const string offsetLiteralName = "offset";
+        const string limitLiteralName = "limit";
+        const string pagingSql = $$"""
+           OFFSET {={{offsetLiteralName}}} ROWS FETCH NEXT {={{limitLiteralName}}} ROWS ONLY
```
NateMerritt, line 588: Dapper allows injecting literals.  This won't use SQL parameters but still allows populating the parameters object with the literal values.  I like this methodology because it makes it clear that you aren't using SQL parameters on purpose and you know there isn't a risk of SQL injection attacks (Dapper only supports safe values for literals).

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -0,0 +1,672 @@
...
+        var dataSql = $"""
+            {selectPropertiesSql}
+            {fromSql}
+            {whereSql}
+            {sortSql}
+            {(hasPaging ? pagingSql : "")}
+            """;
+
+        var resourceContentSearchResults = (await dbContext.Database.GetDbConnection()
+            .QueryAsync<
```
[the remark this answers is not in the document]
NateMerritt: Note to self: I forgot to add `WithRetries` support for this.

src/Aquifer.API/Endpoints/Resources/Content/Get/Response.cs
kaseywright: For [Bib-834](https://linear.app/biblionexus/issue/BIB-834/showplay-audio-on-aquifer-resource-pages) we will need to have the audio information for the resource being viewed/edited. We can:
- send down the audio information in this response or
- only send a flag and then making a separate request for the audio information
Either way should be fine as we have plenty of time on the front end to make a separate request. Whatever works on the server to keep things performant.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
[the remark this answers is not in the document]
kaseywright: Does `WithRetries` need to be in place before approval and merging?

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
```diff
@@ -1,101 +1,121 @@
...
+                .Select(x => x.Count > 1 ? x[1].UserId : null)
+                .OfType<int>())
+            .Distinct()
+            .ToList();
 
-        var queryResponse = await dbContext.Database.SqlQueryRaw<SqlQueryResult>(query).ToListAsync(ct);
-        var resourceContentVersionIds = queryResponse.Select(x => x.ResourceContentVersionId);
-        var lastAssignments =
-            await Helpers.GetLastAssignmentsAsync(resourceContentVersionIds, dbContext, ct);
+        var userDtoByIdMap = await _dbContext.Users
```
jwinston-bn, line 78: At this point are we guaranteed not to have any users that are `Enabled=0`?

src/Aquifer.API/Endpoints/Resources/Content/AssignedToSelf/List/Endpoint.cs
```diff
@@ -1,90 +1,133 @@
...
+                return new Response
+                {
+                    Id = rcs.ResourceContent.Id,
+                    EnglishLabel = rcs.Resource.EnglishLabel,
+                    ParentResourceName = rcs.ParentResource.DisplayName,
+                    LanguageEnglishDisplay = languageEntityByIdMap[rcs.ResourceContent.LanguageId].EnglishDisplay,
+                    WordCount = resourceContentVersion.SourceWordCount,
+                    StatusValue = rcs.ResourceContent.Status,
+                    Status = rcs.ResourceContent.Status.GetDisplayName(),
+                    StatusDisplayName = rcs.ResourceContent.Status.GetDisplayName(),
```
jwinston-bn, line 107: Should check if both of these are even still used. I think one of them is legacy, so might be able to remove one.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
```diff
@@ -1,101 +1,121 @@
...
-                Name = $"{x.UserFirstName} {x.UserLastName}"
-            },
-            DaysSinceContentUpdated = x.ContentUpdated == null ? null : (DateTime.UtcNow - (DateTime)x.ContentUpdated).Days,
-            DaysUntilProjectDeadline =
-                    x.ProjectProjectedDeliveryDate == null
+                // Project should never be null because we filtered to IsInProject = true
+                var project = rcs.Project!;
+                var resourceContentVersion = rcs.ResourceContentVersion!;
+
+                var lastTwoUserAssignments = lastUserAssignmentsByResourceContentVersionIdMap[resourceContentVersion.Id];
```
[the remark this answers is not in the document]
jwinston-bn: For me that query returns `Matthieu 4:12-25`.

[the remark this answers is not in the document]
elisabeth-shakespeare: Did you hit the endpoint with the manager role?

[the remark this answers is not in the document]
elisabeth-shakespeare: Also, you need to have the same company as resource content `44893` in your own

src/Aquifer.Common/Services/ResourceContentSearchService.cs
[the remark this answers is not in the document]
NateMerritt: No, but I just added it.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
NateMerritt: No, this follows the previous logic [here](https://github.com/BiblioNexusStudio/aquifer-server/blob/48bea35b907658d912679a698169bccbc3c9aa7a/src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs#L47) and [here](https://github.com/BiblioNexusStudio/aquifer-server/blob/48bea35b907658d912679a698169bccbc3c9aa7a/src/Aquifer.API/Endpoints/Resources/Content/Helpers.cs#L120-L146) which didn't check for `Enabled`.  Does it need to care about that?

src/Aquifer.API/Endpoints/Resources/Content/AssignedToSelf/List/Endpoint.cs
```diff
@@ -1,90 +1,133 @@
...
+                var resourceContentVersion = rcs.ResourceContentVersion!;
+
+                // The most recent user assignment history should be to assign to the current user.
+                var lastTwoUserAssignments =
+                    lastUserAssignmentsByResourceContentVersionIdMap[resourceContentVersion.Id];
+                var currentUserAssignment = lastTwoUserAssignments[0];
+                if (currentUserAssignment.UserId != user.Id)
+                {
+                    throw new InvalidOperationException(
+                        $"Expected the currently assigned user to be the last assigned user, but the current user ID is \"{user.Id}\" and the last assigned user is \"{currentUserAssignment.UserId}\".");
```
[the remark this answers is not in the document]
NateMerritt: Good call.  I updated to log and error and gracefully continue.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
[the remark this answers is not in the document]
NateMerritt: That ResourceContentVersion exists in QA but there is no assignment history for it.  I [updated the PR](https://github.com/BiblioNexusStudio/aquifer-server/pull/635/commits/f4bddeea847ad608f95c84c74f9728ba7d74775a) to no longer assume that there will always be assignment history, even when the resource content version is currently assigned.

src/Aquifer.Common/Services/Caching/CachingLanguageService.cs
NateMerritt: I updated to no longer cache the language entity and use a new record type, intsead.  After letting the code settle for a bit, I think it's best to never cache EF models as a general rule.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
jwinston-bn: I guess in this particular case it's okay.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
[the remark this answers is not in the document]
jwinston-bn: Sometimes we move things around in SQL and don't always add the assignment history. Definitely a problem with the person running the scripts, but it nonetheless has happened.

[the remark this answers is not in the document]
elisabeth-shakespeare: I've pulled down latest and run with cms. Working now 👍

elisabeth-shakespeare (approved):

src/Aquifer.API/Endpoints/Resources/Content/Get/Response.cs
NateMerritt: @kaseywright, I added both `HasAudio` (for consistency with the other routes) and `AudioResources`, which includes the audio content IDs for a follow-up fetch request.

elisabeth-shakespeare (approved):

PR #636, opened 2025-02-04
Bib 850 update manage dashboard columns

elisabeth-shakespeare:


src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -361,7 +363,9 @@ GROUP BY rcv.ResourceContentId
                 """
                 : "")}
             {(filter.AssignedUserCompanyId.HasValue
-                ? "    LEFT JOIN Users u ON rcv.AssignedUserId = u.Id"
```
NateMerritt, line 364: This change is now unneeded.

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
```diff
@@ -72,6 +72,7 @@ await Helpers.GetLastUserAssignmentsByResourceContentVersionIdMapAsync(
             .Concat(lastUserAssignmentsByResourceContentVersionIdMap.Values
                 .Select(x => x.Count > 1 ? x[1].UserId : null)
                 .OfType<int>())
+            .Concat(resourceContentSummaries.Select(rcs => rcs.ResourceContentVersion!.AssignedReviewerUserId).OfType<int>())
```
jwinston-bn, line 75: Why do we need `OfType<int>` on here?

NateMerritt: It filters out `null` values.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
elisabeth-shakespeare: We don't need to filter on CompanyId at all?

src/Aquifer.API/Endpoints/Resources/Content/AssignedToOwnCompany/List/Endpoint.cs
elisabeth-shakespeare: Screenshot 2025-02-04 at 10 24 39 AM

... and it takes out `null`s

src/Aquifer.Common/Services/ResourceContentSearchService.cs
NateMerritt: The original code is still needed.  But the whitespace changes introduced in this commit from a standard string to a raw literal string are not needed.

NateMerritt (approved):

PR #640, opened 2025-02-07
Support updating a user's role.

NateMerritt:
* Instead of adding more manual request code to Auth0, I swapped out the manual Auth0 `HttpClient` code and used their first-party NuGet packages.  I tested that the create user and disable user flows still work correctly.  This removed some extra error handling code, prevented me from having to duplicate request handling code again, and provided baked-in retries for free.
* Two new permissions have been added in dev and prod: `update:user` (assigned to the `Admin` and `Publisher` roles) and `update:users-in-company` (assigned to the `Manager` role).
* The validation code for the new role update route mirrors the validation used for the pre-existing create user endpoint.


src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Users.Role.Set;
+
+public class Endpoint(AquiferDbContext _dbContext, IAuth0Service _authService, IUserService _userService, ILogger<Endpoint> _logger)
+    : Endpoint<Request>
+{
+    public override void Configure()
+    {
+        Post("/users/{UserId}/role");
```
NateMerritt, line 15: I'm not sold on this endpoint URL but it seemed like the best choice.  This route only sets the user's role.  This could instead be a full "update user" route but where for now we only allow updating the role; perhaps we'd want to allow other edits later (like user name).  I also used POST instead of PUT/PATCH and there's no equivalent GET route (I wrote one initially but realized the GETting the user includes the role in the response so it was redundant).

So all that said, this isn't very RESTful but for just setting the role it seems like a reasonable choice.

src/Aquifer.API/Services/Auth0Service.cs
```diff
@@ -0,0 +1,165 @@
...
+using Auth0.AuthenticationApi.Models;
+using Auth0.ManagementApi;
+using Auth0.ManagementApi.Models;
+using Auth0.ManagementApi.Paging;
+
+namespace Aquifer.API.Services;
+
+public interface IAuth0Service
+{
+    Task<string> GetAccessTokenAsync(CancellationToken ct);
```
NateMerritt, line 15: I thought about building the fetching of the access token into the scoped service so that it would happen automatically for callers and they wouldn't have to pass it around but decided that could be a later refactor if desired.

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
```diff
@@ -0,0 +1,78 @@
...
+        var auth0Roles = await _authService.GetUsersRolesAsync(accessToken, user.ProviderId, ct);
+        foreach (var auth0Role in auth0Roles)
+        {
+            if (!Enum.TryParse(auth0Role.Name, ignoreCase: true, out UserRole _))
+            {
+                throw new InvalidOperationException($"Auth0 Role not found in {nameof(UserRole)} enum: \"{auth0Role.Name}\".");
+            }
+        }
+
+        if (auth0Roles.Count > 1)
```
NateMerritt, line 62: @jwinston-bn, is it ever expected that a user would have more than one role?  I assume not and remove all of a user's roles as part of the role assignment here.

src/Aquifer.API/Endpoints/Users/Create/Endpoint.cs
```diff
@@ -53,70 +51,27 @@ private async Task ValidateCompanyIdAsync(int companyId, CancellationToken ct)
...
-            logger.LogWarning("Requested non-existent role: {requestedRole} - {response}", req.Role, responseContent);
+            logger.LogWarning("Requested non-existent role: {requestedRole}.", req.Role);
             ThrowError("Requested role does not exist", 400);
         }
 
-        return role.Id;
+        return roleId;
     }
 
     private async Task<string> CreateAuth0UserAsync(Request req, string accessToken, CancellationToken ct)
```
jwinston-bn, line 67: This method seems redundant now.

src/Aquifer.API/Endpoints/Users/Create/Endpoint.cs
```diff
@@ -53,70 +51,27 @@ private async Task ValidateCompanyIdAsync(int companyId, CancellationToken ct)
...
-        {
-            logger.LogWarning("Error creating Auth0 user: {statusCode} - {response}", response.StatusCode, responseContent);
-            ThrowError(responseContent, (int)response.StatusCode);
-        }
-
-        return JsonUtilities.DefaultDeserialize<Auth0CreateUserResponse>(responseContent).UserId;
+        return await authService.CreateUserAsync(accessToken, req.Email, $"{req.FirstName} {req.LastName}", ct);
     }
 
     private async Task AssignAuth0RoleAsync(string accessToken, string userId, string roleId, CancellationToken ct)
```
jwinston-bn, line 72: This one too

src/Aquifer.API/Endpoints/Users/Create/Endpoint.cs
```diff
@@ -53,70 +51,27 @@ private async Task ValidateCompanyIdAsync(int companyId, CancellationToken ct)
...
-                        Note that the Auth0 user has been created and recalling this
-                        endpoint will result in different errors. Please ask for help.
-                        {responseContent}
-                        """,
-                (int)response.StatusCode);
-        }
+        await authService.AssignRoleToUserAsync(accessToken, userId, roleId, ct);
     }
 
     private async Task SendPasswordResetAsync(string accessToken, string email, CancellationToken ct)
```
jwinston-bn, line 77: And so on

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
jwinston-bn: I'd probably lean more toward a PATCH on `/users/{UserId}` since I could see name changing or whatever else being a thing at some point. So I personally would just change the route and nothing else, and then if we need more in the future can add to it.

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
jwinston-bn: No one currently has two roles. Auth0 would support it, but our Users table doesn't.

src/Aquifer.API/Endpoints/Users/Role/Set/Request.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Aquifer.Data.Entities;
+
+namespace Aquifer.API.Endpoints.Users.Role.Set;
+
+public sealed class Request
+{
+    public required int UserId { get; init; }
+    public UserRole? Role { get; set; }
```
jwinston-bn, line 8: If we're gonna let this be null, then that further would support a single PATCH endpoint.

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
```diff
@@ -0,0 +1,78 @@
...
+            ThrowError("Requested role does not exist", StatusCodes.Status400BadRequest);
+        }
+
+        // remove existing roles for user with validation that they are valid roles
+        var auth0Roles = await _authService.GetUsersRolesAsync(accessToken, user.ProviderId, ct);
+        foreach (var auth0Role in auth0Roles)
+        {
+            if (!Enum.TryParse(auth0Role.Name, ignoreCase: true, out UserRole _))
+            {
+                throw new InvalidOperationException($"Auth0 Role not found in {nameof(UserRole)} enum: \"{auth0Role.Name}\".");
```
jwinston-bn, line 58: I'd use the ThrowError functionality here to send back a 400 with message. This will cause a 500 to send back.

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+        if (auth0Roles.Count > 1)
+        {
+            _logger.LogError("User {userId} unexpectedly has multiple roles: {roles}. Removing all of them.", user.Id, auth0Roles);
+        }
+
+        _logger.LogInformation("Removing roles {roles} from user {userId}.", auth0Roles, user.Id);
```
jwinston-bn, line 67: I'd add a check that if the new role to set is the same as the existing role, don't do anything else

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
```diff
@@ -0,0 +1,78 @@
...
+        await _authService.RemoveRolesFromUserAsync(accessToken, user.ProviderId, auth0Roles.Select(r => r.Id).ToList(), ct);
+
+        // update the user's role
+        _logger.LogInformation("Adding role {roleId} to user {userId}.", desiredAuth0RoleId, user.Id);
+        await _authService.AssignRoleToUserAsync(accessToken, user.ProviderId, desiredAuth0RoleId, CancellationToken.None);
+        user.Role = req.Role!.Value;
+        await _dbContext.SaveChangesAsync(CancellationToken.None);
```
jwinston-bn, line 74: Why aren't we passing `ct` here and above?

jwinston-bn: Oh is it so they don't go out of sync?

NateMerritt: Yeah.  At this point we don't want to cancel the operation until it's fully completed.

src/Aquifer.API/Endpoints/Users/Role/Set/Endpoint.cs
NateMerritt: Ok, that's what I expected.  I'll keep this Auth0 validation just in case we get into a weird state and always remove all roles from the Auth0 user.

jwinston-bn (approved):

PR #644, opened 2025-02-10
Add notifications routes to Internal API.

NateMerritt:
New routes:
* `GET /notifications`: Gets all notifications for the current user for the past 30 days with optional paging and filtering parameters.
* `PATCH /notifications/{kind}/{id}`: Mark a notification as read/unread.

Note that because there is no true notification entity the notification kind and notification entity ID are used to identify a notification (e.g. "Comment" and a comment ID of 123).  We could introduce a new notification ID later if desired which would only require modifying the PATCH route to use that new ID instead of the above combination.

A new DB table, `Notifications`, was added which only hold read notifications.  For clarity, I still added an `IsRead` column to this table which for now will always be true.  Marking an existing read notification as unread will remove the corresponding row from this table.


src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -0,0 +1,197 @@
...
+        string Comment,
+        DateTime Created,
+        int UserId,
+        string UserFirstName,
+        string UserLastName,
+        int ResourceContentId,
+        string ResourceEnglishLabel,
+        string ParentResourceDisplayName);
+
+    private static async Task<IReadOnlyList<ResourceCommentNotificationData>> GetResourceCommentNotificationDataAsync(
```
NateMerritt, line 146: The [previous logic ](https://github.com/BiblioNexusStudio/aquifer-server/blob/67b0cf92113516b88b4e1543945cffd178c5401d/src/Aquifer.Jobs/Subscribers/NotificationMessageSubscriber.cs#L100-L117) also limited notifications to only users in the same company as the user who commented.  That would add extra DB joins here that don't seem worth it so I left it as notifying anyone previously assigned to the resource content version.

I'd argue that it would also be useful to notify anyone who had previously commented on the thread as well but that wasn't in the previous logic so I didn't add it here.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -0,0 +1,197 @@
...
+                (
+                    SELECT NULL
+                    FROM ResourceContentVersionAssignedUserHistory rcvauh
+                    WHERE
+                        rcvauh.ResourceContentVersionId = rcv.Id AND
+                        rcvauh.Created < c.Created AND
+                        rcvauh.AssignedUserId = @userId
+                ) AND
+                c.UserId <> @userId AND
+                c.Created > @minimumCommentCreatedDate
```
NateMerritt, line 185: Should this also limit to `rcv.IsDraft = 1`?  I think this might be necessary in order to click on the notification and see the relevant comment thread.  That said, if we add this then comment notifications will just disappear when a draft is published.

src/Aquifer.API/Endpoints/Notifications/List/Response.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Aquifer.API.Common.Dtos;
+using Aquifer.Data.Entities;
+
+namespace Aquifer.API.Endpoints.Notifications.List;
+
+public sealed class Response
+{
+    public required NotificationKind Kind { get; init; }
```
NateMerritt, line 8: This uses a discriminated union pattern (though somewhat forced because C# doesn't officially support discriminated unions).  The `Comment` or `HelpDocument` sub-DTO will only be populated if the `Kind` here is `Comment` or `HelpDocument`, respectively.

src/Aquifer.API/Endpoints/Notifications/List/Response.cs
```diff
@@ -0,0 +1,33 @@
...
+    public required int Id { get; init; }
+    public required string Text { get; init; }
+    public required DateTime Created { get; init; }
+    public required UserDto User { get; init; }
+    public required int ResourceContentId { get; init; }
+    public required string ResourceEnglishLabel { get; init; }
+    public required string ParentResourceDisplayName { get; init; }
+}
+
+public sealed class HelpDocumentNotification
```
NateMerritt, line 25: I went ahead and added help documents as notifications because it helped me think through the multiple kinds of notifications.

src/Aquifer.Data/Migrations/20250210190519_AddNotificationEntity.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Aquifer.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddNotificationEntity : Migration
```
NateMerritt, line 9: I already ran this in Dev/QA.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -0,0 +1,197 @@
...
+        string ResourceEnglishLabel,
+        string ParentResourceDisplayName);
+
+    private static async Task<IReadOnlyList<ResourceCommentNotificationData>> GetResourceCommentNotificationDataAsync(
+        DbConnection dbConnection,
+        int userId,
+        DateTime minimumCommentCreatedDate,
+        CancellationToken ct)
+    {
+        // Get comment notifications for the current user where that user is assigned or has been previously been assigned to the associated
```
jwinston-bn, line 152: "been previously been"

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
jwinston-bn: They could have stuff that's now in published, and the sidebar still has the comments even if they're removed from inline.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -0,0 +1,197 @@
...
+                        Type = h.Type,
+                        Url = h.Url,
+                        ThumbnailUrl = h.ThumbnailUrl,
+                    },
+                    IsRead = readNotificationEntityIdsByNotificationKindMap.GetValueOrDefault(NotificationKind.HelpDocument)
+                        ?.Contains(h.Id)
+                            ?? false,
+                }))
+            .Where(x => !req.IsRead.HasValue || x.IsRead == req.IsRead.Value)
+            .OrderByDescending(n => n.Comment?.Created ?? n.HelpDocument!.Created)
```
jwinston-bn, line 91: Would it make more sense to put the `Created` at the top level of `Response`? I assume everything will have that property, and that would simplify this since it can keep growing.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -0,0 +1,197 @@
...
+        HelpDocumentType Type,
+        DateTime Created,
+        string Title,
+        string Url,
+        string? ThumbnailUrl);
+
+    private static async Task<IReadOnlyList<HelpDocumentNotificationData>> GetHelpDocumentNotificationDataAsync(
+        DbConnection dbConnection,
+         DateTime minimumHelpDocumentCreatedDate,
+         CancellationToken ct)
```
jwinston-bn, line 110: Spurious spaces

src/Aquifer.API/Endpoints/Notifications/List/Request.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Aquifer.API.Endpoints.Notifications.List;
+
+public sealed class Request
```
jwinston-bn, line 3: Wondering if we should add notification kind id array here. Thinking about in the future if there were notifications for Well vs Admin, etc. Or different pages/menus on admin wanting different types of notifications.

src/Aquifer.API/Endpoints/Notifications/Update/Endpoint.cs
```diff
@@ -0,0 +1,91 @@
...
+            // Only read notifications are stored in the DB.
+            var existingNotification = await _dbContext.Notifications
+                .AsTracking()
+                .Where(n => n.Kind == req.NotificationKind && n.NotificationEntityId == req.NotificationEntityId)
+                .FirstOrDefaultAsync(ct);
+
+            if (existingNotification is not null && !req.IsRead.Value)
+            {
+                // instead of marking as unread, delete the notification altogether
+                _dbContext.Notifications.Remove(existingNotification);
```
jwinston-bn, line 63: I don't understand why we remove this instead of just setting `IsRead` to 0 since we have the column.

src/Aquifer.Data/Entities/NotificationEntity.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Aquifer.Data.Entities;
+
+[Index(nameof(UserId), nameof(Kind), nameof(NotificationEntityId), IsUnique = true)]
+public sealed class NotificationEntity
+{
+    public int Id { get; set; }
+    public int UserId { get; set; }
+    public NotificationKind Kind { get; set; }
+    public int NotificationEntityId { get; set; }
+    public DateTime Created { get; set; }
```
jwinston-bn, line 12: I would add `Updated` since we have the IsRead that could flip. Also, I would like `Created` and `Updated` to have `= DateTime.UtcNow` so we don't have to worry about accidentally putting bad values in the DB for those. And please also add the `[SqlDefaultValue("getutcdate()")]` attributes. And then with the `Updated` prop, can add `IHasUpdatedTimestamp` interface to enforce that it updates whenever someone changes a value without needing to explicitly update it.

src/Aquifer.Data/Entities/NotificationEntity.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Aquifer.Data.Entities;
+
+[Index(nameof(UserId), nameof(Kind), nameof(NotificationEntityId), IsUnique = true)]
+public sealed class NotificationEntity
+{
+    public int Id { get; set; }
+    public int UserId { get; set; }
+    public NotificationKind Kind { get; set; }
+    public int NotificationEntityId { get; set; }
```
jwinston-bn, line 11: I would call this `NotificationKindId`. Entity makes me think it's something related to Entity Framework, which it isn't.

src/Aquifer.API/Endpoints/Notifications/Update/Request.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Aquifer.Data.Entities;
+
+namespace Aquifer.API.Endpoints.Notifications.Update;
+
+public sealed class Request
+{
+    public NotificationKind NotificationKind { get; init; }
+    public int NotificationEntityId { get; init; }
+    public bool? IsRead { get; init; }
```
jwinston-bn, line 9: I think this should be non-nullable and required.

src/Aquifer.API/Endpoints/Notifications/List/Request.cs
NateMerritt: That makes sense.  I'll add it.

src/Aquifer.API/Endpoints/Notifications/Update/Request.cs
NateMerritt: I wrote this as a PATCH route in which case there could be potential future properties that we also update on the notification and `IsRead` could then be `null` if it isn't updated alongside those potential future properties.

NateMerritt: I could imagine something like a future `bool? IsFlagged` property that allows flagging a notification for later.

Regardless, making patch properties nullable is future proof for a PATCH route that may have additional properties added in the future.  Making it non-nullable would not be future proof.  If we want it non-nullable then I'd advocate for a PUT route instead (which is hard to update with new properties while keeping server and client in-sync) or a POST route that only deals with read/unread, like `POST /notifications/{kind}/{id}/markasread` or something like that.  I definitely prefer PATCH over those other options, though.

jwinston-bn: For clarification, I'm not suggesting removing the nullability, because then the validator gets messed up anyway because it defaults to false. I'm just saying that the validator should require it to be set since the endpoint doesn't do anything else. If in the future we want to add additional functionality, then the validator can update accordingly, and that wouldn't impact an existing caller.

NateMerritt: Ah, I misunderstood.  I agree with that change and will update the validator accordingly.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
NateMerritt: I had a previous iteration where I pulled out a bunch of common properties to the root level.  I ended up moving to this sub-DTO pattern because we don't have a true notification entity and this kept things cleaner when mapping.  It also allows the sub-DTOs to diverge from each other in the future without locking them in to a common naming.

jwinston-bn (approved):

PR #646, opened 2025-02-10
Only translate resource contents for projects using the Aquifer platform.

NateMerritt:


src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -104,13 +104,29 @@ private async Task ProcessAsync(QueueMessage queueMessage,
...
         CancellationToken ct)
     {
+        const int aquiferProjectPlatformId = 1;
+
+        var project = (await _dbContext.Projects
+                .Where(p => p.Id == message.ProjectId)
+                .FirstOrDefaultAsync(ct))
+            ?? throw new InvalidOperationException($"Project with ID {message.ProjectId} does not exist.");
+
+        if (project.ProjectPlatformId != aquiferProjectPlatformId)
```
jwinston-bn, line 114: I'm still skeptical we even need this because we're not longer going to allow these types of projects. But if we did, wouldn't it make more sense to put this check in the Start endpoint for Projects? We already have the project object so no need to grab it again, and then it saves all the overhead of going into the queue and grabbing it in the first place.

NateMerritt: > I'm still skeptical we even need this because we're not longer going to allow these types of projects.

It doesn't hurt to add it even if we no longer have other kinds of projects and it will prevent future errors in the error stream.

> wouldn't it make more sense to put this check in the Start endpoint for Projects?

IMO any subscriber should be agnostic about the caller scenario and should handle validation itself.  It's not expensive to fetch the project and do this check.

jwinston-bn (approved):

PR #648, opened 2025-02-12
Add support for authenticated integration tests.

NateMerritt:
Notes:
* I created a client for each of the CMS roles (leaving out a couple less used roles).  If we want additional roles represented in the future it's easy to add them.
* I wrote code that can create the test users, one per role.  This code will only be run once (I already ran it locally); if a test user exists already (by email lookup) then the existing user will be used for the tests instead of making a new user every test run.
* All test users share the same password.
* In order to create the test users with a known password I used the existing Create User endpoint and then manually updated the password.
* In order to get a bearer token for the test users I created a new Auth0 application which allows authenticated directly with username/password via the [Resource Owner Password Flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/resource-owner-password-flow).  This integration test application only exists in the Dev/QA environment for security reasons.  Only the integration test project currently has the needed configuration to invoke fetching a bearer token via username/password and only it should ever use this functionality.
* The Aquifer.API.IntegrationTests project now has configuration for both Aquifer.API (to set up the API under test) and for its own purposes in order to authenticate users.  I made this distinction clear via comments because I use the API's configuration for accessing the management API and the integration test configuration for authenticating via username/password.
* Necessary environment variables have already been added to the GitHub Actions environment.

### DIFF
--- a/src/Aquifer.API/Services/Auth0Service.cs
+++ b/src/Aquifer.API/Services/Auth0Service.cs
@@ -13,6 +13,7 @@ namespace Aquifer.API.Services;
 public interface IAuth0Service
 {
     Task<string> GetAccessTokenAsync(CancellationToken ct);
+    Task<string> GetAccessTokenUsingResourceOwnerPasswordFlowAsync(string username, string password, CancellationToken ct);
 
     Task<IReadOnlyList<(string Id, string Name)>> GetAllRolesAsync(string accessToken, CancellationToken ct);
     Task<string?> GetRoleIdForRoleNameAsync(string accessToken, string roleName, CancellationToken ct);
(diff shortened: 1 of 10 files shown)


src/Aquifer.API/Services/Auth0Service.cs
```diff
@@ -59,6 +52,35 @@ public async Task<string> GetAccessTokenAsync(CancellationToken ct)
...
+        using var auth0AuthenticationClient = GetAuthenticationApiClient();
+
+        var request = new ResourceOwnerTokenRequest
+        {
+            ClientId = _auth0Options.ApplicationClientId,
+            ClientSecret = await _keyVaultClient.GetSecretAsync(KeyVaultSecretName.ResourceOwnerAuth0ClientSecret),
+            Audience = _auth0Options.Audience,
+            Username = username,
+            Password = password,
+            Scope = "openid",
```
NateMerritt, line 69: I couldn't the Resource Owner Password Flow to work without specifying a scope.

src/Aquifer.API/Services/Auth0Service.cs
```diff
@@ -59,6 +52,35 @@ public async Task<string> GetAccessTokenAsync(CancellationToken ct)
...
+    /// </summary>
+    public async Task<string> GetAccessTokenUsingResourceOwnerPasswordFlowAsync(string username, string password, CancellationToken ct)
+    {
+        using var auth0AuthenticationClient = GetAuthenticationApiClient();
+
+        var request = new ResourceOwnerTokenRequest
+        {
+            ClientId = _auth0Options.ApplicationClientId,
+            ClientSecret = await _keyVaultClient.GetSecretAsync(KeyVaultSecretName.ResourceOwnerAuth0ClientSecret),
+            Audience = _auth0Options.Audience,
```
NateMerritt, line 66: This needs to be the same audience used by the CMS.

jwinston-bn (approved):
Nice

jwinston-bn (approved):

PR #650, opened 2025-02-13
Update .editorconfig

NateMerritt:
Changes:
* Use `PascalCase` for contants, both `private` and `public`.
* Use `s_camelCase` for `static` and `static readonly` `private` fields.
* Use `_` for `private` fields.
* Require an empty line after statement blocks.
* Delete configuration that I accidentally duplicated.
* Update ReSharper wrapping settings for auto-format to put things on their own line.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -81,6 +81,7 @@ dotnet_style_null_propagation = true:error
 dotnet_style_object_initializer = true
 dotnet_style_operator_placement_when_wrapping = beginning_of_line
 dotnet_style_prefer_auto_properties = true
+dotnet_style_prefer_collection_expression.when_types_loosely_match = true:suggestion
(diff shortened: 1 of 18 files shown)


.editorconfig
```diff
@@ -226,6 +245,11 @@ csharp_preserve_single_line_statements = true
 
 # Naming rules
 
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = warning
+dotnet_naming_rule.private_constants_rule.severity = warning
+dotnet_naming_rule.private_constants_rule.style = pascal_case
```
NateMerritt, line 250: I don't personally like pascal case for `private` constants but that seems like what the rest of the solution was doing and is in the .net style guidelines.

jwinston-bn (approved):

PR #651, opened 2025-02-17
Add notifications bulk update route.

NateMerritt:


src/Aquifer.API/Endpoints/Notifications/Update/Endpoint.cs
```diff
@@ -10,7 +10,7 @@ public class Endpoint(AquiferDbContext _dbContext, IUserService _userService) :
 {
     public override void Configure()
     {
-        Patch("/notifications/{notificationKind}/{notificationEntityId}");
+        Patch("/notifications/{notificationKind}/{notificationKindId}");
```
NateMerritt, line 13: This fixes the existing PATCH route.  I missed a rename during review.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
```diff
@@ -0,0 +1,118 @@
...
+using FastEndpoints;
+using Microsoft.EntityFrameworkCore;
+
+namespace Aquifer.API.Endpoints.Notifications.BulkUpdate;
+
+public class Endpoint(AquiferDbContext _dbContext, IUserService _userService) : Endpoint<Request>
+{
+    public override void Configure()
+    {
+        Post("/notifications/update");
```
NateMerritt, line 13: @jwinston-bn, I used `update` in the route definition but named the folder `BulkUpdate` because the `Update` folder was already taken by the existing single notification PATCH route.  Does that naming work for you?

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
```diff
@@ -0,0 +1,118 @@
...
+            .Select(u => u.NotificationKind)
+            .Distinct()
+            .Except([NotificationKind.Comment, NotificationKind.HelpDocument])
+            .ToList();
+
+        if (invalidNotificationKinds.Count > 0)
+        {
+            ThrowError(
+                x => x.Updates,
+                $"Invalid {nameof(NotificationUpdate.NotificationKind)}: \"{string.Join("\", \"", invalidNotificationKinds)}\".",
```
jwinston-bn, line 32: Could we use `$"""..."""` here to make this easier to digest?

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        if (nonExistentRequestedCommentIds.Count > 0)
+        {
+            ThrowError(
+                x => x.Updates,
+                $"The following Comment IDs do not exist: {string.Join(", ", nonExistentRequestedCommentIds)}.",
+                StatusCodes.Status400BadRequest);
+        }
+
+        var requestedHelpDocumentIds = updates
```
jwinston-bn, line 59: I'm a little concerned that there'd be an assumption that a new notification kind would just automatically be able to use this route, but it can't without this also being updated.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        if (invalidNotificationKinds.Count > 0)
+        {
+            ThrowError(
+                x => x.Updates,
+                $"Invalid {nameof(NotificationUpdate.NotificationKind)}: \"{string.Join("\", \"", invalidNotificationKinds)}\".",
+                StatusCodes.Status400BadRequest);
+        }
+
+        var requestedCommentIds = updates
```
jwinston-bn, line 36: Seems like this method can just grow and grow. I'd personally pull each `kind` into a separate private method to handle it.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
```diff
@@ -0,0 +1,118 @@
...
+                StatusCodes.Status400BadRequest);
+        }
+
+        var created = DateTime.UtcNow;
+        foreach (var update in updates)
+        {
+            if (update.IsRead.HasValue)
+            {
+                // Note that only notifications that a user has interacted with are stored in the DB.
+                var existingNotification = await _dbContext
```
jwinston-bn, line 88: Wouldn't it be better to grab all these at once and then in the loop search the in-memory collection? This could be 100 round trips to the database couldn't it?

NateMerritt: I can't easily grab them all in a single call (because we don't have a true "notification ID") but I could get them in bulk by notification kind.  I'll make that change.

NateMerritt: I updated to use a `WHERE` query with an `OR` to fetch them in a single call.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
NateMerritt: An exception will be thrown above if a new kind is passed and isn't handled yet.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
NateMerritt: I don't follow.  It's a single line string so in what way does using a raw string literal help?

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
NateMerritt: Any objection to leaving as is and doing that refactor if we add more kinds?

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
jwinston-bn: It removes all the `\"` so it reads more naturally.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
jwinston-bn: I generally like smaller methods and named private methods make it easier to follow what's going on. To me that's part of SRP and OOP.

jwinston-bn (approved):

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
NateMerritt: 🤔 I tend to use raw string literals only for multi-line strings.  In this case using it would remove only two `\`s and add four `"`s.  Using a verbatim interpolated string would replace the two `\`s with a `"`.  All are about the same to me in terms of readability.

src/Aquifer.API/Endpoints/Notifications/BulkUpdate/Endpoint.cs
NateMerritt: I agree with you.  I think my threshold for "the method is too long" is just longer than your threshold.  I'll do some refactoring.

PR #652, opened 2025-02-18
Allow translated items to be selected for new projects.

NateMerritt:
Improvements:
* Update `GET /resources/untranslated` and `GET /resources/unaquiferized` to use the `ResourceContentSearchService`.  This required supporting an `IsTranslated` filter and an `IsNewestResourceContentVersion` filter.
* Update the `ResourceContentSearchService` to support multiple verse ID ranges (required for the above).
* Update verse ID range calculation for multiple chapters to condense consecutive distinct chapters into a range. This improves DB query performance, especially because the front-end will pass "all" chapters for a book of the Bible as "1, 2, ..., N" instead of "1-N" and each chapter was turned into its own verse ID range.

Additions:
* Add `GET /resources/translated` to fetch only translated resource contents that aren't already in a project.  I considered merging this new route into one route that supports untranslated, unaquiferized, and translated but now that they just call the search service this didn't duplicate as much code and the responses are slightly different so I opted for the new endpoint.
* Support `isAlreadyTranslated` as a flag when creating projects.  This defaults to false and thus is backwards compatible upon deployment.
* Ensure the translation subscriber skips translating these already translated resource contents but still assigns them when the project is started.

See https://github.com/BiblioNexusStudio/content-manager-web/pull/573 for the front-end changes.


tests/Aquifer.Common.UnitTests/Utilities/BibleUtilitiesTests.cs
```diff
@@ -0,0 +1,57 @@
...
+                (1_041_003_000, 1_041_005_999),
+                (1_041_007_000, 1_041_007_999),
+                (1_041_010_000, 1_041_014_999),
+                (1_041_020_000, 1_041_020_999),
+            ]
+        },
+    };
+
+    [Theory]
+    [MemberData(nameof(GetVerseRangesForBookAndChaptersTestData))]
```
NateMerritt, line 49: Unfortunately using attributes for test cases doesn't support arrays of tuples so I had to use `TheoryData` above.

src/Aquifer.Jobs/Subscribers/TranslationMessageSubscriber.cs
```diff
@@ -245,7 +245,8 @@ public async Task UpdateProjectPostTranslationAsync([ActivityTrigger] UpdateProj
             .Where(rcv => rcv.ResourceContent.ProjectResourceContents.Any(prc => prc.Project.Id == project.Id) &&
                 rcv.IsDraft &&
                 (rcv.ResourceContent.Status == ResourceContentStatus.TranslationAiDraftComplete ||
-                    rcv.ResourceContent.Status == ResourceContentStatus.AquiferizeAiDraftComplete) &&
+                    rcv.ResourceContent.Status == ResourceContentStatus.AquiferizeAiDraftComplete ||
+                    rcv.ResourceContent.Status == ResourceContentStatus.New) &&
```
NateMerritt, line 249: Adding `New` here ensures that already translated items that were skipped during translation are still assigned upon project start.

src/Aquifer.API/Endpoints/Projects/Create/Endpoint.cs
```diff
@@ -169,6 +176,75 @@ private async Task<List<ResourceContentEntity>> CreateOrFindAquiferizationResour
         return resourceContents;
     }
 
+    private async Task<List<ResourceContentEntity>> CreateOrFindTranslatedResourceContentAsync(
```
NateMerritt, line 179: This method could benefit from detailed review to make sure I got the validation and creation correct.

jwinston-bn (approved):

PR #656, opened 2025-02-20
Fix member access wrapping style config.

NateMerritt:
I put an example auto-format in the PR as well.


src/Aquifer.API/Program.cs
```diff
@@ -22,24 +22,27 @@
...
-        .EnableSensitiveDataLogging(sensitiveDataLoggingEnabled: builder.Environment.IsDevelopment())
-        .UseQueryTrackingBehavior(QueryTrackingBehavior.NoTracking))
+    .AddDbContext<AquiferDbContext>(
+        options => options
+            .UseAzureSql(configuration.ConnectionStrings.BiblioNexusDb, providerOptions => providerOptions.EnableRetryOnFailure(3))
+            .EnableSensitiveDataLogging(builder.Environment.IsDevelopment())
+            .UseQueryTrackingBehavior(QueryTrackingBehavior.NoTracking))
     .RegisterModules()
     .Configure<JsonOptions>(options => options.SerializerOptions.Converters.Add(new JsonStringEnumConverter()))
-    .AddHttpLogging(logging =>
```
NateMerritt, line 37: For a single lambda argument I prefer this previous style but there's no way that I can see to make an exception for that particular argument scenario.  The new auto-format style is arguably more readable, anyway.

jwinston-bn (approved):

PR #658, opened 2025-02-21
Bib 849 update assign resources modal

elisabeth-shakespeare:
### Two changes:
1. Add a flag `HasAdditionalReviewer` that returns true if any of the drafts has an additional reviewer assigned. This is to help the frontend with whether or not to show "Send to Review" or "Send to Manager"
2. We need a way to advance the user assignment of content already in `Company Review` that an Editor is assigned to. When the Editor hits "Send to Review" in that status, it should be send to the Default Reviewer or Manager, following the same logical flow of the `send-for-company-reviewer` endpoint, only keeping the status. To do that, I've allowed the CompanyReview status into that endpoint, and added a check for the current status so it could determine whether it needs to advance while keeping the logic for determining which manager to assign to.

### DIFF
--- a/src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
@@ -67,7 +67,8 @@ public class Endpoint(AquiferDbContext dbContext, IUserService userService) : En
                 ProjectEntity = rc.ProjectResourceContents.FirstOrDefault(prc => prc.ResourceContentId == request.Id) == null
                     ? null
                     : rc.ProjectResourceContents.First(prc => prc.ResourceContentId == request.Id).Project,
-                HasPublishedVersion = rc.Versions.Any(rcv => rcv.IsPublished)
+                HasPublishedVersion = rc.Versions.Any(rcv => rcv.IsPublished),
+                HasAdditionalReviewer = rc.Versions.Any(rcv => rcv.AssignedReviewerUser != null)
             })
             .AsSplitQuery()
             .FirstOrDefaultAsync(ct);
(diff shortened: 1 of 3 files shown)


src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -67,7 +67,8 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
                 ProjectEntity = rc.ProjectResourceContents.FirstOrDefault(prc => prc.ResourceContentId == request.Id) == null
                     ? null
                     : rc.ProjectResourceContents.First(prc => prc.ResourceContentId == request.Id).Project,
-                HasPublishedVersion = rc.Versions.Any(rcv => rcv.IsPublished)
+                HasPublishedVersion = rc.Versions.Any(rcv => rcv.IsPublished),
+                HasAdditionalReviewer = rc.Versions.Any(rcv => rcv.AssignedReviewerUser != null)
```
NateMerritt, line 71: Should this only check the most recent `ResourceContentVersion` for the `ResourceContent`?

elisabeth-shakespeare: I don't think this has to be that specific, because I wanted to answer the question "Has anyone ever chosen a user from the second dropdown in the Assign Resources modal before?"

jwinston-bn: I think Nate's correct. Wouldn't we only be interested in the current version since that's what's moving through?

jwinston-bn: Specifically the draft version

elisabeth-shakespeare: I think we would if we ever had draft versions in different statuses. I'm using the flag to determine whether to show "Send to Manager" or "Send to Review" on a button in the `EditorReview` status, so I made the assumption that all the draft versions would follow one status with its `ResourceContent`.

jwinston-bn (approved):

PR #660, opened 2025-02-21
Create a read-only DB connection and use it for all Public API data access.

NateMerritt:


src/Aquifer.Public.API/Program.cs
```diff
@@ -17,10 +17,11 @@
     ?? throw new InvalidOperationException($"Unable to bind {nameof(ConfigurationOptions)}.");
 
 builder.Services
-    .AddDbContext<AquiferDbContext>(options => options
-        .UseAzureSql(configuration.ConnectionStrings.BiblioNexusDb, providerOptions => providerOptions.EnableRetryOnFailure(3))
+    .AddDbContext<AquiferDbReadOnlyContext>(options => options
+        .UseAzureSql(configuration.ConnectionStrings.BiblioNexusReadOnlyDb, providerOptions => providerOptions.EnableRetryOnFailure(3))
         .EnableSensitiveDataLogging(sensitiveDataLoggingEnabled: builder.Environment.IsDevelopment())
         .UseQueryTrackingBehavior(QueryTrackingBehavior.NoTracking))
+    .AddScoped<AquiferDbContext, AquiferDbReadOnlyContext>()
```
NateMerritt, line 24: This enables common services that still use `AquiferDbContext` to inject the new `AquiferDbReadOnlyContext` instead for the Public API.

src/Aquifer.Data/AquiferDbContext.cs
```diff
@@ -8,27 +8,38 @@ namespace Aquifer.Data;
 
 public class AquiferDbContext : DbContext
 {
-    private readonly DbContextOptions<AquiferDbContext> _options;
```
NateMerritt, line 11: `AquiferDbReadOnlyContext` uses a `DbContextOptions<AquiferDbReadOnlyContext>` which couldn't be passed down to its base class constructor in `AquiferDbContext` so I had to make it use the common `DbContextOptions` base class.  It's only used for new-ing up an `AquiferDbContext` in the `ResourceStatusChangeHandler`.

jwinston-bn (dismissed):

jwinston-bn (dismissed):

jwinston-bn (approved):

PR #662, opened 2025-02-24
Bib 853 versification UI updates

kaseywright:
To allow the frontend to properly format the message for the user as per the ticket, I'm returning a `BaseMapping` in the response. It will be `null` if the base bible mapping is the same and include the `bookName chapter:verse` string if it differs. 

This will allow the client to easily mark the verse with the required `*` and have the base mapping available in a tooltip (if desired) for clarity.

I made this resource on dev/qa to work as a demo/test resource: `/resources/381634`


src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +55,75 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
                 })
                 .ToListAsync(ct);
+            
+            await MapVersificationDifferencesToResponseVersesAsync(req, response, ct);
         }
 
         await (response is null ? SendNotFoundAsync(ct) : SendOkAsync(response, ct));
     }
+
+    private static string FormatBaseBookChapterVerseMapping(int mappedVerseId)
```
kaseywright, line 65: This might be able to be replaced with `BibleBookCodeUtilities.FullNameFromId()`. I’ll look into this.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +55,75 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+            response.VerseMap = new ReadOnlyDictionary<int, int?>(new Dictionary<int, int?>()); // todo remove
+            return;
+        }
+
+        var bookId = BibleBookCodeUtilities.IdFromCode(response.BookCode);
+        var verseIds = GetVerseIds(response, bookId);
+        
+        if (!verseIds.Any())
+        {
+            response.VerseMap = new ReadOnlyDictionary<int, int?>(new Dictionary<int, int?>()); // todo remove
```
kaseywright, line 93: A result of the versification map is returned in the response for debugging purposes. This, and other occurrences, will be removed before the PR is upgraded.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +55,75 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+            return;
+        }
+
+        // Get versification mapping
+        var verseMap = await VersificationUtilities.ConvertVersificationRangeAsync(
+            req.BibleId,
+            verseIds.Min(),
+            verseIds.Max(),
+            0, // Target Bible ID is 1 (BSB)? 0 (EVB)?
```
kaseywright, line 102: As we do not allow for a target bible in the request, Which should be the default to compare against? Right now, I have it set to 0 ( ENG versification base ) but have done some testing using the 1 (BSB). We have two options:
- Hard code the default bible id. In this case Which should it be?
- Modify the Request to include an optional bible id

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +54,68 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+        var targetBookName = BibleBookCodeUtilities.FullNameFromId(targetBookId);
+        
+        return $"{targetBookName} {targetChapter}:{targetVerse}";
+    }
+
+    private static IReadOnlyList<int> GetVerseIds(Response response, BookId bookId)
+    {
+        return [.. response.Chapters
+            .SelectMany(chapter => chapter.Verses
+                .Select(verse => BibleUtilities.GetVerseId(bookId, chapter.Number, verse.Number)))];
```
NateMerritt, line 76: The collection expression syntax is optional but my IDE yells at me to use it in situations like this.  I think it's cleaner and more understandable to just use `.ToList()` like so instead of `[.. items]`.
```suggestion
        return response.Chapters
            .SelectMany(chapter => chapter.Verses
                .Select(verse => BibleUtilities.GetVerseId(bookId, chapter.Number, verse.Number)))
            .ToList();
```

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +54,68 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+            
+            await MapVersificationDifferencesToResponseVersesAsync(req, response, ct);
         }
 
         await (response is null ? SendNotFoundAsync(ct) : SendOkAsync(response, ct));
     }
+
+    private static string FormatBaseBookChapterVerseMapping(int mappedVerseId)
+    {
+        var ( targetBookId, targetChapter, targetVerse ) = BibleUtilities.TranslateVerseId(mappedVerseId);
```
NateMerritt, line 66: Nit: No spaces should be inside the parentheses for tuples.
```suggestion
        var (targetBookId, targetChapter, targetVerse) = BibleUtilities.TranslateVerseId(mappedVerseId);
```

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +54,68 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+        
+        if (!verseIds.Any())
+        {
+            return;
+        }
+        
+        var versificationMap = await VersificationUtilities.ConvertVersificationRangeAsync(
+            req.BibleId,
+            verseIds.Min(),
+            verseIds.Max(),
```
NateMerritt, line 97: `GetVerseIds` gets all verse IDs but then they are thrown away here unless they are the min or max.  Instead, you could use the request data to get only the first verse ID and the last verse ID to use here.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: The `BibleId` in the request is the target bible.  The source Bible (in our data) is Bible ID `0` (the "eng" English Bible superset "common" versification).

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Response.cs
```diff
@@ -21,4 +21,5 @@ public class ResponseChapterVerses
 {
     public required int Number { get; set; }
     public required string Text { get; set; }
+    public string BaseMapping { get; set; } = null!;
```
NateMerritt, line 24: This name doesn't give a lot of context into the meaning of the data the property contains.  I suggest `public VerseReference? OriginalVerseReference { get; set; }` where `VerseReference` is:
```
public class VerseReference
{
    public required string BookName { get; set; }
    public required string BookCode { get; set; }
    public required int BookNumber { get; set; }
    public required int ChapterNumber { get; set; }
    public required int VerseNumber { get; set; }
}
```
(but only include the book info that you actually need)

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +55,75 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+
+        // Create mapping dictionary only for verses that differ
+        var baseMappings = verseMap
+            .Where(mapping => mapping.Value.HasValue && mapping.Value.Value != mapping.Key)
+            .ToDictionary(
+                mapping => mapping.Key,
+                mapping => FormatBaseBookChapterVerseMapping(mapping.Value!.Value));
+
+        // Apply mappings to verses using LINQ instead of nested loops
+        response.Chapters = [.. response.Chapters.Select(chapter =>
```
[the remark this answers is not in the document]
NateMerritt: Being that this code is mutating the `response.Chapters` data it is probably better to just use a `foreach` loop here instead of mutating in the `Select` and (unnecessarily) reassigning back to the original property.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +54,68 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
 
         await (response is null ? SendNotFoundAsync(ct) : SendOkAsync(response, ct));
     }
+
+    private static string FormatBaseBookChapterVerseMapping(int mappedVerseId)
+    {
+        var ( targetBookId, targetChapter, targetVerse ) = BibleUtilities.TranslateVerseId(mappedVerseId);
+        var targetBookName = BibleBookCodeUtilities.FullNameFromId(targetBookId);
+        
+        return $"{targetBookName} {targetChapter}:{targetVerse}";
```
NateMerritt, line 69: Does this need to be the localized Bible book name?

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
kaseywright: Many requests do not include a verse range. For example, almost every request from the CMS for the components related to this ticket only send the `bookNumber`. The CMS then filters our the non-relevant verses and uses the response data as desired.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
kaseywright: Maybe not as this value this is meant to be the ENG versification mapping and would be displayed in a tooltip, or similar, when the editor is already viewing the localized reference. For example, the verse popover displays "nemiah 3:12*" (in russian) and the tooltip when focusing or mousing over that russian name would say something like "this references Nem 3:11".
The ticket does not currently call for this functionality. I only implemented it as an easy way to add context for the user. We could simply return a boolean in the response but I didn't think that added much value.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
NateMerritt: I'm arguing for a performance improvement.  The `GetVerseIds` method returns all verse IDs for potentially multiple chapters but it only needs to get the first and last verse IDs.  I'd actually inline it here because it only makes sense if there are chapters:
```
    var minChapter = response.Chapters.First();
    int minVerseId = BibleUtilities.GetVerseId(bookId, minChapter .Number, minChapter.Verses.First().Number);
    var maxChapter = response.Chapters.Last();
    int maxVerseId = BibleUtilities.GetVerseId(bookId, maxChapter .Number, maxChapter .Verses.Last().Number);
```

kaseywright: I like it. I'll make the change

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Response.cs
```diff
@@ -21,4 +21,12 @@ public class ResponseChapterVerses
 {
     public required int Number { get; set; }
     public required string Text { get; set; }
+    public VerseReference? EnglishBaseVerseMapping { get; set; }
```
NateMerritt, line 24: I'm ok with putting `English` here but could also see using `SourceTextVerseReference`.  @jwinston-bn, do you have an opinion?

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -51,8 +53,60 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+            .ToDictionary(
+                mapping => mapping.Key,
+                mapping => FormatBaseBookChapterVerseMapping(mapping.Value!.Value));
+        
+        foreach (var chapter in response.Chapters)
+        {
+            foreach (var verse in chapter.Verses)
+            {
+                var verseId = BibleUtilities.GetVerseId(bookId, chapter.Number, verse.Number);
+                verse.EnglishBaseVerseMapping = baseBibleVerseMappings.GetValueOrDefault(verseId)!;
```
NateMerritt, line 108: Why is the `!` appended?  `EnglishBaseVerseMapping` allows `null`.
```suggestion
                verse.EnglishBaseVerseMapping = baseBibleVerseMappings.GetValueOrDefault(verseId);
```

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Response.cs
jwinston-bn: Source is probably better for consistency, since the public API will need more flexibility in that regard. Wouldn't necessarily always be English.

src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
kaseywright: the `!` was a holdover from the LINQ implementation. It was crying about it but now it is not needed.

NateMerritt (approved):

PR #663, opened 2025-02-24
Use TableValuedParameter for verse ID range filtering.

NateMerritt:
This solved the poor SQL query performance by reducing the number of unique SQL queries we send to SQL Server by moving Verse ID ranges into a SQL Table Valued Parameter.  This allows reuse of existing query execution plans instead of SQL Server thinking every query is a new SQL query.  I also ran `EXEC sp_updatestats` in dev to update query statistics.

This change required adding the User-Defined Table Type `VerseIdRanges`.  Adding types is not supported by code first so I created a manual migration.  See https://dappertutorial.net/parameter-table-valued-parameter for details on Table Valued Parameters.  Note that TVPs are only supported by MS SQL Server so this will create more lock-in than we already have.  I don't believe this is an issue but I wanted to call it out.


src/Aquifer.Data/DapperExtensions.cs
```diff
@@ -407,6 +407,43 @@ public static async Task<IEnumerable<TReturn>> QueryWithRetriesAsync<TFirst, TSe
...
 
+    /// <summary>
+    /// Creates a Dapper TVP from the given <paramref name="source"/> where the properties of <typeref name="T"/>
+    /// correspond to the SQL column names.
+    /// </summary>
+    /// <typeparam name="T">The type corresponding to the SQL object's schema.</typeparam>
+    /// <param name="source">The items to convert.</param>
+    /// <param name="typeName">Database type name.</param>
+    /// <returns>Table-Valued Parameter</returns>
+    public static SqlMapper.ICustomQueryParameter AsTableValuedParameter<T>(this IEnumerable<T> source, string typeName)
```
NateMerritt, line 418: Dapper only supports creation of TVPs from a `DataTable`.  This method allows passing any `IEnumerable` (note that it comes with a minor performance cost of using reflection but that cost should be negligible compared to the DB operation itself).

I copied and tweaked this approach from https://sii.pl/blog/en/using-table-valued-parameters-with-dapper-in-net/.

src/Aquifer.Data/Migrations/20250224185220_AddVerseIdRangesTableType.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("""
+                CREATE TYPE [dbo].[VerseIdRanges] AS TABLE
+                (
+                    StartVerseId INT,
+                    EndVerseId INT
+                );
+                GRANT EXECUTE ON TYPE::[dbo].[VerseIdRanges] TO public;
```
NateMerritt, line 19: You have to grant execute rights after creation for user-defined types.

src/Aquifer.Data/Migrations/20250224185220_AddVerseIdRangesTableType.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Aquifer.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddVerseIdRangesTableType : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("""
+                CREATE TYPE [dbo].[VerseIdRanges] AS TABLE
```
NateMerritt, line 14: We could add a prefix or suffix to these (`UDTT` or `TableType` is common) if desired.

jwinston-bn (approved):

PR #664, opened 2025-02-24
Improve query performance of list `/resources/content` route.

NateMerritt:


jwinston-bn (approved):

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -376,10 +378,27 @@ public sealed class ResourceContentSearchService(AquiferDbContext dbContext) : I
...
 
         // Using CROSS APPLY and OUTER APPLY with inner groupings significantly improves performance over using JOINs with a main grouping.
+        const string verseIdRangesParamName = "verseIdRanges";
         var fromSql = $"""
             FROM ResourceContents rc
+            {(filter.VerseIdRanges?.Count > 0
+                ? $"""
+                    CROSS APPLY
+                    (
+                        SELECT DISTINCT vr.ResourceId
```
NateMerritt, line 388: Moving the `WHERE EXISTS` to a `CROSS APPLY` (and throwing away the result) seems to have better query performance.

PR #668, opened 2025-02-26
Update CI/CD to also deploy the Public API to India.

NateMerritt:
With this implementation we will deploy first to East US and then after that succeeds deploy to Central India.  If steps fail it should be possible to re-run each regional deployment independently.  I also added the region name to all deployment Slack notifications for clarity.

Future improvements:
* We could deploy to all regions in parallel if we move the migration script execution into a separate job that runs before deployment.
* [This commit](https://github.com/BiblioNexusStudio/aquifer-server/commit/b946fb00a6d5bb9b8d989bb506c9556e00ca90e5) added a bunch of redundant Git checkout steps.  I'm not sure why (possibly for Slack notifications?).  We could speed up deployments by removing that step.

### DIFF
--- a/.github/workflows/post-merge-public-api.yml
+++ b/.github/workflows/post-merge-public-api.yml
@@ -101,6 +101,7 @@ jobs:
                   azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
                   azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
                   azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+                  service-name: Public API (East US)
 
     deploy_public_api_to_qa:
         if: needs.check_should_deploy.outputs.should_deploy == 'true'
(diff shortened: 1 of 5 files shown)


.github/workflows/post-merge-public-api.yml
```diff
@@ -160,7 +160,53 @@ jobs:
...
+              with:
+                  ref: ${{ github.event.pull_request.head.sha }}
+            - uses: ./.github/actions/deploy-to-env
+              id: deploy
+              with:
+                  environment: prod
+                  full-environment: prod-public
+                  connection-url: ${{ vars.DB_CONNECTION_URL_PROD }}
+                  web-app-name: ${{ vars.AZURE_WEB_APP_NAME_PUBLIC_PROD_INDIA }}
+                  resource-group: ${{ vars.AZURE_RESOURCE_GROUP_PROD_INDIA }}
```
NateMerritt, line 188: I already added these new variables in the repo's GitHub environment settings.

.github/workflows/post-merge-public-api.yml
```diff
@@ -160,7 +160,53 @@ jobs:
...
+                  azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
+                  azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+                  azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  linear-api-key: ${{ secrets.LINEAR_API_KEY }}
+                  slack-bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
+                  service-name: Public API (Central India)
+
+    delete_hotfix_branch:
```
NateMerritt, line 197: Moving this to a separate job that depends upon the deployment jobs prevents deleting the hotfix branch until all deployments have succeeded.

jwinston-bn (approved):

PR #669, opened 2025-02-26
Populate counts of project response by sum of source words

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+
+public interface IProjectWithCount
+{
+    public ProjectResourceStatusCounts Counts { get; set; }
+    public int Id { get; set; }
+}
+
+public static class ProjectResourceStatusCountHelper
+{
+    private static async Task<List<ProjectResourceSourceWordCount>> GetCountsPerProjectAsync(
```
NateMerritt, line 15: This method and query are not filtering by project ID(s) so this operation is performed for every project in the DB.  Is that intentional?

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        const string query = """
+                             SELECT
+                                 SUM(RCVD.SourceWordCount) AS WordCount,
+                                 RC.Status as Status,
+                                 PRC.ProjectId as ProjectId
+                             FROM ResourceContents RC
+                             JOIN ProjectResourceContents PRC on RC.Id = PRC.ResourceContentId
```
NateMerritt, line 25: Nit: Casing.
```suggestion
                             JOIN ProjectResourceContents PRC ON RC.Id = PRC.ResourceContentId
```

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
elisabeth-shakespeare: yes, this method/query is for the project list endpoints

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    private static async Task<List<ProjectResourceSourceWordCount>> GetCountsPerProjectAsync(
+        AquiferDbContext dbContext,
+        CancellationToken ct)
+    {
+        const string query = """
+                             SELECT
+                                 SUM(RCVD.SourceWordCount) AS WordCount,
+                                 RC.Status as Status,
+                                 PRC.ProjectId as ProjectId
```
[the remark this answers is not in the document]
elisabeth-shakespeare: Thanks, had a theory that it had introduced a bug earlier and then found the real bug and didn't go back and fix this.

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+                                 WHERE x.LatestVersionRank = 1
+                             ) RCVD ON RCVD.ResourceContentId = RC.Id
+                             GROUP BY [ProjectId], [Status]
+                             """;
+
+        return await dbContext.Database.SqlQueryRaw<ProjectResourceSourceWordCount>(query).ToListAsync(ct);
+    }
+
+    public static async Task PopulateCountsPerProjectAsync<T>(List<T> projects, AquiferDbContext dbContext, CancellationToken ct)
+        where T : class, IProjectWithCount
```
NateMerritt, line 45: It looks like the `IProjectWithCount` interface was introduced in order to allow mutation of the `projects` argument here in this method.  I like the goal of having one helper class to do this operation for both usages.  Using the interface as a way to do that, especially on the `Response` class, is unexpected.

I suggest moving to the following pattern:
1. Don't mutate objects in `static` helper class methods.  The more places an object is mutated the harder it is to understand the code.  Embracing a level of functional programming makes it easier to understand how code interacts with an object.  I'd personally argue that all object mutation should happen at most across methods in a single class, or even better, only in a single method.
2. Therefore, update this method to return data instead of mutating the argument object.  This can be done by accepting an `IEnumerable<int> projectIds` and returning a `Task<Dictionary<int, ProjectResourceStatusCounts>>` (or `ReadOnlyDictionary` if desired) from this method.  The `Endpoints` can then set the `ProjectResourceStatusCounts` on the `Response` objects using a dictionary lookup.

elisabeth-shakespeare: Thank you. I got stuck on this and needed some help to clean it up

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
NateMerritt: It's also called from the `/projects/assigned-to-self` endpoint [here](https://github.com/BiblioNexusStudio/aquifer-server/pull/669/files#diff-85fd9cb9b48ca94c00176c339697b5ccbe61e88db73cc16f877b6153bc7a2362R61) which doesn't need all project data, only the projects assigned to the currently authenticated user, correct?

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+                NotStarted =
+                    counts.Where(x => ProjectResourceStatusCounts.NotStartedStatuses.Contains(x.Status)).Sum(x => x.WordCount ?? 0),
+                EditorReview =
+                    counts.Where(x => ProjectResourceStatusCounts.EditorReviewStatuses.Contains(x.Status)).Sum(x => x.WordCount ?? 0),
+                InCompanyReview =
+                    counts.Where(x => ProjectResourceStatusCounts.InCompanyReviewStatuses.Contains(x.Status)).Sum(x => x.WordCount ?? 0),
+                InPublisherReview =
+                    counts.Where(x => ProjectResourceStatusCounts.InPublisherReviewStatuses.Contains(x.Status)).Sum(x => x.WordCount ?? 0),
+                Completed = counts.Where(x => ProjectResourceStatusCounts.CompletedStatuses.Contains(x.Status)).Sum(x => x.WordCount ?? 0)
+            };
```
NateMerritt, line 62: I've seen this code duplicated in a bunch of places.  It may make sense to add a new `ProjectResourceStatusCounts` constructor that takes an `IEnumerable<(ResourceContentStatus status, int? WordCount)>` and does this logic in the constructor.

elisabeth-shakespeare: wouldn't `ProjectResourceStatusCounts.NotStartedStatuses.Contains(x.Status)).Sum(x => x.WordCount ?? 0)` still be repeated when making the constructor?

src/Aquifer.API/Endpoints/Projects/Get/Endpoint.cs
```diff
@@ -33,6 +33,20 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
...
+            projectResourceItems.Where(i => ProjectResourceStatusCounts.CompletedStatuses.Contains(i.Status))
+                .Sum(i => i.WordCount ?? 0),
+            projectResourceItems.Where(i => ProjectResourceStatusCounts.InCompanyReviewStatuses.Contains(i.Status))
+                .Sum(i => i.WordCount ?? 0),
+            projectResourceItems.Where(i => ProjectResourceStatusCounts.InPublisherReviewStatuses.Contains(i.Status))
+                .Sum(i => i.WordCount ?? 0),
+            projectResourceItems.Where(i => ProjectResourceStatusCounts.EditorReviewStatuses.Contains(i.Status))
+                .Sum(i => i.WordCount ?? 0),
+            projectResourceItems.Where(i => ProjectResourceStatusCounts.NotStartedStatuses.Contains(i.Status))
+                .Sum(i => i.WordCount ?? 0));
```
elisabeth-shakespeare, line 48: I'm not sure how much this helps with the new constructor on `ProjectResourceStatusCounts`. Is this what you meant?

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
NateMerritt: It would move to the constructor.

src/Aquifer.API/Endpoints/Projects/Get/Endpoint.cs
```diff
@@ -33,6 +33,10 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
         var items = await GetProjectItemsAsync(req.ProjectId, ct);
 
         project.Items = items.OrderBy(x => x.SortOrder).ThenBy(x => x.EnglishLabel);
+
+        var projectResourceItems = project.Items as ProjectResourceItem[] ?? project.Items;
+        project.Counts = new ProjectResourceStatusCounts(projectResourceItems.Select(pri => (pri.Status, pri.WordCount)).ToList());
```
NateMerritt, line 38: This looks like it can be simplified.
```suggestion
        project.Counts = new ProjectResourceStatusCounts(project.Items.Select(pri => (pri.Status, pri.WordCount)).ToList());
```

src/Aquifer.API/Endpoints/Projects/List/Endpoint.cs
```diff
@@ -17,58 +17,58 @@ public override void Configure()
...
+                    Resource = p.ProjectResourceContents.First().ResourceContent.Resource.ParentResource.DisplayName,
+                    ItemCount = p.ProjectResourceContents.Count,
+                    WordCount = p.SourceWordCount,
+                    IsStarted = p.Started != null,
+                    IsCompleted = p.ActualPublishDate != null,
+                    Days =
+                        p.ProjectedDeliveryDate.HasValue
+                            ? p.ProjectedDeliveryDate.Value.DayNumber - DateOnly.FromDateTime(DateTime.UtcNow).DayNumber
+                            : null,
+                    Counts = new ProjectResourceStatusCounts()
```
NateMerritt, line 57: I suggest making `Counts` not `required` on the `Response` and then you can omit this unnecessary allocation (by removing this line entirely).

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+                         FROM
+                         (
+                             SELECT
+                                 *,
+                                 ROW_NUMBER() OVER (PARTITION BY ResourceContentId ORDER BY [Version] DESC) AS LatestVersionRank
+                             FROM ResourceContentVersions
+                         ) x
+                         WHERE x.LatestVersionRank = 1
+                     ) RCVD ON RCVD.ResourceContentId = RC.Id
+                     WHERE PRC.ProjectId IN ({string.Join(", ", projectIds)})
```
NateMerritt, line 33: I experienced a lot of query performance pain this week because of SQL query plans being different when different SQL query strings are passed to SQL server.  By injecting the `projectIds` directly into the SQL query text, SQL Server thinks that the queries are entirely different if any project IDs are different because the SQL query strings don't match.  If SQL parameters are used (e.g. `WHERE PRC.ProjectId IN @projectIds`) then it will use the same query plan because the SQL query string remains the same.

I'm not suggesting changing anything right now for this PR (there are existing endpoints that do this same thing) but I'm happy to demo how to use Dapper some time which allows passing `IEnumerable<T>` parameter arguments, unlike `SqlQueryRaw().

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        var counts = await GetCountsPerProjectAsync(projectIds, dbContext, ct);
+        var countsPerProject = new Dictionary<int, ProjectResourceStatusCounts>();
+        foreach (var id in projectIds)
+        {
+            countsPerProject[id] = new ProjectResourceStatusCounts(
+                counts.Where(p => p.ProjectId == id).Select(pri => (pri.Status, pri.WordCount)).ToList());
+        }
+
+        return countsPerProject;
```
NateMerritt, line 55: Using a LINQ `GroupBy` can reduce the amount of code (this suggestion is untested):
```suggestion
        return (await GetCountsPerProjectAsync(projectIds, dbContext, ct))
            .GroupBy(c => c.ProjectId)
            .ToDictionary(
                grp => grp.Key,
                grp => new ProjectResourceStatusCounts(grp.Select(c=> (c.Status, c.WordCount)).ToList());
        }
```
Note that this will only work if `GetCountsPerProjectAsync` returns at least one item for each project ID passed into it, which should be true.

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCounts.cs
```diff
@@ -39,9 +39,30 @@ public class ProjectResourceStatusCounts
...
+        ResourceContentStatus.AquiferizeAiDraftComplete,
+        ResourceContentStatus.AquiferizeEditorReview,
+        ResourceContentStatus.TranslationEditorReview,
+        ResourceContentStatus.AquiferizeCompanyReview,
+        ResourceContentStatus.TranslationCompanyReview
+    ];
+
+    public ProjectResourceStatusCounts() { }
+
+    public ProjectResourceStatusCounts(List<(ResourceContentStatus Status, int? WordCount)> resourceData)
```
NateMerritt, line 53: Using an `IReadOnlyList<T>` allows callers to pass an `Array<T>`, a `List<T>`, or other implementing types, which makes it easier for caller to use this method.  I generally choose `IEnumerable<T>` or `IReadOnlyList<T>` depending upon if my method needs to enumerate the items more than once.
```suggestion
    public ProjectResourceStatusCounts(IReadOnlyList<(ResourceContentStatus Status, int? WordCount)> resourceData)
```

src/Aquifer.API/Endpoints/Projects/List/Response.cs
```diff
@@ -16,5 +16,5 @@ public class Response
     public required bool IsStarted { get; set; }
     public required bool IsCompleted { get; set; }
 
-    public required ProjectResourceStatusCounts Counts { get; set; }
+    public ProjectResourceStatusCounts? Counts { get; set; }
```
NateMerritt, line 19: Same note here as the above.

src/Aquifer.API/Endpoints/Projects/List/Endpoint.cs
```diff
@@ -17,58 +17,68 @@ public override void Configure()
...
+        var statusCountsPerProject =
+            await ProjectResourceStatusCountHelper.GetResourceStatusCountsPerProjectAsync(
+                null,
+                dbContext,
+                ct);
+
+        foreach (var p in projects)
+        {
+            p.Counts = statusCountsPerProject.TryGetValue(p.Id, out var projectId)
+                ? statusCountsPerProject[p.Id]
```
NateMerritt, line 69: The dictionary item is contained in the `Dictionary.TryGetValue()` out variable.
```suggestion
            p.Counts = statusCountsPerProject.TryGetValue(p.Id, out var projectResourceStatusCounts)
                ? projectResourceStatusCounts
```

src/Aquifer.API/Endpoints/Projects/ProjectResourceStatusCountHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+                              ) x
+                              WHERE x.LatestVersionRank = 1
+                          ) RCVD ON RCVD.ResourceContentId = RC.Id
+                          """;
+        if (projectIds != null)
+        {
+            selectQuery += $" WHERE PRC.ProjectId IN ({string.Join(", ", projectIds)}) ";
+        }
+
+        selectQuery += " GROUP BY ProjectId, Status";
```
NateMerritt, line 39: IMO using string interpolation with a ternary operator is easier to read. It also allows .net to build the string more efficiently than string concatenation, which creates a new string for every `+=` operator (not great).  Here's [the essay](https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/) explaining how string interpolation is more efficient but the tl;dr: is to prefer to use it everywhere you can (or use a `StringBuilder` if you have an unknown number of strings to add).
```suggestion
        var selectQuery = $"""
                          SELECT
                              SUM(RCVD.SourceWordCount) AS WordCount,
                              RC.Status,
                              PRC.ProjectId
                          FROM ResourceContents RC
                          JOIN ProjectResourceContents PRC ON RC.Id = PRC.ResourceContentId
                          LEFT JOIN
                          (
                              SELECT *
                              FROM
                              (
                                  SELECT
                                      *,
                                      ROW_NUMBER() OVER (PARTITION BY ResourceContentId ORDER BY [Version] DESC) AS LatestVersionRank
                                  FROM ResourceContentVersions
                              ) x
                              WHERE x.LatestVersionRank = 1
                          ) RCVD ON RCVD.ResourceContentId = RC.Id
                          {(projectIds != null ? $"WHERE PRC.ProjectId IN ({string.Join(", ", projectIds)})" : "")}
                          GROUP BY ProjectId, Status
                          """;

```

NateMerritt (changes requested):
I left some comments for minor improvements and am happy to merge after they are addressed.

NateMerritt (approved):

PR #670, opened 2025-02-26
Add Public API Integration Test Suite.

NateMerritt:
Notes:
* The tests are purposefully not very specific in terms of response content because the DB data could change and then break the tests.  The tests do check that all fields are populated and have reasonable results.  When possible they also check for specific results in the response output.
* I discovered that I needed to turn off caching (both response caching on the client and output caching on the server) due to how FastEndpoints creates test request URIs (see https://github.com/FastEndpoints/FastEndpoints/issues/892).  As part of doing that I had to update `FastEndpoints.Testing` which ultimately resulted in an update of all test NuGet packages and migrating from Xunit 2 to Xunit 3.
* FastEndpoints uses the `Response` class to _deserialize_ the response JSON so that we can interact with it in the test method.  This works fine if that object is immutable or has no properties derived from other properties.  In the case where we (a) read from the DB into a response and (b) ignore that property on the response and serialize to a different property on the response, it results in incorrect deserialization.  The solution to this was to make all properties have setters and not be derived.  Doing this only required moving that derivation logic into the `Endpoint` itself.
* I also added an `AnonymousClient` to the testing `App`s; this has the `api-key` header set and should be used for anonymous requests.  The baked in `App.Client` now does not have any headers set and likely won't be used (except for possibly testing api-key header handling).


src/Aquifer.Public.API/Endpoints/Resources/Get/Associations/Endpoint.cs
```diff
@@ -48,6 +49,19 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
                 .ToList()
         };
 
+        foreach (var passageAssociation in response.PassageAssociations)
```
jwinston-bn, line 52: What is all this? Is it related to the testing?

NateMerritt: This is related to the note in the PR description:

> FastEndpoints uses the `Response` class to deserialize the response JSON so that we can interact with it in the test method. This works fine if that object is immutable or has no properties derived from other properties. In the case where we (a) read from the DB into a response and (b) ignore that property on the response and serialize to a different property on the response, it results in incorrect deserialization. The solution to this was to make all properties have setters and not be derived. Doing this only required moving that derivation logic into the `Endpoint` itself.

The `Response` class for this endpoint was also updated.

NateMerritt: I'd argue that all `Response` classes should be immutable.  Right now we often read from the DB into the `Response` classes and then modify things on the response.  Making them immutable would require updates to use an intermediary type when reading from the DB for these scenarios and I didn't want to do substantial refactoring.

jwinston-bn (dismissed):

jwinston-bn (approved):

PR #675, opened 2025-02-27
Use channels to publish track resource content requests in background.

NateMerritt:
I confirmed locally that the response returns before the channel processing runs.

### DIFF
--- a/src/Aquifer.Common/Messages/Publishers/ResourceContentRequestTrackingMessagePublisher.cs
+++ b/src/Aquifer.Common/Messages/Publishers/ResourceContentRequestTrackingMessagePublisher.cs
@@ -1,6 +1,8 @@
-﻿using Aquifer.Common.Messages.Models;
+﻿using System.Threading.Channels;
+using Aquifer.Common.Messages.Models;
 using Aquifer.Common.Services.Caching;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Hosting;
 
 namespace Aquifer.Common.Messages.Publishers;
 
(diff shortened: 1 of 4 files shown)


src/Aquifer.Common/Messages/Publishers/ResourceContentRequestTrackingMessagePublisher.cs
jwinston-bn, line 1: This doesn't feel like the right place or name for this file since it's a background service that's always running. I think of this folder as housing things that publish to a queue. I feel like it should go into a services folder, or even BackgroundServices to be explicit.

NateMerritt: I see the channel usage as an implementation detail of the queue publishing which is why I abstracted it.

src/Aquifer.Common/Services/TrackResourceContentRequestBackgroundService.cs
```diff
@@ -0,0 +1,45 @@
...
+            while (await _channel.Reader.WaitToReadAsync(stoppingToken))
+            {
+                try
+                {
+                    var message = await _channel.Reader.ReadAsync(stoppingToken);
+                    await PublishTrackResourceContentRequestMessageCoreAsync(message, stoppingToken);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, $"Failed to publish a {nameof(TrackResourceContentRequestMessage)}. Gracefully skipping.");
```
NateMerritt, line 28: Without exception handling the app will crash if an exception is thrown.  See https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/hosting-exception-handling.

PR #678, opened 2025-03-04
Fix key not found error

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Projects/AssignedToSelf/List/Endpoint.cs
```diff
@@ -48,7 +48,17 @@ await ProjectResourceStatusCountHelper.GetResourceStatusCountsPerProjectAsync(
...
+            p.Counts = countsPerProject.TryGetValue(p.Id, out var projectResourceStatusCounts)
+                ? projectResourceStatusCounts
+                : new ProjectResourceStatusCounts
+                {
+                    NotStarted = 0,
+                    EditorReview = 0,
+                    InCompanyReview = 0,
+                    InPublisherReview = 0,
+                    Completed = 0,
+                    Remaining = 0
```
jwinston-bn, line 60: All these default to 0 anyway. Think you can just do new().

jwinston-bn (approved):

PR #694, opened 2025-03-07
Create endpoints and subscriber to handle audio uploads for resource content.

NateMerritt:
Details:
1. Add two new endpoints, one to upload a resource content audio file, the other to generically get the status of uploads (which could also be used for future upload types like images or videos).
2. Add a new subscriber to process the uploaded audio files.
3. Use a new CDN container, `aquifer-dev-content`, which allows dev/QA environment uploads to live on the CDN without overwriting production CDN data.
4. Add configuration for accessing the CDN blob storage in `aquiferstorage` to Jobs for publishing to the CDN.
5. Add configuration for accessing the blob storage for the current environment (`aquiferstoragedev` or `aquiferstorageprod`) to Aquifer.API and Jobs in order to upload the audio file to temp blob storage.
6. Use ffmpeg in the subscriber to do audio processing, including speech normalization and silence detection.  This copies the strategies already in use in aquifer-utilities and content-loader.
7. Create a new resource content version which references the uploaded file as well as a status history for the resource content.


src/Aquifer.API/Modules/Resources/ResourceContentJsonSchema.cs
```diff
@@ -1,17 +0,0 @@
-namespace Aquifer.API.Modules.Resources;
```
NateMerritt, line 1: Moved to `Aquifer.Data.Schemas`.

src/Aquifer.Common/Services/AzureClientService.cs
```diff
@@ -1,30 +1,41 @@
...
+            ExcludeEnvironmentCredential = true,
+            ExcludeInteractiveBrowserCredential = true,
+            ExcludeWorkloadIdentityCredential = true,
+            ExcludeAzureDeveloperCliCredential = true,
+            ExcludeAzurePowerShellCredential = true,
+            ExcludeSharedTokenCacheCredential = true
+        });
+    }
+
+    public AzureClientService(TokenCredential credential)
```
NateMerritt, line 33: I added this overload to support testing in LINQPad.

src/Aquifer.Jobs/Subscribers/UploadResourceContentAudioMessageSubscriber.cs
```diff
@@ -0,0 +1,421 @@
...
+
+namespace Aquifer.Jobs.Subscribers;
+
+public sealed class CdnOptions
+{
+    public required string AquiferContentContainerName { get; init; }
+    public required string CdnBaseUri { get; init; }
+}
+
+public sealed class FfmpegOptions
```
NateMerritt, line 28: I could see moving all of the Ffmpeg stuff into its own service in the future.  I decided to not do that refactor yet because it's only used in this one place.  However, I organized the code to make it easier to pull things out in the future.

src/Aquifer.Jobs/Subscribers/UploadResourceContentAudioMessageSubscriber.cs
```diff
@@ -0,0 +1,421 @@
...
+        // Update ResourceContentVersion.Content with new CDN URLs.
+        var hasSteps = audioContent.Mp3?.Steps?.Any() == true && audioContent.Webm?.Steps?.Any() == true;
+        if (message.StepNumber.HasValue)
+        {
+            var mp3Step = audioContent.Mp3?.Steps?.FirstOrDefault(s => s.StepNumber == message.StepNumber);
+            var webmStep = audioContent.Webm?.Steps?.FirstOrDefault(s => s.StepNumber == message.StepNumber);
+
+            if (mp3Step == null || webmStep == null)
+            {
+                throw new InvalidOperationException(
```
NateMerritt, line 222: If these exceptions throw then we'll have uploaded a file to the CDN but the ResourceContentVersion.Content won't be referencing it.  I thought about refactoring but it got messier to split the validation and assignment.  The endpoint is already validating this so I think this is sufficient and should never happen in practice.

src/Aquifer.Jobs/Subscribers/UploadResourceContentAudioMessageSubscriber.cs
```diff
@@ -0,0 +1,421 @@
...
+                int.TryParse(chapterAndVerseParts[1], out var verseNumber))
+            {
+                bibleReference = (BookId: bookId, ChapterNumber: chapterNumber, VerseNumber: verseNumber);
+            }
+        }
+
+        // get a name like "043_LUK_001_001" or "angel-of-the-lord"
+        var blobifiedResourceName = bibleReference.HasValue
+            ? $"{(int) bibleReference.Value.BookId:D3}_{BibleBookCodeUtilities.CodeFromId(bibleReference.Value.BookId)}_{bibleReference.Value.ChapterNumber:D3}_{bibleReference.Value.VerseNumber:D3}"
+            : resourceEnglishLabel.ToKebabCase();
```
NateMerritt, line 329: This is why the `CaseConverter` package was added to the project.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+    AquiferDbContext _dbContext,
+    IBlobStorageService _blobStorageService,
+    IUploadResourceContentAudioMessagePublisher _uploadResourceContentAudioMessagePublisher,
+    IUserService _userService,
+    ILogger<Endpoint> _logger)
+    : Endpoint<Request, Response>
+{
+    public override void Configure()
+    {
+        Post("/resources/content/{resourceContentId}/uploads");
```
jwinston-bn, line 27: Should this have Audio in its path? I'm wondering about a situation in the future where we need an image upload, and this endpoint is very tied to audio. Or would you see it using this same endpoint and splitting on mime type into separate helpers/methods?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+            .FirstOrDefaultAsync(rc => rc.Id == request.ResourceContentId, ct);
+
+        if (resourceContent is null || resourceContent.MediaType != ResourceContentMediaType.Audio)
+        {
+            await SendNotFoundAsync(ct);
+            return;
+        }
+
+        // we could change this in the future if desired and support additional audio types
+        if (request.File.ContentType != "audio/mpeg")
```
jwinston-bn, line 46: Is there a technical reason to restrict this to just mp3? I would allow other types like webm, etc.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            ThrowError(x => x.File, "File must be an mp3.", StatusCodes.Status400BadRequest);
```
jwinston-bn, line 48: Reminder that `ThrowError` will log an actual error and result in a Slack alert. It's useful outside of `HandleAsync`, but if you're in that anyway you might as well just return a 400 if it's just something the client needs to do differently.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        var resourceContentVersion = await _dbContext.ResourceContentVersions
```
jwinston-bn, line 51: What's the point of going back to the database here? Why not just get this along with `resourceContent` at line 36?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
jwinston-bn: Also, wouldn't it make more sense to do this check before any database calls?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+            .Where(rcv => rcv.ResourceContentId == request.ResourceContentId)
+            .OrderByDescending(rcv => rcv.Version)
+            .FirstAsync(ct);
+
+        var audioContent = JsonUtilities.DefaultDeserialize<ResourceContentAudioJsonSchema>(resourceContentVersion.Content);
+        var audioContentHasSteps = audioContent.Mp3?.Steps?.Any() == true && audioContent.Webm?.Steps?.Any() == true;
```
jwinston-bn, line 57: Should this be `||`? My concern is if there's any existing data or something gets inserted in the future where it doesn't include both for some reason, it's now stuck via CMS.

Also, am I missing it? Where are we handling non-FIA content?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+        // Note: FastEndpoints also supports unbuffered stream processing
+        // (see https://fast-endpoints.com/docs/file-handling#handling-large-file-uploads).
+        // We could switch to that if buffering consumes too many server resources.
+        var tempBlobName = $"uploads/{Guid.NewGuid()}.mp3";
+        await using (var fileStream = request.File.OpenReadStream())
+        {
+            await _blobStorageService.UploadStreamAsync(_uploadOptions.TempStorageContainerName, tempBlobName, fileStream, ct);
+        }
+
+        var uploadEntity = new UploadEntity
```
jwinston-bn, line 101: If one of these gets stuck do we have any way from the database to know what it's in relation to? Any reason not to add `ResourceContentId` to this table, even if it's nullable?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+        else if (audioContentHasSteps)
+        {
+            ThrowError(
+                x => x.StepNumber,
+                "The current Resource Content Version's Content has steps but no step number was passed.",
+                StatusCodes.Status400BadRequest);
+        }
+
+        _logger.LogInformation(
+            "Creating upload for Resource Content ID {ResourceContentId} and Step Number {StepNumber} from file \"{fileName}\".",
```
jwinston-bn, line 87: Does this make sense if `StepNumber` is null?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
jwinston-bn: I guess this is all just checks if it is FIA and then if it doesn't apply it just uploads regularly. I think what throws me off with this is the method is so big it feels like it's a FIA-only method. This is why I like smaller methods, where something like from here to line 84 would be a "ValidateIfContentHasSteps` method or something. Makes it much easier to process what each part is doing and how it fits into the broader functionality.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+    IUploadResourceContentAudioMessagePublisher _uploadResourceContentAudioMessagePublisher,
+    IUserService _userService,
+    ILogger<Endpoint> _logger)
+    : Endpoint<Request, Response>
+{
+    public override void Configure()
+    {
+        Post("/resources/content/{resourceContentId}/uploads");
+        AllowFileUploads();
+        Permissions(PermissionName.CreateContent, PermissionName.CreateCommunityContent);
```
jwinston-bn, line 29: Community reviewers wouldn't do this, it's publisher only, so I think we can just use the `CreateContent` permission or `Roles(Publisher)` Same for `Get` endpoint

src/Aquifer.API/Endpoints/Uploads/Get/Endpoint.cs
```diff
@@ -0,0 +1,36 @@
...
+        Get("/uploads/{uploadId}");
+        Permissions(PermissionName.CreateContent, PermissionName.CreateCommunityContent);
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
+    {
+        var upload = await _dbContext.Uploads
+            .FirstOrDefaultAsync(u => u.Id == req.UploadId, ct);
+
+        if (upload == null)
```
jwinston-bn, line 21:
```suggestion
        if (upload is null)
```

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
jwinston-bn: I'd put `UserId` on it too so we know who initiated it.

jwinston-bn: Probably wouldn't hurt to put a file name on it as well so we know what people are doing. Do we have file size at some point? I'd add that also. I'm just thinking about situations where someone says things aren't working, or we end up with a weird file, it's going to be good to know who did it, what they did, etc.

src/Aquifer.API/Program.cs
```diff
@@ -18,12 +18,19 @@
...
+
 var configuration = builder.Configuration.Get<ConfigurationOptions>() ??
     throw new InvalidOperationException($"Unable to bind {nameof(ConfigurationOptions)}.");
 
 builder.Services
+    .AddOptions<ConfigurationOptions>().Bind(builder.Configuration);
+
+builder.Services
+    .AddSingleton(cfg => cfg.GetRequiredService<IOptions<ConfigurationOptions>>().Value.AzureStorageAccount)
+    .AddSingleton(cfg => cfg.GetRequiredService<IOptions<ConfigurationOptions>>().Value.Upload)
```
jwinston-bn, line 33: Should we move these to a generic `AddOptions` extension method?

src/Aquifer.Common/Configuration/AzureStorageAccountOptions.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Aquifer.Common.Configuration;
+
+public sealed class AzureStorageAccountOptions
+{
+    public required Uri? BlobEndpoint { get; init; }
```
jwinston-bn, line 5: Are these really required if they're nullable? To me a nullable config value would imply it's not required.

src/Aquifer.Common/Services/AzureClientService.cs
```diff
@@ -1,30 +1,41 @@
...
+﻿using Azure.Core;
+using Azure.Identity;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Aquifer.Common.Services;
 
 public interface IAzureClientService
 {
-    DefaultAzureCredential GetCredential();
+    TokenCredential GetCredential();
```
jwinston-bn, line 9: What's the implication of this?

src/Aquifer.Data/Entities/UploadEntity.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Aquifer.Data.EventHandlers;
+
+namespace Aquifer.Data.Entities;
+
+public class UploadEntity : IHasUpdatedTimestamp
```
jwinston-bn, line 6: As mentioned above, I think for auditing purposes it'd be useful to include more data here, such as `ResourceContentId`, `UserId`, `FileName`, `FileSize`, etc. Possibly `ResourceContentVersionId` if we ultimately have that when setting status to completed.

src/Aquifer.Jobs/Program.cs
```diff
@@ -32,14 +32,17 @@
...
-        services.AddOptions<ConfigurationOptions>().Bind(context.Configuration);
-        services.AddSingleton(cfg => cfg.GetService<IOptions<ConfigurationOptions>>()!.Value.Ffmpeg);
-        services.AddSingleton(cfg => cfg.GetService<IOptions<ConfigurationOptions>>()!.Value.OpenAi);
-        services.AddSingleton(cfg => cfg.GetService<IOptions<ConfigurationOptions>>()!.Value.OpenAiTranslation);
-
         var configuration = context.Configuration.Get<ConfigurationOptions>() ??
             throw new InvalidOperationException($"Unable to bind {nameof(ConfigurationOptions)}.");
 
+        services.AddOptions<ConfigurationOptions>().Bind(context.Configuration);
+        services.AddSingleton(cfg => cfg.GetRequiredService<IOptions<ConfigurationOptions>>().Value.AzureStorageAccount);
```
jwinston-bn, line 39: Same as other spot, does it make sense to move all these to an extension method to clean up Program.cs

src/Aquifer.Public.API/Program.cs
```diff
@@ -30,7 +36,8 @@
     .AddFastEndpoints()
     .AddMemoryCache()
     .AddCachingServices()
-    .AddQueueServices(configuration.ConnectionStrings.AzureStorageAccount)
+    .AddSingleton<IBlobStorageService, BlobStorageService>()
+    .AddSingleton<IQueueClientFactory, QueueClientFactory>()
```
jwinston-bn, line 40: Why are these needed here?

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
NateMerritt: > Or would you see it using this same endpoint and splitting on mime type into separate helpers/methods?

I was thinking of doing it based upon resource content media type.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
NateMerritt: I tend to follow the philosophy of returning all 404s before any 400s.  This makes the API less confusing to consume and more secure because it doesn't leak details.

NateMerritt: > Is there a technical reason to restrict this to just mp3? I would allow other types like webm, etc.

I'll make that change.  Do we have a set list of allowed types?  The limit technically is whatever ffmpeg allows (with the Essentials build which doesn't have as many codecs).

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
NateMerritt: We only need one version, the most recent one.  I could get it all in one big query but that would either (a) fetch data we don't need or (b) only have a single version which feels like a lie.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        var uploadEntity = new UploadEntity
+        {
+            Status = UploadStatus.Pending,
+        };
+        _dbContext.Add(uploadEntity);
+        await _dbContext.SaveChangesAsync(ct);
+
+        await _uploadResourceContentAudioMessagePublisher.PublishUploadResourceContentAudioMessageAsync(
```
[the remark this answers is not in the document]
NateMerritt: I thought about having `Faulted` but figured we could add it in the future if it comes up.  As it is the upload would be left `Processing` and have an old `Created` date which allows us to detect them.

src/Aquifer.Common/Services/AzureClientService.cs
NateMerritt: `TokenCredential` is the base class of `DefaultAzureCredential` so no implications.  All code should work as it did previously.

src/Aquifer.Public.API/Program.cs
NateMerritt: I got rid of the `AddQueueServices` method because it only had one thing left in it.  I also made the PublicApi consistent in using the same azure storage `appsettings.json` properties even though it didn't technically need an update.

src/Aquifer.Jobs/Program.cs
NateMerritt: In general I find distributed registries harder to read/understand because you have to jump around between files to see all of the dependencies.  An exception is if the types are truly isolated from one another or there's complexity that is needed by multiple apps.  This setting is used by multiple classes (queues and blobs) so it didn't really make sense to move it to a queue or blob specific extension.

General philosophy: I'd be happy to move _all_ service registrations out of `Program.cs` into a separate file just for service registrations for each app (named `ServiceRegistry` or something like that).

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
jwinston-bn: But do you even need ResourceContent? Why can't you just grab this RCV and check if `x => x.ResourceContent.MediaType == ...` and then check if its null? And if you need the RC, just do `.Include(x => x.ResourceContent)`.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
NateMerritt: It will log something like the following:

> Creating upload for Resource Content ID 306295 and Step Number (null) from file "ENG_FIAKeyTerms_angel-of-the-lord.webm".

That seems reasonable to me.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
NateMerritt: I'll pull the validation out into a separate method.

src/Aquifer.API/Endpoints/Resources/Content/Uploads/Create/Endpoint.cs
```diff
@@ -0,0 +1,126 @@
...
+        await using (var fileStream = request.File.OpenReadStream())
+        {
+            await _blobStorageService.UploadStreamAsync(_uploadOptions.TempStorageContainerName, tempBlobName, fileStream, ct);
+        }
+
+        var uploadEntity = new UploadEntity
+        {
+            Status = UploadStatus.Pending,
+        };
+        _dbContext.Add(uploadEntity);
```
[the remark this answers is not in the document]
NateMerritt: Thanks, I missed that.  I'm going with the sync version, though, because we shouldn't need to make a DB call as part of the Add so async is unnecessary:
```
_dbContext.Uploads.Add(uploadEntity);
```

src/Aquifer.API/Program.cs
NateMerritt: See [comment below](https://github.com/BiblioNexusStudio/aquifer-server/pull/694/files#r1992027239).

jwinston-bn (approved):

src/Aquifer.Common/Messages/QueueClientFactory.cs
```diff
@@ -52,13 +53,15 @@ public async Task<QueueClient> GetQueueClientAsync(string queueName, Cancellatio
     // Note: This method creates the queue if it doesn't yet exist.
     private async Task<QueueClient> GetQueueClientCoreAsync(string queueName, CancellationToken ct)
     {
-        var client = _queueConfigurationOptions.AzureQueueStorageConnectionString.StartsWith("http")
+        var client = string.IsNullOrEmpty(_azureStorageAccountOptions.ConnectionStringOverride)
             ? new QueueClient(
-                new Uri($"{_queueConfigurationOptions.AzureQueueStorageConnectionString}/{queueName}"),
```
NateMerritt, line 57: This was a regression.  Fixed here: https://github.com/BiblioNexusStudio/aquifer-server/pull/711

PR #695, opened 2025-03-10
Bib 853 versification UI updates

kaseywright:


src/Aquifer.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -75,38 +78,158 @@ private static VerseReference FormatBaseBookChapterVerseMapping(int mappedVerseI
...
+                        BookName = response.BookName,
+                        ChapterNumber = targetChapter,
+                        VerseNumber = targetVerse
+                    },
+                    SourceBibleVerse = mapping.Value
+                });
+            }
+        }
+        // catch start verse {VerseId} is invalid for bible {BibleId}
+        catch (ArgumentException e)
```
kaseywright, line 137: When we request a single verse, like Jonah 1:17, with the request parameters all set as:
```
startChapter: 1
EndChapter: 1
startVerse: 17
endVerse: 17
```
The `ArgumentException` can be caught and we can swap the bibles to get the versification and then just add it to the response. 

But this feels like it could be better. This implementation is a WIP and was written to validate that swapping does work.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,166 @@
...
+            (minVerseId, maxVerseId) = await GetMinAndMaxVerseIdFromRequestAsync(req, cancellationToken);
+        }
+        catch (ArgumentException e)
+        {
+            AddError(e.Message);
+            await SendErrorsAsync(cancellation: cancellationToken, statusCode: 400);
+            return;
+        }
+        
+        // todo: error handling for versification
```
kaseywright, line 36: When a user requests a chapter or verse that does not exist in the requested book, versification will fail. Should we either:

- validate that the requested chapters and verses are valid and return the versification for the valid passage range
- or return a 400 and make the client handle the error?

For example, a request is made for Jonah chapters 1-5 or Jonah verses 1-27, versification will fail. Should we just return the valid versification for Jonah - which only has 4 chapters?

src/Aquifer.API/Endpoints/Bibles/Versification/Response.cs
```diff
@@ -0,0 +1,20 @@
+namespace Aquifer.API.Endpoints.Bibles.Versification;
+
+public class Response
+{
+    public IEnumerable<VerseMapping> VerseMappings { get; set; } = [];
+}
+
+public class VerseReference
+{
+    public required int VerseId { get; set; }
```
kaseywright, line 10: I had to rename these properties because the CMS has both a `Verse` and `VerseReference` type that are very close to what I am returning here. Adding yet another similar type would make the CMS logic very confusing as these types are used in the same locations as a new type would be.

src/Aquifer.API/Endpoints/Bibles/Versification/Request.cs
```diff
@@ -0,0 +1,11 @@
+namespace Aquifer.API.Endpoints.Bibles.Versification;
+
+public class Request
+{
+    public int TargetBibleId { get; set; }
```
NateMerritt, line 5: Note for future work: We may want `SourceBibleId` in order to do arbitrary mappings but omitting is fine for now.

src/Aquifer.API/Endpoints/Bibles/Versification/Response.cs
NateMerritt: 👍  If it causes issues on the client end to have slightly differing types then we may want to move this class into a shared location so that all endpoints can use it.

src/Aquifer.API/Endpoints/Bibles/Versification/Response.cs
```diff
@@ -0,0 +1,20 @@
...
+    public required string Book { get; set; }
+    public required int Chapter { get; set; }
+    public required int Verse { get; set; }
+}
+
+public class VerseMapping
+{
+    public required VerseReference TargetBibleVerse { get; set; }
+    public required VerseReference SourceBibleVerse { get; set; }
```
NateMerritt, line 19: None of the other response properties use "Bible" in the names so I suggest removing them here, too, for consistency.
```suggestion
    public required VerseReference TargetVerse { get; set; }
    public required VerseReference SourceVerse { get; set; }
```

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+        
+        // We have ensured that the values returned here will always be valid 
+        var (minVerseId, maxVerseId) = await GetMinAndMaxVerseIdFromRequestAsync(req, cancellationToken);
+        
+        var versificationMap = await VersificationUtilities.ConvertVersificationRangeAsync(
+            CachingVersificationService.EngVersificationSchemeBibleId,
+            minVerseId,
+            maxVerseId,
+            req.TargetBibleId,
```
NateMerritt, line 36: Missing validation: There should be a 404 response returned above if the `TargetBibleId` doesn't match an existing Bible.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+            verseMappings.Add(new VerseMapping
+            {
+                SourceBibleVerse = new VerseReference
+                {
+                    VerseId = mapping.Key,
+                    Book = BibleBookCodeUtilities.FullNameFromId(sourceBookId),
+                    Chapter = sourceChapter,
+                    Verse = sourceVerse
+                },
+                TargetBibleVerse = mapping.Value
```
NateMerritt, line 60: We may need to return exclusions in the future even though it's not required for this issue.  I suggest making `TargetBibleVerse` nullable and having a `null` value represent that the target does not have that verse.

It's not too hard to also return `null` for exclusions here so I'd probably just do that now so that this endpoint doesn't have to be updated in the future if/when we need that.

src/Aquifer.API/Endpoints/Bibles/Versification/Response.cs
kaseywright: I agree. I was tempted to do this with this ticket but was afraid that I'd get bogged down in the refactor.

src/Aquifer.API/Endpoints/Bibles/Versification/Response.cs
```diff
@@ -0,0 +1,20 @@
+namespace Aquifer.API.Endpoints.Bibles.Versification;
+
+public class Response
+{
+    public IEnumerable<VerseMapping> VerseMappings { get; set; } = [];
```
NateMerritt, line 5:
```suggestion
    public required IReadOnlyList<VerseMapping> VerseMappings { get; set; }
```

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+            VerseId = mappedVerseId,
+            Book = targetBookName,
+            Chapter = targetChapter,
+            Verse = targetVerse
+        };
+    }
+
+    private static bool IsRequestedBookIdValid(int bookId)
+    {
+        return Enum.IsDefined(typeof(BookId), bookId);
```
NateMerritt, line 176: This will return true if `BookId.None` (`0`) is passed.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private async Task<bool> IsChapterValidForBookAsync(int chapterNumber, int bookId, CancellationToken ct)
+    {
+        return await dbContext.BookChapters
```
NateMerritt, line 181: FYI, `BookChapters` only corresponds to Bible ID `0` (`eng` versification).  This is correct for now but if we ever want to pass source Bible ID then this will be wrong.  It would be better to check `BibleTexts`.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
kaseywright: The Validator catches a bookId of `0`

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
NateMerritt: Hold off on changing things, though.  I have an idea and am writing up another comment.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+    {
+        if (!IsRequestedBookIdValid(req.BookId))
+        {
+            AddError($"Invalid book id: {req.BookId}.");
+            await SendErrorsAsync(cancellation: cancellationToken, statusCode: 400);
+            return;
+        }
+        
+        // We have ensured that the values returned here will always be valid 
+        var (minVerseId, maxVerseId) = await GetMinAndMaxVerseIdFromRequestAsync(req, cancellationToken);
```
NateMerritt, line 37: I'm also concerned about the number of DB calls being made here:
1. Start chapter is valid (or fetch start chapter).
2. End chapter is valid (or fetch end chapter).
3. Start verse is valid (or fetch start verse of start chapter).
4. End verse is valid (or fetch end verse of end chapter).

The good news is that the `CachingVersificationService` does all of this work already and it's available from the cache via [GetMaxChapterNumberAndVerseNumbersByBookIdMapAsync](https://github.com/BiblioNexusStudio/aquifer-server/blob/a825b4b8568fb1346e80870ec588bbddc9700569/src/Aquifer.Common/Services/Caching/CachingVersificationService.cs#L17).  That doesn't have the minimum chapter number or minimum verse number, though.  We could either add it to that caching service as a similar method or you could do something like:
```
bool isValid = false;
int startChapter = 1;
int startVerse = 1;
while (!isValid)
{
    isValid = await IsValidVerseIdAsync(...);
}
```
using [IsValidVerseIdAsync](https://github.com/BiblioNexusStudio/aquifer-server/blob/a825b4b8568fb1346e80870ec588bbddc9700569/src/Aquifer.Common/Utilities/VersificationUtilities.cs#L12).  In almost every case it will be 1,1 or maybe 1,2 so there's not going to be big performance hit.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            return (BibleUtilities.GetVerseId((BookId)req.BookId, minChapterNumber, minVerse),
+                    BibleUtilities.GetVerseId((BookId)req.BookId, maxChapterNumber, req.EndVerse!.Value));
+        }
+        
+        return (BibleUtilities.GetVerseId((BookId)req.BookId, minChapterNumber, minVerse),
+                BibleUtilities.GetVerseId((BookId)req.BookId, maxChapterNumber, maxVerse));
+    }
+    
+    private static VerseReference FormatBaseBookChapterVerseMapping(int mappedVerseId)
```
NateMerritt, line 167: Nit: I'd probably name this with a `Map` verb.  If you disagree that's fine.
```suggestion
    private static VerseReference MapToVerseReference(int verseId)
```

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -0,0 +1,192 @@
...
+    {
+        var (targetBookId, targetChapter, targetVerse) = BibleUtilities.TranslateVerseId(mappedVerseId);
+        var targetBookName = BibleBookCodeUtilities.FullNameFromId(targetBookId);
```
NateMerritt, line 170: Nit: The `target` prefix isn't needed here and below.

src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
kaseywright: This is good. that format name was a holdover from the previous iteration when returning a string

NateMerritt (approved):

PR #696, opened 2025-03-11
Add health endpoint to Jobs

jwinston-bn:


Directory.Packages.props
```diff
@@ -1,63 +1,64 @@
 <Project>
-  <PropertyGroup>
-    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
-  </PropertyGroup>
-  <ItemGroup>
-    <PackageVersion Include="Auth0.AuthenticationApi" Version="7.33.0" />
```
NateMerritt, line 7: Every XML file in the solution is currently formatted with ` />` (includes a space) for self-closing elements, not `/>` (no space).  Do you have an objection to that?  If not, then it would be best to update the code formatter to follow that convention.  If so, then a separate PR that updates all files would be best.

It looks like it might be a Rider only setting: https://stackoverflow.com/a/78043398

jwinston-bn: I don't care. I didn't add it manually.

NateMerritt: > I didn't add it manually.

Whitespace tooling fights are no good for source control diffs and blames.  I just want to get a consistent setting so that only the necessary lines are changed.  If we don't then the next time that I add a NuGet file via Visual Studio's Nuget Package UI all of the whitespace will revert back and be a huge diff again.

PR #699, opened 2025-03-12
Add AllowedReports to Reports Table

elisabeth-shakespeare:
An column that will be a comma-separated list of roles so that we can know which reports can be seen by which roles.


src/Aquifer.Data/AquiferDbContext.cs
```diff
@@ -40,6 +40,7 @@ public AquiferDbContext(DbContextOptions options, bool isReadOnly) : base(option
 
     public bool IsReadOnly { get; }
 
+    public DbSet<AllowedReportRolesEntity> AllowedReportRoles { get; set; }
```
jwinston-bn, line 43: I wasn't expecting us to create a whole new table for this. Figured it'd just be a new column on the existing Reports table.

elisabeth-shakespeare: Since it's going to be multiples we'd have to store that either as json or add some sort of functionality to DBContext that can parse the array into a comma-separated string

jwinston-bn: I figured we'd just have a comma separated list in the column and then you could just grab records that contain that role id. Maybe down the road we could have a more robust solution, but I don't think there's any reason to create a whole new table for this.

jwinston-bn (approved):

PR #701, opened 2025-03-14
Support read-only DB connections for Bible Well.

NateMerritt:
This still needs GHA support for deployment to India.


NateMerritt:
Closing because we're going to move forward with a new API to back the new mobile application.

PR #702, opened 2025-03-14
Update Resources Content List endpoint to support N/A status filtering.

NateMerritt:
Front-end PR: https://github.com/BiblioNexusStudio/content-manager-web/pull/598


src/Aquifer.API/Endpoints/Resources/Content/List/Endpoint.cs
```diff
@@ -33,9 +37,13 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
                 ResourceEnglishLabelQuery = req.SearchQuery,
                 LanguageId = req.LanguageId,
                 ExcludeContentMediaTypes = [ResourceContentMediaType.Audio],
-                ExcludeContentStatuses = [ResourceContentStatus.TranslationNotApplicable, ResourceContentStatus.CompleteNotApplicable],
-                IsPublished = req.IsPublished.HasValue && req.IsPublished.Value ? true : null,
-                IsDraft = req.IsPublished.HasValue && !req.IsPublished.Value ? true : null,
+                ExcludeContentStatuses = req.IsNotApplicable.HasValue && req.IsNotApplicable.Value
```
jwinston-bn, line 40:
```suggestion
                req.IsNotApplicable == true
```

src/Aquifer.API/Endpoints/Resources/Content/List/Request.cs
```diff
@@ -10,6 +10,7 @@ public record Request
     public string? BookCode { get; set; }
     public int? StartChapter { get; set; }
     public int? EndChapter { get; set; }
+    public bool? IsNotApplicable { get; set; }
     public bool? IsPublished { get; set; }
```
jwinston-bn, line 14: `IsPublished` was already here to set the precedent, but since it's being updated anyway I wonder if these should get better names like `ListNotApplicableOnly` or something.

NateMerritt: I could instead expose an include and an exclude status filter.  This seemed to go along with the "expose what the front-end needs" philosophy which is why I used the `boolean`.

jwinston-bn: Yeah, I wasn't suggesting changing the boolean, just renaming the boolean to be more explicit about what it's doing. `IsNotApplicable` doesn't really indicate that it's _only_ going to return Not Applicable items.

NateMerritt: As a search request I think "only" is not necessary as then we'd have `ListOnlyLanguageId`, `ListOnlyParentResourceId`, etc.  It's pretty clear that it's an optional filter.

jwinston-bn (approved):

PR #703, opened 2025-03-14
Bib 869 create a reporting page for managers

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Reports/Dynamic/Get/Endpoint.cs
```diff
@@ -24,6 +25,13 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
             return;
         }
 
+        var user = await userService.GetUserFromJwtAsync(ct);
```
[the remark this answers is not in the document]
elisabeth-shakespeare: Just need it to get the user's current role.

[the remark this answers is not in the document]
elisabeth-shakespeare: Do you know if that returns the number representations of the roles?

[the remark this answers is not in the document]
jwinston-bn: Seems to be a string. You could add something to convert it to our enums. We only support a single role anyway, so it'd probably be helpful to have one that gets that list of a single string and converts it to the right UserRole enum.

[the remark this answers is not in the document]
elisabeth-shakespeare: Good point. I'll add in something for that

src/Aquifer.API/Services/UserService.cs
```diff
@@ -42,6 +43,18 @@ public List<string> GetAllJwtRoles()
         return httpContextAccessor.HttpContext?.User.FindAll(Permissions.RolesClaim).Select(c => c.Value).ToList() ?? [];
     }
 
+    public UserRole GetUserRoleFromJwt()
```
jwinston-bn, line 46: I think we need to be a bit more defensive here. If a user wasn't set correctly and had no role, they would blow up here. We need to return `UserRole?` at least and verify that `GetAllJwtRoles.Count > 0`.

src/Aquifer.API/Services/UserService.cs
```diff
@@ -42,6 +43,18 @@ public List<string> GetAllJwtRoles()
...
+    {
+        var role = GetAllJwtRoles()[0];
+        var isParsed = Enum.TryParse(string.Concat(role[0].ToString().ToUpper(), role.AsSpan(1)), out UserRole userRole);
+        if (!isParsed)
+        {
+            throw new InvalidCastException($"Role {role} is not a valid role");
```
jwinston-bn, line 52: Again, I wouldn't blow up the app over this. I would return null, and then the caller can address it as needed that there is no role (presumably sending back a 403 or something).

src/Aquifer.API/Endpoints/Reports/Dynamic/List/Endpoint.cs
```diff
@@ -1,25 +1,48 @@
...
+                    r.Name,
+                    r.Description,
+                    r.Type,
+                    r.AllowedRoles,
+                    r.ShowInDropdown
+                })
             .ToListAsync(ct);
 
+        var reports = reportsWithAllowedRoles.Where(
+                r => ReportRoleHelper.RoleIsAllowedForReport(r.AllowedRoles, userService.GetUserRoleFromJwt()))
```
jwinston-bn, line 34: This over and over and over again gets the user role from the JWT. Why not just grab it once?

src/Aquifer.API/Services/UserService.cs
```diff
@@ -42,6 +43,18 @@ public List<string> GetAllJwtRoles()
         return httpContextAccessor.HttpContext?.User.FindAll(Permissions.RolesClaim).Select(c => c.Value).ToList() ?? [];
     }
 
+    public UserRole GetUserRoleFromJwt()
+    {
+        var role = GetAllJwtRoles()[0];
+        var isParsed = Enum.TryParse(string.Concat(role[0].ToString().ToUpper(), role.AsSpan(1)), out UserRole userRole);
```
jwinston-bn, line 49: I find this pretty confusing. What's the reasoning to do all this instead of something simpler like this?
```csharp
Enum.TryParse(role, true, out UserRole userRole);
```

elisabeth-shakespeare: Because what comes out of the jwt is all lowercase, and the enum has the first letter uppercase

elisabeth-shakespeare: so a jwt role like "admin" will not match `UserRole` "Admin"

jwinston-bn: Yes, what I posted solves that. The `true` in there is for ignore case.

elisabeth-shakespeare: ....... oh. the boolean is for case sensitivity. Changing...

src/Aquifer.API/Services/UserService.cs
```diff
@@ -42,6 +43,20 @@ public List<string> GetAllJwtRoles()
         return httpContextAccessor.HttpContext?.User.FindAll(Permissions.RolesClaim).Select(c => c.Value).ToList() ?? [];
     }
 
+    public UserRole GetUserRoleFromJwt()
+    {
+        var jwtRoles = GetAllJwtRoles();
+
+        if (jwtRoles.Count <= 0)
```
jwinston-bn, line 50: I think we can just check if it's `== 0` here. It can't be negative.

src/Aquifer.API/Services/UserService.cs
```diff
@@ -42,6 +43,20 @@ public List<string> GetAllJwtRoles()
...
+        {
+            return UserRole.None;
+        }
+
+        var role = jwtRoles[0];
+        var isParsed = Enum.TryParse(role, true, out UserRole userRole);
+        return !isParsed ? UserRole.None : userRole;
```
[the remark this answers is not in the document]
jwinston-bn: In fact, `null` will out `None` also, so you can get rid of the previous check too.
```csharp
Enum.TryParse(jwtRoles.FirstOrDefault(), true, out UserRole userRole);
return userRole;
```

PR #714, opened 2025-03-18
Add route to clear caches to Aquifer.API and Aquifer.PublicAPI.

NateMerritt:
I tested and both cache clearing scenarios work as expected on both endpoints.  I also confirmed that the swagger docs do not include the new endpoint and no client code is generated for it in the Public API.  The endpoints are gated by a new Admin ApiKeyScope which, in addition to everything granted by `All`, also grants access to admin routes.

### DIFF
--- a/src/Aquifer.API/Telemetry/RequestTelemetryInitializer.cs
+++ b/src/Aquifer.API/Telemetry/RequestTelemetryInitializer.cs
@@ -26,7 +26,7 @@ public class RequestTelemetryInitializer(IHttpContextAccessor httpContextAccesso
 
         if ((httpContextAccessor.HttpContext?.Items.TryGetValue(Constants.HttpContextItemCachedApiKey, out var maybeCachedApiKey) ??
                 false) &&
-            maybeCachedApiKey is CachedApiKey cachedApiKey)
+            maybeCachedApiKey is ApiKey cachedApiKey)
         {
             requestTelemetry.Properties[Constants.TelemetryBnApiCallerIdPropertyName] = cachedApiKey.Id.ToString();
             requestTelemetry.Properties[Constants.TelemetryBnApiCallerPropertyName] = cachedApiKey.Organization;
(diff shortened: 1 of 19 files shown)


src/Aquifer.API/Telemetry/RequestTelemetryInitializer.cs
```diff
@@ -26,7 +26,7 @@ public void Initialize(ITelemetry telemetry)
 
         if ((httpContextAccessor.HttpContext?.Items.TryGetValue(Constants.HttpContextItemCachedApiKey, out var maybeCachedApiKey) ??
```
NateMerritt, line 27: I didn't update this to use the new property on `CachingApiKeyService` because this is probably the one time where the api-key might be null.

src/Aquifer.Common/Messages/Publishers/ResourceContentRequestTrackingMessagePublisher.cs
```diff
@@ -71,13 +72,12 @@ public async Task PublishTrackResourceContentRequestMessageAsync(
...
 
         var sourceHeader = httpContext.Request.Headers["bn-source"].ToString();
-        httpContext.Items.TryGetValue(Constants.HttpContextItemCachedApiKey, out var maybeCachedApiKey);
 
         var message = new TrackResourceContentRequestMessage
         {
             Source = string.IsNullOrEmpty(sourceHeader) ? source : sourceHeader,
             IpAddress = GetClientIp(httpContext),
-            SubscriptionName = maybeCachedApiKey is CachedApiKey cachedApiKey ? cachedApiKey.Organization : null,
+            SubscriptionName = _cachingApiKeyService.CurrentApiKey.Organization,
```
NateMerritt, line 80: I'm assuming that the API Key can never be null here.  Any access to resource data (via an endpoint) requires an API Key.  Jobs doesn't publish these.

src/Aquifer.Common/Services/Caching/CachingApiKeyService.cs
```diff
@@ -1,51 +1,73 @@
...
+public sealed class ApiKey(int _id, string _key, ApiKeyScope _scope, string? _organization)
+{
+    public int Id { get; } = _id;
+    public string Key { get; } = _key;
+    public ApiKeyScope Scope { get; } = _scope;
+    public string? Organization { get; } = _organization;
+
+    public bool HasScope(ApiKeyScope scope)
+    {
+        return scope is ApiKeyScope.Admin or ApiKeyScope.None
```
NateMerritt, line 69: I wouldn't expect `All` or `Admin` to match a scope of `None` so I put an exception for it in here.

jwinston-bn (approved):

PR #715, opened 2025-03-20
Handle min verses and correctly return the max verse number for Bible chapters to the Bible Well.

NateMerritt:
1. Cache the min verse ID for each chapter.  For now it's always 1 based upon DB data but this paves the way for having verse 0 data for Psalms with superscripts. As part of this I factored out some of the new Versifications endpoint into a shared method in the VersificationsUtility.
2. Use the `CachingVersificationService` to populate the max verse ID per chapter for display in the Bible Well's verse selector.  This fixes a bunch of bugs where ENG's max verse number (which was always previously returned) was used instead of the actual Bible's max verse numbers:
    * Psalm 3:9 cannot currently be selected in the RSB because ENG only has up to 3:8.
    * Revelation 12:18 is shown in the verse picker for the BSB [but displays nothing](https://qa.well.bible/view-content/-/REV012018-012018?tab=Bible).  Since it is the last verse in the chapter it will now be omitted from the verse picker.
    * Note the the verse picker in the Well appears to use the first Bible selected in order to display the verse picker.
    * Excluded verses in the middle of a chapter will still be shown in the picker for the given Bible.

### DIFF
--- a/src/Aquifer.API/Endpoints/Bibles/Book/List/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Bibles/Book/List/Endpoint.cs
@@ -1,11 +1,14 @@
 ﻿using Aquifer.API.Helpers;
+using Aquifer.Common.Services.Caching;
+using Aquifer.Common.Utilities;
 using Aquifer.Data;
 using FastEndpoints;
 using Microsoft.EntityFrameworkCore;
 
 namespace Aquifer.API.Endpoints.Bibles.Book.List;
 
-public class Endpoint(AquiferDbContext dbContext) : Endpoint<Request, List<Response>>
+public class Endpoint(AquiferDbContext _dbContext, ICachingVersificationService _cachingVersificationService)
+    : Endpoint<Request, IReadOnlyList<Response>>
 {
     public override void Configure()
     {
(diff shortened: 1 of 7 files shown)


src/Aquifer.API/Endpoints/Bibles/Book/List/Endpoint.cs
```diff
@@ -16,23 +19,41 @@ public override void Configure()
...
+                    TotalChapters = x.ChapterCount,
+                    Chapters = maxChapterNumberAndBookendVerseNumbersByBookIdMap
+                            .GetValueOrNull(x.BookId)
+                            ?.BookendVerseNumbersByChapterNumberMap
+                            .OrderBy(kvp => kvp.Key)
+                            .Select(
+                                kvp => new ResponseChapter
+                                {
+                                    Number = kvp.Key,
+                                    TotalVerses = kvp.Value.MaxVerseNumber,
```
NateMerritt, line 51: I'm thinking we might need to add `MinVerseNumber` here and update Well so that it displays verse 0 in the picker.  And probably rename `TotalVerses` to `MaxVerseNumber` while we're at it.

src/Aquifer.Data/Migrations/20250320163256_RemoveChaptersAssociation.cs
```diff
@@ -0,0 +1,22 @@
...
+
+#nullable disable
+
+namespace Aquifer.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class RemoveChaptersAssociation : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
```
NateMerritt, line 11: Is this migration needed?  The snapshot updated but I'm not sure that matters.

jwinston-bn: Wouldn't think so since there's no actual DB change

jwinston-bn (approved):

jwinston-bn (approved):

PR #717, opened 2025-03-20
Support passing 0 as verse number.

NateMerritt:
I audited all requests and validators with `verse` in them and the other ones should already work correctly.


src/Aquifer.Common/Utilities/BibleUtilities.cs
```diff
@@ -110,26 +110,42 @@ public static (int StartVerseId, int EndVerseId)? VerseRangeForBookAndChapters(B
         return (LowerBoundOfBook(bookId), UpperBoundOfBook(bookId));
     }
 
-    public static (int StartVerseId, int EndVerseId) GetVerseIds(string bookCode,
```
NateMerritt, line 113: Only endpoints in this PR use this method.

PR #719, opened 2025-03-21
Sli columns on Parent Resources

jwinston-bn:


src/Aquifer.Data/Entities/ParentResourceEntity.cs
```diff
@@ -22,6 +22,11 @@ public class ParentResourceEntity : IHasUpdatedTimestamp
     public bool? ForMarketing { get; set; }
     public bool AllowCommunityReview { get; set; }
 
+    [MaxLength(128)]
+    public string? SliCategory { get; set; }
```
NateMerritt, line 26: Are these going to live in code as an enum?  `string` is surprising.

jwinston-bn: No, they come from a third party and can change randomly. The intention is to include it in the public API when listing resource categories. Don't want to have to deploy a code change to update the name. Originally thought about doing a join table but decided this was simpler for now.

PR #721, opened 2025-03-24
add column for whether to show report in dropdown

elisabeth-shakespeare:


elisabeth-shakespeare:
Wanted thoughts on the naming of this column. Once the name is nailed down I'll run the migration and take this pr out of draft status

jwinston-bn (approved):

PR #722, opened 2025-03-24
Support one-to-many versification mapping.

NateMerritt:
* Consider verse parts (data was already in the DB) during versification mapping, which allows a single verse to map to more than one verse.  This update resulted in a breaking change for the versifications endpoint which is [addressed in the CMS]().
* Add `MinVerseNumber` to `BookChapters` which allows the `ENG` data to know when there is a verse 0 (Psalm superscript).

Example of where this allows two verses in ENG (verse 0 for superscript and standard verse 1) to map to a single verse (verse 1 only which includes superscript) in Russian:
image

Example of new multi-mapping in Russian from existing Nehemiah data:
image

### DIFF
--- a/src/Aquifer.Common/Utilities/VersificationUtilities.cs
+++ b/src/Aquifer.Common/Utilities/VersificationUtilities.cs
@@ -48,7 +48,7 @@ public static class VersificationUtilities
 
         // Edge Case: It's theoretically possible that a Bible could be missing an entire chapter.
         if (!maxChapterNumberAndBookendVerseNumbers.BookendVerseNumbersByChapterNumberMap
-                .TryGetValue(verse.chapter, out var bookendVerseNumbers))
+            .TryGetValue(verse.chapter, out var bookendVerseNumbers))
         {
             return false;
         }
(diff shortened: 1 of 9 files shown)


src/Aquifer.Common/Utilities/VersificationUtilities.cs
```diff
@@ -271,64 +360,90 @@ public static async Task<IReadOnlyList<int>> ExpandVerseIdRangeAsync(
...
+                                    : $"{baseVerseId}a")
+                            ?? [int.Parse(baseVerseId)];
+
+                    // Handle two scenarios:
+                    // 1. The target verse ID is in the exclusions list.  In this case omit it.
+                    // 2. The target verse ID is not in a valid book for the target Bible even though it exists in the base mapping.
+                    //    In this case omit it.
+                    // 3. The target verse ID is not in a valid chapter/verse range for the target Bible/book
+                    //    (e.g. it's "excluded" but at the very end of a chapter so it's not in the exclusions list).
+                    //    In this case constrain the verse to the nearest valid verse.
```
NateMerritt, line 427: This comment describes an existing bug fixed by this PR.

src/Aquifer.Data/Migrations/20250324175904_AddMinVerseNumberToBookChapters.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<int>(
+                name: "MinVerseNumber",
+                table: "BookChapters",
+                type: "int",
+                nullable: false,
+                defaultValue: 1);
```
NateMerritt, line 18: Manually edited to `1` instead of `0`.

jwinston-bn (approved):

kaseywright (approved):

PR #724, opened 2025-03-27
Update comment endpoints to support user mentions.

NateMerritt:
I didn't add validation that user IDs in mentions are legit; if it does happen then the DB will explode due to an invalid FK and the server will return a 500.  There aren't any scenarios where this should happen in practice, though.

### DIFF
--- a/src/Aquifer.Data/Entities/CommentMentionEntity.cs
+++ b/src/Aquifer.Data/Entities/CommentMentionEntity.cs
@@ -1,0 +1,36 @@
+﻿using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace Aquifer.Data.Entities;
+
+[EntityTypeConfiguration(typeof(CommentMentionEntityConfiguration))]
+public sealed class CommentMentionEntity
+{
+    public int Id { get; set; }
+
+    public int CommentId { get; set; }
+    public CommentEntity Comment { get; set; } = null!;
+
+    public int UserId { get; set; }
+    public UserEntity User { get; set; } = null!;
+
+    [SqlDefaultValue("getutcdate()")]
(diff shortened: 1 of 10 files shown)


src/Aquifer.Data/Entities/CommentMentionEntity.cs
```diff
@@ -0,0 +1,36 @@
...
+    public CommentEntity Comment { get; set; } = null!;
+
+    public int UserId { get; set; }
+    public UserEntity User { get; set; } = null!;
+
+    [SqlDefaultValue("getutcdate()")]
+    public DateTime Created { get; set; }
+
+    [SqlDefaultValue("getutcdate()")]
+    public DateTime Updated { get; set; }
```
jwinston-bn, line 21: I like to add `= DateTime.UtcNow()` onto these so we don't accidentally create them with default dates.

src/Aquifer.Data/Services/CommentMentionsUtility.cs
jwinston-bn, line 1: I tend to think of services as classes that get registered and injected, and that's generally how we've treated them in this code base. I think this belongs in a Utilities folder similar to how we've broken things like this out in other projects.

src/Aquifer.Data/Services/CommentMentionsUtility.cs
```diff
@@ -0,0 +1,49 @@
...
+        return MentionsRegex()
+            .Matches(text)
+            .Where(m => m.Success)
+            .Select(m => int.Parse(m.Groups["userId"].Value))
+            .ToHashSet();
+    }
+
+    // examples:
+    // {@123|John Doe}
+    // {@98342|X Æ A-12}
```
jwinston-bn, line 46: small-smile

jwinston-bn (approved):

PR #725, opened 2025-03-27
Bib 852 show english source for retranslate drafts

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -182,6 +182,31 @@ await dbContext.ResourceContentVersionAssignedUserHistory
                 })
             .ToListAsync(ct);
 
+        if (resourceContent.Language.Id != 1 && versions.Any(v => v.IsPublished) && relevantContentVersion.IsDraft)
```
jwinston-bn, line 185: I don't think this logic is quite right. A draft could exist without something being published. Really the issue is if there's a version > 1. Two things I would change here:

Avoid drilling into an additional object:
`resourceContent.Language.Id != 1` -> `resourceContent.LanguageId !=1`

Check that the current version is greater than 1:
`versions.Any(v => v.IsPublished)` -> `relevantContentVersion.Version > 1`

src/Aquifer.API/Endpoints/Resources/Content/Get/Endpoint.cs
```diff
@@ -182,6 +182,31 @@ await dbContext.ResourceContentVersionAssignedUserHistory
...
+        {
+            var englishSourceVersion =
+                await dbContext.ResourceContentVersions
```
jwinston-bn, line 188: It gets weird, but this isn't really right either. The currently published English version isn't necessarily the source of the original translated version. We need the original English snapshot for the translated version. Since you already have `Versions`, I think you grab the `ResourceContentVersionId` of `Version = 1`, and then grab the first snapshot for that `ResourceContentVersionId`.

elisabeth-shakespeare: Does it matter if the original translated version is translated from a different version of English? I'm thinking a scenario like this:
- An English resource is Aquiferized, then reworked and Aquiferized again
- The latest Aquiferized English version is translated.

jwinston-bn: That's what I'm saying. The original snapshot for a translated version is the source English. We can't look at the English versions because we don't know which one the original was based off. You need something like

```csharp
var firstVersion = versions.FirstOrDefault();
if (firstVersion is not null)
{
    var firstVersionSnapshot = await dbContext.ResourceContentVersionSnapshots
        .OrderBy(x => x.Created)
        .FirstOrDefaultAsync(x => x.ResourceContentVersionId == firstVersion.Id, ct);
}
```

elisabeth-shakespeare: The versions here are for the `relevantContentVersion.ResourceId`, not the English. That would be another db hit to get the English ones

elisabeth-shakespeare: Nope. never mind. I get it now

jwinston-bn (approved):

PR #729, opened 2025-03-28
Fix error when marking comment notifications as read.

NateMerritt:


src/Aquifer.API/Endpoints/Notifications/Update/Endpoint.cs
```diff
@@ -54,8 +54,11 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
             // Note that only notifications that a user has interacted with are stored in the DB.
             var existingNotification = await _dbContext.Notifications
                 .AsTracking()
-                .Where(n => n.Kind == req.NotificationKind && n.NotificationKindId == req.NotificationKindId)
-                .FirstOrDefaultAsync(ct);
+                .Where(n =>
+                    n.UserId == currentUserId &&
```
NateMerritt, line 58: The absence of this check has been a bug since notifications shipped.  I'm surprised nobody reported that marking as read wasn't working.

PR #733, opened 2025-04-02
Fix versifications

NateMerritt:
### Fix Bug

The CMS currently tries to fetch verse 0 mappings so it can handle Psalm superscripts correctly.  Unfortunately, there is no current way to distinguish between when (a) there are no mappings for verse 0 from source -> target and (b) verse 0 doesn't exist in the source Bible.  When this happens the front-end can map to a start verse ID that is greater than an end verse ID:
* `Unexpected versification mapping: sourceStartVerseId: 1019043000; sourceEndVerseId: 1019043001; targetBibleId: 11; targetStartVerseId: 1019043000; targetEndVerseId: 1019042001; languageId: 10`
* [Thread](https://biblionexusworkspace.slack.com/archives/C0662HZ7RKM/p1743600233099929)
* [Repro](https://qa.admin.aquifer.bible/resources/385026)

The only way I could think to solve this bug is to return all versification mappings, even identity mappings which are currently being excluded (e.g. verse 3 -> verse 3) because that's the only way to know if verse 0 is included or not.  If verse 0 is in the source then we'll see a mapping for it in the returned versification data and if it isn't included then that mapping will be missing.

Front-end CMS PR: https://github.com/BiblioNexusStudio/content-manager-web/pull/629

### Fall back to default language Bible versification mappings

I also updated versification mapping to have a fall back.  If the source or target Bible doesn't have any mappings then we try to use the mappings for the language default Bible for the same language.  If that also doesn't have mappings then we continue with the current behavior from before this PR of always doing an identity mapping.  Note that this fallback does _not_ happen for exclusions, just for mappings.

Adding this fall back mostly fixed versification for the other English Bibles by having them use the BSB versification, especially in Psalms where there are a lot of differences between ENG and ORG (base).  The only way to truly fix this is to import correct mapping data for all Bibles but this is a significant improvement when we don't have it.

### DIFF
--- a/src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
+++ b/src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
@@ -57,9 +57,7 @@ public class Endpoint(AquiferDbContext dbContext, ICachingVersificationService v
                 versificationService,
                 ct);
 
-            // exclude mappings where the source and target verse IDs are the same
             verseMappings = versificationMap
-                .Where(mapping => mapping.Value.Count != 1 || mapping.Value[0] != mapping.Key)
                 .Select(
                     mapping => new VerseMapping
                     {
--- a/src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
+++ b/src/Aquifer.Common/Services/Caching/CachingVersificationService.cs
@@ -11,10 +11,12 @@ public interface ICachingVersificationService
 {
     Task<ReadOnlyDictionary<int, IReadOnlyList<string>>> GetBaseVerseIdWithOptionalPartByBibleVerseIdMapAsync(
         int bibleId,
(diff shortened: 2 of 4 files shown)


src/Aquifer.API/Endpoints/Bibles/Versification/Endpoint.cs
```diff
@@ -57,9 +57,7 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
                 versificationService,
                 ct);
 
-            // exclude mappings where the source and target verse IDs are the same
             verseMappings = versificationMap
-                .Where(mapping => mapping.Value.Count != 1 || mapping.Value[0] != mapping.Key)
```
NateMerritt, line 62: This is the main change that required a front-end PR.

jwinston-bn (approved):

PR #737, opened 2025-04-03
Allow creating a project from non-English source data.

NateMerritt:
Implemented:
1. Fetching resource content to translate is now based upon both a source and target language.
2. Fetching resource content to Aquiferize now allows languages other than English (the front-end does not yet use this behavior but it works).  AKA English isn't hardcoded everywhere anymore.
3. Creating a project likewise respects a source and target language ID.  There is still a gate to prevent Aquiferization of non-English resource content but if we want to remove that one validation condition in the future this code should work for creating non-English Aquiferization projects (it's the translation part that is still preventing non-English Aquiferizations).
4. The `TranslationMessageSubscriber` translates these projects after they are started (this required updating the Projects and ResourceContents tables in the DB to hold source language information).

CMS PR: https://github.com/BiblioNexusStudio/content-manager-web/pull/631

Project using non-English source (Traditional Chinese -> Simplified Chinese): https://qa.admin.aquifer.bible/projects/1285 and https://qa.admin.aquifer.bible/resources/385057.


src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -608,6 +617,9 @@ SELECT NULL
                             FROM ResourceContents rc2
                             WHERE
                                 rc2.ResourceId = rc.ResourceId AND
+                                {(filter.ExcludeContentMediaTypes is { Count: > 0 }
```
NateMerritt, line 620: This wasn't necessary for the PR but it was missing so I added it while I was in there.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -241,17 +242,24 @@ public sealed class ResourceContentSearchService(AquiferDbContext dbContext) : I
                 nameof(filter));
         }
 
-        if (filter is { IsTranslated: not null, LanguageId: Constants.EnglishLanguageId })
+        if (filter is { IsTranslated: true, IsPublished: not true })
```
jwinston-bn, line 245: Can you say `false` instead of `not true`?

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -241,17 +242,24 @@ public sealed class ResourceContentSearchService(AquiferDbContext dbContext) : I
...
-        if (filter is { IsTranslated: true, IsPublished: not true })
+        if (filter is { IsTranslated: false, TranslationSourceLanguageId: null })
         {
             throw new ArgumentException(
-                $"The \"{nameof(filter.IsPublished)}\" filter must also be true when filtering by an \"{nameof(filter.IsTranslated)}\" value of true.",
+                $"The \"{nameof(filter.TranslationSourceLanguageId)}\" filter must also be passed when filtering by an \"{nameof(filter.IsTranslated)}\" value of false.",
+                nameof(filter));
+        }
+
+        if (filter is { IsTranslated: not false, TranslationSourceLanguageId: not null })
```
jwinston-bn, line 259: `true`?

src/Aquifer.Common/Services/ResourceContentSearchService.cs
NateMerritt: `IsPublished` is nullable so `not true` pattern matching catches when it is either `null` or `false`.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
NateMerritt: `IsTranslated` is nullable so `not false` pattern matching catches when it is either `null` or `true`.

jwinston-bn (dismissed):

src/Aquifer.Data/Migrations/20250404202836_AddSourceLanguageIdToProjectAndResourceContent.cs
```diff
@@ -0,0 +1,125 @@
...
+            migrationBuilder.DropIndex(
+                name: "IX_ResourceContents_Status",
+                table: "ResourceContents");
+
+            migrationBuilder.AddColumn<int>(
+                name: "SourceLanguageId",
+                table: "ResourceContents",
+                type: "int",
+                nullable: false,
+                defaultValue: 1);
```
NateMerritt, line 30: I set the default to 1.

src/Aquifer.Data/Migrations/20250404202836_AddSourceLanguageIdToProjectAndResourceContent.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            migrationBuilder.AddColumn<int>(
+                name: "SourceLanguageId",
+                table: "Projects",
+                type: "int",
+                nullable: true,
+                defaultValue: 1);
```
NateMerritt, line 37: I set the default to 1.

src/Aquifer.API/Endpoints/Projects/Create/Endpoint.cs
```diff
@@ -102,39 +116,51 @@ private async Task<ProjectEntity> BuildProjectFromRequestAsync(Request request,
...
         UserEntity user,
         CancellationToken ct)
     {
-        if (languageId == Constants.EnglishLanguageId)
+        // Aquiferization
+        if (request.SourceLanguageId == request.TargetLanguageId)
+        {
+            if (request.TargetLanguageId.Value != Constants.EnglishLanguageId)
+            {
+                ThrowError(r => r.TargetLanguageId, "Only English is currently supported for Aquiferization.");
```
jwinston-bn, line 129: We do currently support other languages being Aquiferized though. It's something we added recently, to cover the cases where we bulk loaded/translated something and now they want to add it to a project to review it.

src/Aquifer.Data/Migrations/20250404202836_AddSourceLanguageIdToProjectAndResourceContent.cs
jwinston-bn: Would this get overwritten by a new migration? Maybe it doesn't matter, but the correct way to do this, without manually updating a generated file, would be like how it's done in `ResourceEntity` in the `Configure` method.

NateMerritt: I only wanted the default to be one for the migration itself.  All future inserts should then require this column.  I _thought_ this change would accomplish that based upon docs/SO but looking at the DB schema (I already ran in Dev/QA) it set the column default to 1.

I'll plan to roll forward with this and then manually remove the default after deployment because I only need this for the initial schema update.  Then the code will match the DB schema.

src/Aquifer.API/Endpoints/Projects/Create/Endpoint.cs
```diff
@@ -245,30 +271,34 @@ private async Task<List<ResourceContentEntity>> CreateOrFindTranslatedResourceCo
...
+        IReadOnlyList<int> resourceIds,
         UserEntity user,
         CancellationToken ct)
     {
-        var englishOrLanguageResourceContents = await dbContext.ResourceContents.AsTracking()
-            .Where(rc => request.ResourceIds.Contains(rc.ResourceId) && rc.MediaType != ResourceContentMediaType.Audio)
-            .Where(rc => rc.LanguageId == languageId ||
-                (rc.Resource.ResourceContents.All(rci => rci.LanguageId != languageId) && rc.LanguageId == Constants.EnglishLanguageId))
+        var sourceLanguageOrTargetLanguageResourceContents = await dbContext.ResourceContents.AsTracking()
+            .Where(rc => resourceIds.Contains(rc.ResourceId) && rc.MediaType != ResourceContentMediaType.Audio)
```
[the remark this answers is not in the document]
NateMerritt: Makes sense.  I'll update.

src/Aquifer.Common/Services/ResourceContentSearchService.cs
```diff
@@ -608,6 +617,9 @@ SELECT NULL
                             FROM ResourceContents rc2
                             WHERE
                                 rc2.ResourceId = rc.ResourceId AND
+                                {(filter.ExcludeContentMediaTypes is { Count: > 0 }
+                                    ? $"rc2.MediaType NOT IN @{includeContentMediaTypesParamName} AND"
```
NateMerritt, line 621: 🐛

src/Aquifer.API/Endpoints/Projects/Create/Endpoint.cs
NateMerritt: @jwinston-bn, the front-end only allows creating Aquiferization projects for English.  I think the scenario you are describing falls under the `if (request.IsAlreadyTranslated)` condition immediately below.

jwinston-bn (approved):

PR #738, opened 2025-04-04
Bib 900 leave community reviewer comments open

elisabeth-shakespeare:


src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
```diff
@@ -42,11 +42,6 @@ await historyService.AddSnapshotHistoryAsync(contentVersion,
             contentVersion.ResourceContent.Status,
             ct);
 
-        if (contentVersion.ResourceContent.MediaType == ResourceContentMediaType.Text)
-        {
-            Helpers.SanitizeTiptapContent(contentVersion);
```
NateMerritt, line 47: Does this need to be sanitized in a later step instead?

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -151,82 +161,101 @@ private static async Task<IReadOnlyList<ResourceCommentNotificationData>> GetRes
...
+        var userIds = GetNotifyUserIdsOnCommunityReviewerComment(cofiguration);
+        var isContainingCommunityReviewerComments = userIds is not null && userIds.Contains(userId);
+        const string communityReviewerCommentsSubquery = """
+                                                         OR EXISTS (
+                                                             SELECT NULL
+                                                             FROM CommentThreads ct2
+                                                             JOIN Comments c2 ON c2.ThreadId = ct2.Id
+                                                             WHERE  
+                                                                 c2.UserId IN (
+                                                                     SELECT Id FROM Users WHERE Role = 6
```
NateMerritt, line 176: Please use the constant injected via string interpolation.
```suggestion
                                                                     SELECT Id FROM Users WHERE Role = {UserRole.CommunityReviewer}
```

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -151,82 +161,101 @@ private static async Task<IReadOnlyList<ResourceCommentNotificationData>> GetRes
...
+        IConfiguration cofiguration,
         CancellationToken ct)
     {
+        var userIds = GetNotifyUserIdsOnCommunityReviewerComment(cofiguration);
+        var isContainingCommunityReviewerComments = userIds is not null && userIds.Contains(userId);
+        const string communityReviewerCommentsSubquery = """
+                                                         OR EXISTS (
+                                                             SELECT NULL
+                                                             FROM CommentThreads ct2
+                                                             JOIN Comments c2 ON c2.ThreadId = ct2.Id
```
NateMerritt, line 173: Did this work during testing?  I'd expect that unique aliases would be needed here and `ct2` and `c2` are both already used below.

src/Aquifer.API/Endpoints/Resources/Content/SendForPublisherReview/Community/Endpoint.cs
elisabeth-shakespeare: Nah, this is calls `TiptapConverter.DeserializeForPublish(version.Content);` which takes out the inline markings for Publisher Review. According to BIB-900, for community reviewers sending to publisher, it shouldn't happen.

elisabeth-shakespeare: This endpoint is also specifically for community reviewers sending to publisher so I don't think we have to worry for the other cases in this instance.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
elisabeth-shakespeare: Doesn't string interpolation make it slower?

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
elisabeth-shakespeare: Been working so far, I'll take a closer look.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -151,82 +161,101 @@ private static async Task<IReadOnlyList<ResourceCommentNotificationData>> GetRes
         DbConnection dbConnection,
         int userId,
         DateTime minimumCommentCreatedDate,
+        IConfiguration cofiguration,
         CancellationToken ct)
     {
+        var userIds = GetNotifyUserIdsOnCommunityReviewerComment(cofiguration);
```
NateMerritt, line 167: It's a long name but It's easier to read the code.  I also suggest passing this list into this method instead of `configuration`.
```suggestion
        var communityReviewerCommentNotificationUserIds = GetNotifyUserIdsOnCommunityReviewerComment(configuration);
```

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -151,82 +161,101 @@ private static async Task<IReadOnlyList<ResourceCommentNotificationData>> GetRes
...
+                                     c.Created > c2.Created
+                             ) OR
+                             EXISTS
+                             (
+                                 SELECT NULL
+                                 FROM CommentMentions cm
+                                 WHERE
+                                     cm.CommentId = c.Id AND
+                                     cm.UserId <> c.UserId AND
+                                     rcv.IsDraft = 1 AND
```
NateMerritt, line 248: Why did the `IsDraft` check need to move?  My understanding is the comments all disappear after publishing so even community reviewer comments would disappear in the UI.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
```diff
@@ -100,6 +104,12 @@ public override async Task HandleAsync(Request req, CancellationToken ct)
         await SendOkAsync(notifications, ct);
     }
 
+    private static int[]? GetNotifyUserIdsOnCommunityReviewerComment(IConfiguration configuration)
+    {
+        var userIds = configuration.GetValue<string>("NotifyIdsOnCommunityReviewerComment");
```
NateMerritt, line 109: There's a nicer way to do this that doesn't require embedding magic strings everywhere we want to consume configuration.  I suggest the following:
* Create a new `NotificationOptions` subclass [similar to [this one](https://github.com/BiblioNexusStudio/aquifer-server/blob/fb525bc82f2e3a965cbc46acad0b3f4874b2e865/src/Aquifer.AI/OpenAiTranslationService.cs#L41)).
* Add it to the configuration class as a property [liks this](https://github.com/BiblioNexusStudio/aquifer-server/blob/fb525bc82f2e3a965cbc46acad0b3f4874b2e865/src/Aquifer.Jobs/Configuration/ConfigurationOptions.cs#L21).
* Inject it into the constructor [like so](https://github.com/BiblioNexusStudio/aquifer-server/blob/fb525bc82f2e3a965cbc46acad0b3f4874b2e865/src/Aquifer.AI/OpenAiTranslationService.cs#L66).
* Wire it up in DI [like this](https://github.com/BiblioNexusStudio/aquifer-server/blob/fb525bc82f2e3a965cbc46acad0b3f4874b2e865/src/Aquifer.Jobs/Program.cs#L43).
* Use a JSON object in the appsettings instead of a stand-alone string.  It will only have the one property it in but we can easily add more in the future if needed for additional notification settings.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
NateMerritt: It's only building the string in this case, not doing parameterization of the SQL query.  So yes, it will be very slightly slower.  But not embedding constant values is worth that minor performance cost and also makes the code more readable.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
elisabeth-shakespeare: It didn't make sense to me to have an options object if there was only going to be one setting string.

src/Aquifer.API/Endpoints/Notifications/List/Endpoint.cs
NateMerritt: Yes, it should be.

NateMerritt (approved):

PR #746, opened 2025-04-08
Add Spanish post processor

jwinston-bn:


tests/Aquifer.Public.API.IntegrationTests/Endpoints/Resources/Updates/List/EndpointTests.cs
```diff
@@ -19,7 +19,7 @@ public sealed class EndpointTests(App _app) : TestBase<App>
         new Request
         {
             LanguageCode = "spa",
-            Offset = 10,
+            Offset = 0,
```
jwinston-bn, line 22: Unrelated to PR, but happened to start breaking with it.

PR #756, opened 2025-05-08
Support generating Aquifer.Well.API client code into BibleWell solution.

NateMerritt:
Also, fixes regression with latest FastEndpoints update where the OpenAPI spec switched from camelCase properties to PascalCase properties.  JSON is case-sensitive so the Kiota generated client is also case-sensitive and this change broke the generated clients for the Public API (we haven't deployed the regression commit to prod yet).  This PR fixes that only for response properties.  I couldn't figure out request query string parameter casing and opened a FastEndpoints issue: https://github.com/FastEndpoints/FastEndpoints/issues/947.  Technically speaking, PascalCase should be accepted in query strings on the server so there's no breaking change.  I still want to fix it if we get a good answer from them, though.

See https://github.com/BiblioNexusStudio/bible-well/pull/50 for usage.

### DIFF
--- a/src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
+++ b/src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
@@ -48,6 +48,7 @@ public static class SwaggerDocumentSettings
                         In = OpenApiSecurityApiKeyLocation.Header,
                         Type = OpenApiSecuritySchemeType.ApiKey
                     });
+                ds.MarkNonNullablePropsAsRequired();
                 ds.SchemaSettings.SchemaProcessors.Add(new EnumSchemaProcessor());
                 ds.OperationProcessors.Add(new DefaultParameterOperationProcessor());
             };
--- a/src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
+++ b/src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
@@ -1,4 +1,6 @@
-﻿using Aquifer.Public.API.Helpers;
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+using Aquifer.Public.API.Helpers;
 using FastEndpoints.Swagger;
 using NSwag;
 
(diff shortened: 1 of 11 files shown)


src/Aquifer.Public.API/OpenApi/SwaggerDocumentSettings.cs
```diff
@@ -48,6 +55,7 @@ All endpoints require an API key in the `api-key` header.<br><br>
                         In = OpenApiSecurityApiKeyLocation.Header,
                         Type = OpenApiSecuritySchemeType.ApiKey
                     });
+                ds.MarkNonNullablePropsAsRequired();
```
NateMerritt, line 58: Unfortunately the Kiota client uses nullable properties everywhere, even if the property is required.  Fixing this is their most upvoted issue: https://github.com/microsoft/kiota/issues/3911.  Adding this FastEndpoints setting at least gets us a correct OpenApi JSON spec and whenever Kiota gets a v2 release we should be able to take advantage of non-nullable reference types in the generated request/response objects.

src/Aquifer.Well.API/Endpoints/Languages/List/ListLanguagesEndpoint.cs
```diff
@@ -11,6 +11,11 @@ public override void Configure()
     {
         Get("/languages");
         Options(EndpointHelpers.ServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
+        Summary(s =>
+        {
+            s.Summary = "Return language list.";
+            s.Description = "Return a list of languages that can have associated resources in the Aquifer.";
+        });
```
NateMerritt, line 18: I'm ambivalent on if this documentation is valuable to add or not for the generated client.  I didn't turn on the redoc `/docs` page for the Well API but this description does show up in intellisense comments on the generated C# client types which may be useful.

src/Aquifer.Well.API/OpenApi/ClientGenerationSettings.cs
```diff
@@ -0,0 +1,28 @@
...
+        string swaggerDocumentName)
+    {
+        await app.GenerateApiClientsAndExitAsync(c =>
+        {
+            c.SwaggerDocumentName = swaggerDocumentName;
+            c.Language = GenerationLanguage.CSharp;
+            c.ClientNamespaceName = "BibleWell.Aquifer.API.Client";
+            c.ClientClassName = "AquiferWellApiClient";
+            c.OutputPath = Path.Combine(c.OutputPath, "../../../../bible-well/src/BibleWell.Aquifer.Api/Client");
+            c.ExcludePatterns = ["**/admin/**"];
```
NateMerritt, line 25: There are no admin routes yet but I added in an exclusion just in case.

jwinston-bn (approved):

PR #757, opened 2025-05-08
Fix OpenAPI spec query param property name casing.

NateMerritt:
It turns out that the ordering of `app.UseOpenAPI()` and `app.UseFastEndpoints()` matters.

See https://github.com/FastEndpoints/FastEndpoints/issues/947 for details.


src/Aquifer.Public.API/Program.cs
```diff
@@ -66,11 +71,6 @@
         options.DocumentTitle = "Aquifer API Documentation";
     })
     .UseMiddleware<ApiKeyAuthorizationMiddleware>()
-    .UseFastEndpoints(config =>
```
NateMerritt, line 69: Moving this up above `UseOpenApi()` fixed the casing problem.

jwinston-bn (approved):

PR #762, opened 2025-05-13
Bib 915 set up means to receive push notifications

kaseywright:
This is a basic implementation that will:

- 'install' a device with the respective push notifications service (Google, Apple, etc) via Azure Notifications hub. 
- send the requested push notification with the correct template

This server interacts with Azure Notifications Hub which then intelligently communicates push notifications with all of the service providers instead of us having to manually integrate all of the service providers.

### DIFF
--- a/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationEndpoint.cs
+++ b/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationEndpoint.cs
@@ -4,7 +4,7 @@ using FastEndpoints;
 
 namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Create;
 
-public class DeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<Request>
+public class CreateDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<Request>
 {
     public override void Configure()
     {
--- a/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationEndpoint.cs
+++ b/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationEndpoint.cs
@@ -1,14 +1,16 @@
+using System.Net;
 using Aquifer.Well.API.Endpoints.PushNotifications.Models;
 using Aquifer.Well.API.Services;
 using FastEndpoints;
 
 namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Create;
 
-public class CreateDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<Request>
+public class CreateDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<CreateDeviceInstallationRequest>
 {
     public override void Configure()
     {
         Post("/push-notifications/device-installation");
+        Description(d => d.ProducesProblemFE((int)HttpStatusCode.UnprocessableContent));
         Summary( s =>
         {
             s.Summary = "Create or update a device installation.";
@@ -16,7 +18,7 @@ public class CreateDeviceInstallationEndpoint(INotificationService notificationS
         });
     }
 
-    public override async Task HandleAsync(Request req, CancellationToken ct)
+    public override async Task HandleAsync(CreateDeviceInstallationRequest req, CancellationToken ct)
     {
         var installationSuccess = await notificationService.CreateOrUpdateInstallationAsync(
             new DeviceInstallation {
@@ -29,10 +31,10 @@ public class CreateDeviceInstallationEndpoint(INotificationService notificationS
 
         if (!installationSuccess)
         {
-            await SendAsync("Installation Failed.",422, ct);
+            await SendAsync("Installation Failed.", (int)HttpStatusCode.UnprocessableContent, ct);
             return;
         }
         
-        await SendOkAsync(ct);
+        await SendNoContentAsync(ct);
     }
 }
--- a/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationRequest.cs
+++ b/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationRequest.cs
@@ -2,7 +2,7 @@
 
 namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Create;
 
-public sealed class Request
+public sealed class CreateDeviceInstallationRequest
 {
     public required string InstallationId { get; set; }
     
--- a/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Delete/DeleteDeviceInstallationEndpoint.cs
+++ b/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Delete/DeleteDeviceInstallationEndpoint.cs
@@ -1,13 +1,15 @@
+using System.Net;
 using Aquifer.Well.API.Services;
 using FastEndpoints;
 
 namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Delete;
 
-public class DeleteDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<Request>
+public class DeleteDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<DeleteDeviceInstallationRequest>
 {
     public override void Configure()
     {
         Delete("/push-notifications/device-installation/{DeviceInstallationId}");
+        Description(d => d.ProducesProblemFE((int)HttpStatusCode.UnprocessableContent));
         Summary( s =>
         {
             s.Summary = "Delete a device installation.";
@@ -15,7 +17,7 @@ public class DeleteDeviceInstallationEndpoint(INotificationService notificationS
         });
     }
 
-    public override async Task HandleAsync(Request req, CancellationToken ct)
+    public override async Task HandleAsync(DeleteDeviceInstallationRequest req, CancellationToken ct)
     {
         var deletionSuccess = await notificationService.DeleteInstallationByIdAsync(
             req.DeviceInstallationId,
@@ -23,10 +25,10 @@ public class DeleteDeviceInstallationEndpoint(INotificationService notificationS
 
         if (!deletionSuccess)
         {
-            await SendAsync("Deleting the device installation Failed.",422, ct);
+            await SendAsync("Deleting the device installation Failed.", (int)HttpStatusCode.UnprocessableContent, ct);
             return;
         }
         
-        await SendOkAsync(ct);
+        await SendNoContentAsync(ct);
     }
 }
--- a/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Delete/DeleteDeviceInstallationEndpoint.cs
+++ b/src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Delete/DeleteDeviceInstallationEndpoint.cs
@@ -1,0 +1,27 @@
+using Aquifer.Well.API.Services;
+using FastEndpoints;
+
+namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Delete;
+
+public class DeleteDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<Request>
+{
+    public override void Configure()
+    {
+        Delete("/push-notifications/device-installation/{DeviceInstallationId}");
+    }
+
+    public override async Task HandleAsync(Request req, CancellationToken ct)
+    {
+        var deletionSuccess = await notificationService.DeleteInstallationByIdAsync(
+            req.DeviceInstallationId,
+            ct);
+
+        if (!deletionSuccess)
+        {
+            await SendAsync("Deleting the device installation Failed.",422, ct);
+            return;
+        }
+        
+        await SendOkAsync(ct);
+    }
+}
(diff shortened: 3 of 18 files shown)


src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationEndpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public override async Task HandleAsync(CreateDeviceInstallationRequest req, CancellationToken ct)
+    {
+        var installationSuccess = await notificationService.CreateOrUpdateInstallationAsync(
+            new DeviceInstallation {
+                InstallationId = req.InstallationId,
+                Platform = req.Platform,
+                PushChannel = req.PushChannel,
+                Tags = req.Tags ?? Array.Empty<string>()
```
jwinston-bn, line 28: `Tags` isn't nullable

src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationEndpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+using FastEndpoints;
+
+namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Create;
+
+public class CreateDeviceInstallationEndpoint(INotificationService notificationService) : Endpoint<CreateDeviceInstallationRequest>
+{
+    public override void Configure()
+    {
+        Post("/push-notifications/device-installation");
+        Description(d => d.ProducesProblemFE((int)HttpStatusCode.BadRequest));
```
jwinston-bn, line 13: Looks like 400 is the default, so no need to specify

src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationRequest.cs
```diff
@@ -0,0 +1,14 @@
+// using Aquifer.Well.API.Endpoints.PushNotifications.Models;
```
jwinston-bn, line 1: need this?

src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Delete/DeleteDeviceInstallationEndpoint.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public override async Task HandleAsync(DeleteDeviceInstallationRequest req, CancellationToken ct)
+    {
+        var deletionSuccess = await notificationService.DeleteInstallationByIdAsync(
+            req.DeviceInstallationId,
+            ct);
+
+        if (!deletionSuccess)
+        {
+            await SendAsync("Deleting the device installation Failed.", (int)HttpStatusCode.BadRequest, ct);
```
jwinston-bn, line 28: Is it correct that this is a bad request, rather than a server error or something?

src/Aquifer.Well.API/Endpoints/PushNotifications/Installations/Create/CreateDeviceInstallationRequest.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Aquifer.Well.API.Endpoints.PushNotifications.Installations.Create;
+
+public sealed class CreateDeviceInstallationRequest
+{
+    public required string InstallationId { get; set; }
```
jwinston-bn, line 7: We need to be careful with `required` keyword in requests, because if it's not passed, the json deserializer can behave unexpectedly. We should Use a Validator class to handle required properties, and add `= null!` after these as needed.

src/Aquifer.Well.API/Endpoints/PushNotifications/Requests/NotificationRequestsEndpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+        Summary( s =>
+        {
+            s.Summary = "Requests a push notification.";
+            s.Description = "Requests a push notification be sent with the given content and action.";
+        });
+    }
+
+    public override async Task HandleAsync(NotificationRequest req, CancellationToken ct)
+    {
+        if ((req.Silent && string.IsNullOrWhiteSpace(req?.Action)) ||
```
jwinston-bn, line 23: `req` isn't nullable

jwinston-bn: Also this should be handled in Validator

src/Aquifer.Well.API/Endpoints/PushNotifications/Requests/NotificationRequestsEndpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+            (!req.Silent && !string.IsNullOrWhiteSpace(req?.Text)))
+        {
+            await SendAsync("Bad notification request.", (int)HttpStatusCode.BadRequest, ct);
+        }
+        
+        var success = await notificationService.RequestNotificationAsync(req!, ct);
```
jwinston-bn, line 29: @NateMerritt Wonder your thoughts on this. I don't necessarily think an endpoint request should be forwarded to another service. Feels like separate concerns acting on the object, and it could lead to unexpected behavior in the endpoint. But I also think it'd be heavy-handed to duplicate it.

src/Aquifer.Well.API/Endpoints/PushNotifications/Models/NotificationHubOptions.cs
jwinston-bn, line 1: I'd move this to the Configuration folder, so everything mapped from appsettings is in a single place.

src/Aquifer.Well.API/Services/NotificationHubService.cs
```diff
@@ -0,0 +1,170 @@
...
+            return false;
+        }
+
+        try
+        {
+            await _hub.CreateOrUpdateInstallationAsync(installation, token);
+        }
+        catch
+        {
+            return false;
```
jwinston-bn, line 56: Do we need to know what happened here? This swallows the exception, so we'd have no logging or anything on it.

src/Aquifer.Well.API/Endpoints/PushNotifications/Models/DeviceInstallation.cs
```diff
@@ -0,0 +1,12 @@
+namespace Aquifer.Well.API.Endpoints.PushNotifications.Models;
+
+public class DeviceInstallation
+{
+    public required string InstallationId { get; set; }
+    
+    public required string Platform { get; set; }
```
jwinston-bn, line 7: You should be able to make this NotificationPlatform enum here, to avoid later conversions. Then you can pass either the enum value or the name (e.g. "Apns" or 2 should be accepted).

src/Aquifer.Well.API/Endpoints/PushNotifications/Models/PushNotificationTemplates.cs
jwinston-bn, line 1: I would recommend making this more of a utility that has a helper method to do the replacement, and then the `PrepareNotificationPayload` method you have in `NotificationHubService` doesn't have to do a blind string replace. So I would add that logic to a utility and, for simplicity, remove the const string. So you have a helper method like:

```csharp
public string GetGenericAndroidPayload(string text, string action)
{
    return /* lang=json */$$"""
            {"message":{"notification":{"title":"PushDemo","body":"{{text}}"},"data":{"action":"{{action}}"}}}
        """;
}
```

The `/* lang=json */` tells the IDE to use proper syntax highlighting as well as validation for that language type. The `$$` says to treat a double `{{ }}` as an escape. The triple quotes allows you to put singles in without needing to escape them. I like this approach because it makes it much cleaner to read the actual string.

src/Aquifer.Well.API/Endpoints/PushNotifications/Requests/NotificationRequestsEndpoint.cs
NateMerritt: I agree.  `Request` objects should not be passed to a service layer.  Instead, there should be a model and the endpoint needs to map the request type into that model.  It only takes a few minutes to write mapping code like that.

src/Aquifer.Well.API/Endpoints/PushNotifications/Requests/NotificationRequestsEndpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+        Console.WriteLine("Requested");
```
NateMerritt, line 30: This looks like leftover debugging code.

src/Aquifer.Well.API/Program.cs
```diff
@@ -50,6 +52,11 @@
 builder.Services.AddOptions<ConfigurationOptions>().Bind(builder.Configuration);
 builder.Services.Configure<ApiKeyAuthorizationMiddlewareOptions>(o => o.Scope = ApiKeyScope.WellApi);
 
+builder.Services.AddSingleton<INotificationService, NotificationHubService>();
+builder.Services.AddOptions<NotificationHubOptions>()
```
NateMerritt, line 56: This shouldn't be necessary if you add a `NotificationHub` property to the `ConfigurationOptions` class which is already bound above.

src/Aquifer.Well.API/Services/NotificationHubService.cs
```diff
@@ -0,0 +1,170 @@
...
+
+namespace Aquifer.Well.API.Services;
+
+public class NotificationHubService : INotificationService
+{
+    private readonly NotificationHubClient _hub;
+    private readonly Dictionary<string, NotificationPlatform> _installationPlatform;
+    private readonly ILogger<NotificationHubService> _logger;
+    public NotificationHubService(IOptions<NotificationHubOptions> options, ILogger<NotificationHubService> logger)
+
```
NateMerritt, line 13:
```suggestion
```

src/Aquifer.Well.API/Services/NotificationHubService.cs
```diff
@@ -0,0 +1,170 @@
...
+        };
+    }
+
+    public async Task<bool> CreateOrUpdateInstallationAsync(DeviceInstallation deviceInstallation, CancellationToken token)
+    {
+        if (string.IsNullOrWhiteSpace(deviceInstallation?.InstallationId) ||
+            string.IsNullOrWhiteSpace(deviceInstallation?.Platform) ||
+            string.IsNullOrWhiteSpace(deviceInstallation?.PushChannel))
+        {
+            return false;
```
NateMerritt, line 31: I'd expect that (a) invalid device installation requests would be caught by an endpoint validator, and then (b) this `public` method can then throw an `ArgumentException` or `ArgumentNullException` if it gets data that it can't handle (which should never happen).

src/Aquifer.Well.API/Services/NotificationHubService.cs
```diff
@@ -0,0 +1,170 @@
...
+                await SendPlatformNotificationsAsync(androidPayload, iOSPayload, notificationRequest.Tags, token);
+            }
+            else
+            {
+                var notificationTasks = notificationRequest.Tags
+                    .Select((value, index) => (value, index))
+                    .GroupBy(g => g.index / 20, i => i.value)
+                    .Select(tags => SendPlatformNotificationsAsync(androidPayload, iOSPayload, tags, token));
+
+                await Task.WhenAll(notificationTasks);
```
NateMerritt, line 128: Is there any risk of being rate limited if we send to many requests at once in parallel?

src/Aquifer.Well.API/Services/NotificationHubService.cs
```diff
@@ -0,0 +1,170 @@
...
+            _logger.LogError(e, "Unexpected error sending notification");
+            return false;
+        }
+    }
+
+    private string PrepareNotificationPayload(string template, string text, string action)
+    {
+        return template
+        .Replace("$(alertMessage)", text, StringComparison.InvariantCulture)
+        .Replace("$(alertAction)", action, StringComparison.InvariantCulture);
```
NateMerritt, line 144:
```suggestion
            .Replace("$(alertMessage)", text, StringComparison.InvariantCulture)
            .Replace("$(alertAction)", action, StringComparison.InvariantCulture);
```

src/Aquifer.Well.API/Endpoints/PushNotifications/Requests/NotificationRequestsEndpoint.cs
```diff
@@ -0,0 +1,40 @@
...
+using Aquifer.Well.API.Services;
+using FastEndpoints;
+
+namespace Aquifer.Well.API.Endpoints.PushNotifications.Requests;
+
+public class NotificationRequestsEndpoint(INotificationService notificationService) : Endpoint<NotificationRequest>
+{
+    public override void Configure()
+    {
+        Post("/push-notifications/requests");
```
NateMerritt, line 12: This route should not be available anonymously because then anyone could send a post notification to all of the BibleWell mobile clients.  If we want admin behaviors in the Well API then I think we'll need (a) Auth0 authentication and permission authorization as well as (b) all such routes to begin with the `/admin/...` path (which is already excluded from client source generation).

jwinston-bn: That's a good point. Maybe this should be in Internal API. I don't think we want to set up Auth0 on this endpoint just for this.

src/Aquifer.Well.API/Services/NotificationHubService.cs
kaseywright: The only limits that I can find are:
- [This MS forums answer](https://learn.microsoft.com/en-us/answers/questions/783989/how-many-requests-per-min-we-can-make-to-send-mess)
- [The pricing page for Hubs](https://azure.microsoft.com/en-us/pricing/details/notification-hubs/)
- [This FAQ doc](https://learn.microsoft.com/en-us/azure/notification-hubs/notification-hubs-push-notification-faq) - which defines some terms that are ambiguous on the pricing page

In short, no. It should be fine unless we have an unexpected insane amount of sends.

PR #764, opened 2025-05-15
Fix book audio deserialization on every chapter and use dictionaries for audio lookup.

NateMerritt:


jwinston-bn (approved):

src/Aquifer.Public.API/Endpoints/Bibles/Texts/Get/Endpoint.cs
```diff
@@ -82,30 +94,37 @@ public override async Task HandleAsync(Request request, CancellationToken ct)
                 Chapters = bookData.Chapters!
                     .Select(ch =>
                     {
-                        var chapterAudio = !request.ShouldReturnAudioData
-                            ? null
-                            : DeserializeAudioUrls(bookData.AudioUrls)?.Chapters?.FirstOrDefault(c => c.Number == ch.Number.ToString());
```
NateMerritt, line 87: This `DeserializeAudioUrls` call deserializes all _book_ audio data but was previously being run for every _chapter_ in the book, resulting in performance issues.

PR #765, opened 2025-05-16
Bible endpoints for Well

jwinston-bn:


src/Aquifer.Well.API/Endpoints/Bibles/Get/GetBibleResponse.cs
```diff
@@ -0,0 +1,57 @@
...
+
+public class GetBibleResponse
+{
+    public required int BibleId { get; init; }
+    public required string BibleName { get; init; }
+    public required string BibleAbbreviation { get; init; }
+    public required IReadOnlyList<ResponseBook> Books { get; init; }
+}
+
+public class ResponseBook
```
NateMerritt, line 14: These class names are the same ones used for code generation.  I would prefer either `GetBibleResponseBook` or just `Book`.  Thoughts?

src/Aquifer.Well.API/Endpoints/Bibles/List/ListBiblesEndpoint.cs
```diff
@@ -0,0 +1,47 @@
...
+    public override void Configure()
+    {
+        Get("/bibles");
+        Options(EndpointHelpers.ServerCacheInSeconds(EndpointHelpers.OneHourInSeconds));
+        Description(d => d.ProducesProblemFE());
+        Summary(s =>
+        {
+            s.Summary = "Get a list of Bibles.";
+            s.Description =
+                "For a given optional language id, returns the Bibles in the system including language information, abbreviation codes, whether or not the Bible supports a Greek alignment, and license information.";
```
NateMerritt, line 21:
```suggestion
                "For a given optional language id, returns the Bibles in the system including language information, abbreviation codes, and license information.";
```

src/Aquifer.Data/Schemas/BibleLicenseInfoSchema.cs
```diff
@@ -1,6 +1,6 @@
-﻿namespace Aquifer.API.Endpoints.Bibles;
+namespace Aquifer.Data.Schemas;
 
-public sealed class BibleLicenseInfo
+public sealed class BibleLicenseInfoSchema
```
NateMerritt, line 3: Thoughts on dropping the `Schema` suffix?  These types will be the same name in the auto-generated client.

src/Aquifer.Data/Schemas/ParentResourceLicenseInfoSchema.cs
```diff
@@ -1,6 +1,6 @@
-﻿namespace Aquifer.API.Endpoints.Resources;
+namespace Aquifer.Data.Schemas;
 
-public sealed class ResourceLicenseInfo
+public sealed class ParentResourceLicenseInfoSchema
```
NateMerritt, line 3: Technically this will break the auto-generated Public API client if anyone is using the old `ResourceLicenseInfo` generated class but that's probably fine.

But see my thoughts above about dropping `Schema`.

PR #769, opened 2025-06-16
Remove BN Name

jwinston-bn:


NateMerritt (changes requested):
This is going to break CI/CD for tests.  At a minimum the left side of these two lines needs to change to the new connection string name: https://github.com/BiblioNexusStudio/aquifer-server/blob/53d471ab6d50740bee699368c9d2f689da9e9576/.github/workflows/pre-merge.yml#L44-L45.  If we change the right side then we'll need to update the GitHub Actions secret names.

